PR #4, opened 2023-01-11
Added operating instructions

416rehman:
When first cloning the repo, nothing is really clear apart from the docker-compose file. To clear things up and help new devs get started quickly without having to delve into research to understand the services, I have added instructions on how to setup a local dev environment and all the resources exposed by the services, and what those services do, such as the webserver, its endpoints, and where some of these are configured.

Not sure, if these docs should be in their own separate folder, or be part of the main README.md.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,7 @@ dist
 
 # TernJS port file
 .tern-port
+
+#IDEs
+.idea
+.vscode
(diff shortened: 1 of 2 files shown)


416rehman:
Never mind, apparently we will be using Route53, not PowerDNS.

PR #39, opened 2023-01-26
Add package-lock.json, remove deps to fix security issues in npm install

humphd:
I've fixed a number security issues by removing the `express-prometheus-middleware` and its types.  We aren't going to use Prometheus, so it's not worth having in the tree.

I've also added a lock file so that we can pin our versions properly.

Finally, I've fixed the package `name` and add a `version`, as well as bumping a few of the dependencies that were out of date.

I've made node 16 our required version.


package.json
```diff
@@ -43,7 +43,6 @@
     "compression": "^1.7.4",
     "cross-env": "^7.0.3",
     "express": "^4.18.2",
-    "express-prometheus-middleware": "^1.2.0",
     "isbot": "^3.6.5",
     "morgan": "^1.10.0",
     "prom-client": "^14.1.1",
```
SerpentBytes, line 48: Thanks for the PR. Since we opt not to use Prometheus, I am unsure if we need `prom-client`. Other than that, it looks good to me.

humphd: Great catch, removed in a new commit.

SerpentBytes (dismissed):

SerpentBytes (approved):
Looks good to me.

Myrfion (approved):
LGTM

PR #47, opened 2023-01-28
Adding rules and coding conventions (EsLint, Prettier)

Ririio:
- I looked through the formats given to [telescope](https://github.com/Seneca-CDOT/telescope), and added/remove some of it based on what we'll be required for the project
- added rules based on [EsLint](https://eslint.org/docs/latest/rules/) recommendation

I'm not sure if I should be adding the [eslint-plugin](https://www.npmjs.com/package/eslint-plugin-playwright) for playwright & jest knowing they haven't been implemented yet


sirinoks:
@Ririio Do these changes require developers to additionally install/configure EsLint/Prettier by themselves? 

Do they need to match their settings of the project onto their, let's say, VScode IDE?
Asking for documentation sake. If there is any actions required to run it, please add the instructions if you can. And if not - let's talk to figure out what is needed, I will write it down.

.prettierignore
```diff
@@ -9,3 +12,7 @@ node_modules
 /postgres-data
 
 /app/styles/tailwind.css
+
+*.md
```
humphd, line 16: Let's include `.md` in Prettier, it will happily format that for us too.

humphd (changes requested):
Looks good.  One thing to fix.

.prettierignore
Ririio: Alright, I just remember having issues with prettier formatting my .md files and breaking it. The issue was on my side, I'm just not quite sure what the main cause was

Ririio:
> @Ririio Do these changes require developers to additionally install/configure EsLint/Prettier by themselves?
> 
> Do they need to match their settings of the project onto their, let's say, VScode IDE? Asking for documentation sake. If there is any actions required to run it, please add the instructions if you can. And if not - let's talk to figure out what is needed, I will write it down.

No as far as I know prettier and EsLint was already part of blue stacks (`npm install` should allow them to use it). I have the `.vscode` comitted so that there is not need for configuration in the system-- I do recommend installing prettier extension to make the vscode settings work.

 I also, suggest reading through the EsLint rules for [recommendation](https://eslint.org/docs/latest/rules/) to see if there are any that we can ignore; I read through all the recommended rules, but there's only so much I can understand I would appreciate if you can suggest for people to review to see if there are any rules we can ignore or implement to linter

.prettierignore
```diff
@@ -1,3 +1,6 @@
+app/routes/entry.server.tsx
```
humphd, line 1: We don't want to ignore anything in `app/`.  All of our code should be formatted.

package.json
```diff
@@ -29,7 +29,6 @@
     "typecheck": "tsc && tsc -p cypress",
     "validate": "run-p \"test -- --run\" lint typecheck test:e2e:run"
   },
-  "prettier": {},
   "eslintIgnore": [
```
humphd, line 32: Since you're preferring ignore files vs. doing things in `package.json`, I'd move this to a separate file as well.

humphd (changes requested):

package.json
Ririio: It's my first time every seeing that line, I was just wondering what it's purpose is? I assumed it's where we put all our formatting options from? If that was the case I already created a `.prettierrc` for it

humphd: I'm talking about `eslintIgnore`.  See https://eslint.org/docs/latest/use/configure/ignore

humphd:
A few git issues here.  First, you're working on your `main` branch.  You can't do this: all work must be done on a topic branch.  Otherwise you'll get your `main` branch out of sync with the rest of the project.

When you want to start working on a fix, you would do this:

```sh
# Start by updating your main branch to match the upstream repo
$ git checkout main
$ git pull upstream main
# Now create a new branch for your work
$ git checkout -b issue-47
# Do your work, add, commit, etc. then push
$ git push origin issue-47
```

In your case, your `main` branch is going to be in the wrong place when you finish this PR.  After you get this PR landed, you'll need to reset your `main` branch:

```sh
# Update your local copies of the branches in the upstream repo (i.e., main)
git fetch upstream
# Reset your main branch to match where the upstream main branch is now
git checkout -B main upstream/main
```

Now your branch will be in the right spot going forward.

Also, I notice you pushing `0 commits` over and over. What are you doing that is causing this?  Are you forgetting to `add`?  Make sure you use `git status` all the time to see what's in the staging area.

Let me know if you have questions.  I want to tighten up your git workflow so we can go fast together.

shawnyu5:
Im not sure if the `.vscode` folder should be in the repo. Not everyone uses vscode, and the people that do may have their own config for it. We as a project should not enforce the settings for a particular editor.

shawnyu5:
If we *really* care about enforcing people's editor settings, there is the idea of an [editorconfig](https://editorconfig.org/), which is "cross-editor", and is not tailored towards a single editor. But I doubt this is worth the effort. This is an open-source project, people can configure their own editors

SerpentBytes:
@shawnyu5, VSCode is still widely used. Having the `.vscode` folder in the tree would override the user's default IDE settings with the settings we have specified for Starchart for consistency reasons. If the user runs the project on Webstorm or another IDE, the `.vscode` folder should not cause a problem.

shawnyu5:
That should be the developer's responsibility, not us as an open-source project to configure their editor for them. We should enforce those things via CI CD pipelines to check for formatting, etc. Not through their editor config.

shawnyu5:
Also think about if in every project, your editor looks and works a little differently, because the project enforced some settings on there... That would drive me mad...

humphd:
I'd like `.vscode` in the tree.  If you don't use VSCode, it won't affect you.  Add an `.editorconfig` if you want something generic.

Issue #25, opened 2023-01-25
Add playwright for end-to-end testing


Eakam1007:
Remix's Blues stack, which is to be used as a basis for initial setup (#7) includes [Cypress](https://www.cypress.io/) for end-to-end testing. 
We would need to swap this out for [Playwright](https://playwright.dev/).

humphd:
In the CI work that @shawnyu5 is doing in #55, we're getting [type checking errors](https://github.com/shawnyu5/starchart/actions/runs/4049824200/jobs/6966608142) on the cypress tests.  Ripping out Cypress will fix this:

```
Run npm run typecheck

> @senecacdot/starchart@0.0.0 typecheck
> tsc && tsc -p cypress

Error: cypress/e2e/smoke.cy.ts(5,8): error TS2339: Property 'cleanupUser' does not exist on type 'cy & CyEventEmitter'.
Error: cypress/e2e/smoke.cy.ts(15,8): error TS2339: Property 'visitAndCheck' does not exist on type 'cy & CyEventEmitter'.
Error: cypress/e2e/smoke.cy.ts(32,8): error TS2551: Property 'login' does not exist on type 'cy & CyEventEmitter'. Did you mean 'log'?
Error: cypress/e2e/smoke.cy.ts(33,8): error TS2339: Property 'visitAndCheck' does not exist on type 'cy & CyEventEmitter'.
Error: cypress/support/create-user.ts(23,22): error TS2554: Expected 3 arguments, but got 2.
Error: cypress/support/create-user.ts(2[7](https://github.com/shawnyu5/starchart/actions/runs/4049824200/jobs/6966608142#step:6:8),5): error TS2345: Argument of type '{ request: Request; userId: any; remember: false; redirectTo: string; }' is not assignable to parameter of type '{ request: Request; username: string; remember: boolean; redirectTo: string; }'.
  Object literal may only specify known properties, and 'userId' does not exist in type '{ request: Request; username: string; remember: boolean; redirectTo: string; }'.
Error: cypress/support/create-user.ts(27,1[8](https://github.com/shawnyu5/starchart/actions/runs/4049824200/jobs/6966608142#step:6:9)): error TS233[9](https://github.com/shawnyu5/starchart/actions/runs/4049824200/jobs/6966608142#step:6:10): Property 'id' does not exist on type 'User'.
Error: Process completed with exit code 2.Run npm run typecheck

> @senecacdot/starchart@0.0.0 typecheck
[...]

PR #55, opened 2023-01-30
ci: formatting and linting

shawnyu5:
Create a GitHub action to check formatting and linting using prettier on all PRs to main branch

### DIFF
--- a/.github/workflows/node.js.yaml
+++ b/.github/workflows/node.js.yaml
@@ -1,6 +1,3 @@
-# This workflow will do a clean installation of node dependencies, cache/restore them, build the source code and run tests across different versions of node
-# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs
-
 name: Node.js CI
 
 on:
@@ -11,14 +8,11 @@ on:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
-
     strategy:
       matrix:
         node-version: [14.x, 16.x, 18.x]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
-
     steps:
     - uses: actions/checkout@v3
     - name: Use Node.js ${{ matrix.node-version }}
@@ -29,3 +23,9 @@ jobs:
     - run: npm ci
     - run: npm run build --if-present
     - run: npm test
+    - name: Prettier Action
+      uses: creyD/prettier_action@v4.2
+      with:
+        prettier_options: --check **/*.{ts,md}
+
+
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,0 +1,31 @@
+# This workflow will do a clean installation of node dependencies, cache/restore them, build the source code and run tests across different versions of node
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs
+
+name: Node.js CI
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [14.x, 16.x, 18.x]
+        # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v3
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'npm'
+    - run: npm ci
+    - run: npm run build --if-present
+    - run: npm test


.github/workflows/node.js.yaml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: "npm"
+      - run: npm ci
+      - run: npm run build --if-present
+      - run: npm test
+      - name: Prettier check
+        uses: creyD/prettier_action@v4.2
+        with:
+          prettier_options: --check **/*.{ts,md}
```
humphd, line 29: We have more file types than these two (e.g., `tsx`).  Let's make sure this includes them all.

.github/workflows/node.js.yaml
```diff
@@ -0,0 +1,35 @@
...
+        # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: "npm"
+      - run: npm ci
+      - run: npm run build --if-present
```
humphd, line 24: `--if-present` not needed, since we have `build`

.github/workflows/node.js.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - run: npm test
+      - name: Prettier check
+        uses: creyD/prettier_action@v4.2
+        with:
+          prettier_options: --check **/*.{ts,md}
+      - name: eslint check
+        run: |
+              # have to install eslint here for some reason... Other wise eslint wont be found
```
humphd, line 32: I don't understand this, how are you testing?

.github/workflows/node.js.yaml
```diff
@@ -0,0 +1,35 @@
...
+              npm install -D eslint
+              npm run lint
+      - run: npm run typecheck
```
humphd, line 35: Some steps have names, and some don't.  I'd pick one and do it that way throughout.

.github/workflows/node.js.yaml
```diff
@@ -0,0 +1,35 @@
...
+      matrix:
+        node-version: [16.x, 18.x]
+        # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: "npm"
```
humphd, line 22: You use quotes on strings in some cases, and in others don't.  You don't need to quote in YAML, I'd remove quotes

humphd (changes requested):

.github/workflows/node.js.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [16.x, 18.x]
```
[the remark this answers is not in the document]
shawnyu5: oh, we only commit to supporting LTS, alright

.github/workflows/node.js.yaml
shawnyu5: For some reason, running `npm run lint` locally works, but in the CI environment, it tells me it can't find eslint. I should do more testing, installing eslint in the action manually shouldn't be necessary I feel.

humphd: That is bizarre.  If you want to file a follow-up bug in a new issue to look into this later, we can sort it out then.  It doesn't need to block this; but we shouldn't need to do it.

shawnyu5: Hmm, let me do some debugging. If I cant find anything in a reasonable time frame, I'll file an issue for later

shawnyu5: Found the issue. It was the prettier action doing some weird things making the `node_modules` vanish. Moving the eslint action above the prettier action fixed it.

shawnyu5: Hmm that would involve caching the node_modules manually? Let me look into how to do that

shawnyu5: The node action currently caches the node install it self, it does not cache node_modules

https://github.com/shawnyu5/starchart/actions/runs/4039795706/jobs/6944864665

shawnyu5:
Here is an implementation of attempting to fetch node_modules from cache. If it doesn't exist, download dependencies, run a job, then put node_modules into the cache. But restoring the cache and creating the cache seems to have quite a bit of boilerplate code. Im not sure if I like it. Thoughts @humphd?

I've only implemented it for one job for now. If we like this, I'll do it for the other jobs as well

.github/workflows/node.js.yaml
```diff
@@ -0,0 +1,66 @@
...
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
```
humphd, line 12: Because this is only going to use one version of node, I'd skip the matrix, and just install node 18 below and use that throughout.

.github/workflows/node.js.yaml
```diff
@@ -0,0 +1,66 @@
...
+        # run: npm ci
+
+      # - name: Get npm cache directory
+        # id: npm-cache-dir
+        # shell: bash
+        # run: echo "dir=$(npm config get cache)" >> ${GITHUB_OUTPUT}
+
+      - name: Restore node_modules cache
+        id: node_modules-cache-restore
+        uses: actions/cache/restore@v3
```
humphd, line 33: I've never used this.  If you think it will work, let's try.

However, we can also do this later if you want to keep things simple and always install the deps in every job

humphd (commented):
Let's go ahead with splitting into separate jobs, and you can either use caching or not as you see fit.

I think the down side of it is that it's going to necessarily add complexity and another layer of potential failure.  However, it may also speed things up.  We can always do the opposite after we learn how it works.

shawnyu5:
CI seems to be working well. We cache the node_modules whenever possible, and restore it from cache when needed. We will only reinstall node_modules if `package.lock.json` changes.

[attempt 1](https://github.com/shawnyu5/starchart/actions/runs/4049644357/attempts/1) is ran with a regenerated package.lock, and reinstalls the node_modules

[attempt 2](https://github.com/shawnyu5/starchart/actions/runs/4049644357/jobs/6966265175) fetches node modules from cache (cached from previous job).

Also note CI currently fails with the [latest code from main](https://github.com/shawnyu5/starchart/actions/runs/4049661334). The typecheck job currently is at a failing state.

.github/workflows/node.js.yaml
```diff
@@ -0,0 +1,167 @@
+name: Node.js CI
```
humphd, line 1: Let's rename this to "CI" or "Continuous Integration" and the file to `ci.yml`.  The node.js part is irrelevant.

humphd (changes requested):
This is looking awesome.  A couple of minor nits

humphd:
> Also note CI currently fails with the [latest code from main](https://github.com/shawnyu5/starchart/actions/runs/4049661334). The typecheck job currently is at a failing state.

I left a note for @Eakam1007 to fix this in https://github.com/Seneca-CDOT/starchart/issues/25#issuecomment-1409618758

SerpentBytes:
Quick question: Are we only supporting Node version 18?

humphd:
> Quick question: Are we only supporting Node version 18?

I would, since it's LTS

SerpentBytes (dismissed):
This is looking good!

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,167 @@
...
+
+      - name: Cache node_modules
+        uses: actions/cache@v3
+        with:
+          path:
+            ./node_modules
+            # cache node modules using the same key as restore.
+          key: ${{ steps.node_modules-cache-restore.outputs.cache-primary-key }}
+
+  Eslint:
```
humphd, line 95: ESLint

humphd (dismissed):
One nit, looks great.

I'm OK to land this with the TS failure, since it's filed.

shawnyu5:
Changed the job name. Seems like the previous review is stale now

SerpentBytes (approved):

humphd (approved):
Love it.

PR #56, opened 2023-01-30
Implementing prettier formatting through external file

Ririio:
## This a new pull request based on #47 ##

Adding eslint recommendation | Creating external eslintignore file

Initial .vscode settings configuration

ignoring my-sql directory for linter

removing cypress and tailwind from prettierignore

adding telescope configuration for vscode

Adding newline at the end of each file


.eslintignore
```diff
@@ -0,0 +1,4 @@
+/node_modules
+/build
+/public/build
+/postgres-data
```
humphd, line 4: `/mysql-data`

.prettierignore
```diff
@@ -8,4 +8,4 @@ node_modules
 /cypress/videos
 /postgres-data
 
-/app/styles/tailwind.css
+/app/styles/tailwind.css
```
humphd, line 11: You can remove this.

.prettierignore
```diff
@@ -8,4 +8,4 @@ node_modules
 /cypress/videos
 /postgres-data
```
humphd, line 9: You can remove this too.

.prettierrc
```diff
@@ -0,0 +1,15 @@
...
+    "jsxSingleQuote": false,
+    "bracketSpacing": true,
+    "arrowParens": "always",
+    "embeddedLanguageFormatting": "auto",
+    "endOfLine": "lf",
+    "insertPragma": false,
+    "proseWrap": "preserve",
+    "quoteProps": "consistent",
+    "requirePragma": false
+}
```
humphd, line 15: You should always include a newline at the end of your files (GitHub is marking it to show you it's missing).  The reason we do this is so that files don't get concatenated together, and the end of one file be on the same line as the start of another.

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
humphd, line 1: While you're in this file, can you add what we have in Telescope? https://github.com/Seneca-CDOT/telescope/blob/master/.vscode/settings.json

humphd (changes requested):

.eslintignore
```diff
@@ -0,0 +1,5 @@
...
```
humphd, line 4: Remove `/postgres-data`, we don't have it anymore.

.prettierignore
```diff
@@ -3,9 +3,3 @@ node_modules
 /build
 /public/build
 .env
-
-/cypress/screenshots
-/cypress/videos
-/postgres-data
```
humphd, line 9: Add `/mysql-data` so it doesn't run prettier in there (that's the new database volume).

humphd (changes requested):

.eslintrc.js
```diff
@@ -1,15 +1,16 @@
...
-    "prettier",
+    'eslint:recommended',
+    '@remix-run/eslint-config',
+    '@remix-run/eslint-config/node',
+    '@remix-run/eslint-config/jest-testing-library',
+    'prettier',
   ],
   env: {
-    "cypress/globals": true,
+    'cypress/globals': true,
```
Myrfion, line 11: Since we are removing `cypress` maybe it makes sense to remove it from this file?

.eslintrc.js
```diff
@@ -1,15 +1,16 @@
...
   },
-  plugins: ["cypress"],
+  plugins: ['cypress'],
```
Myrfion, line 13: same thing regarding the `cypress` is there a good reason to have it here?

Ririio: I wasn't sure if I should remove it considering someone was already given a task regarding removal of cypress and tailwind from our repo

.prettierrc
```diff
@@ -0,0 +1,15 @@
+{
+  "tabWidth": 2,
+  "printWidth": 100,
+  "singleQuote": true,
+  "trailingComma": "es5",
```
dadolhay, line 5: A lot of these are already the default, like
"tabWidth": 2,
"trailingComma": "es5",
"jsxSingleQuote": false,
 "bracketSpacing": true,
 etc.
 
 I would not hardwire in the defaults again

.eslintrc.js
dadolhay: I would suggest to remove those too

.eslintrc.js
dadolhay: I would suggest to remove those too

dadolhay (changes requested):

.prettierrc
Ririio: What would happen if a contributor has their `tabwidth` set to a different value? I thought we implement them here to keep consistency just in case some people have their settings altered?

humphd: One of the reasons people specify the defaults for Prettier like this is that the project has been known to change the defaults in the past.  What then happens is you update the prettier dependency, and your entire tree gets reformatted for no good reason.

I think explicit defaults are useful, even if only to document our decisions.

humphd (dismissed):

humphd (approved):

shawnyu5:
This branch seems to have conflicts with main...

humphd:
cc @Eakam1007, who just wrote this stuff for Chakra UI

humphd:
you still have conflicts here.

Ririio:
> you still have conflicts here.

It was conflict through imports, it should be fixed now

humphd:
You are failing prettier CI now.  Please get this passing CI and mergable.

shawnyu5:
ahh, since this PR changes `.prettierrc`, prettier on this branch will format according to this PRs prettierrc, which is different than the one on master. Hence prettier check will always fail

shawnyu5:
I think we can just merge with the failing prettier check for now. When this PR lands, CI on main will use the new prettierrc, and will pass CI again.

Tho this PR handling the reformatting of all the files is a little risky. Ideally, formatting all files should've been its own separate PR. But before we merge this, we got to make sure we only changed formatting from main, not any of the code contents

Ririio:
> I think we can just merge with the failing prettier check for now. When this PR lands, CI on main will use the new prettierrc, and will pass CI again.
> 
> Tho this PR handling the reformatting of all the files is a little risky. Ideally, formatting all files should've been its own separate PR. But before we merge this, we got to make sure we only changed formatting from main, not any of the code contents

If we're not going to worry about the prettier, I can roll it back to the commit that didn't affect the files?

shawnyu5:
If we don't worry about formatting in this PR, when we merge, prettier check on main will fail. A separate PR would be needed to reformat all files. This is a little safer and keeps the scope of the PR limited

Ririio:
> If we don't worry about formatting in this PR, when we merge, prettier check on main will fail. A separate PR would be needed to reformat all files. This is a little safer and keeps the scope of the PR limited

I rolled it back to the original commit without any formatting. There shouldn't be any issue when merging now

shawnyu5:
This PR adds a `.prettierrc`, and reformats all files to use the new formatting standard

.prettierrc
```diff
@@ -0,0 +1,15 @@
+{
+   "tabWidth": 3,
```
humphd, line 2: This should be 2

humphd (changes requested):

.prettierrc
shawnyu5: oops, forgot to change it back after testing

Ririio:
There's always bound to be a conflict, and prettier will always fail unless we push this pr. We can roll it back to the commit which only affects 5 files as to the entire repo...

shawnyu5:
Prettier should not always fail, [this commit](https://github.com/Seneca-CDOT/starchart/pull/56/commits/9e21218c3a3ddc832eba8575da63b65c5565e71c) has prettier passing ci

Ririio:
> Prettier should not always fail, [this commit](https://github.com/Seneca-CDOT/starchart/pull/56/commits/9e21218c3a3ddc832eba8575da63b65c5565e71c) has prettier passing ci

I'm referring to it consistently failing whenever a new a commit is pass. We can fix those issues sure, but isn't it better to send the settings now and do formatting later?

.eslintignore
dadolhay, line 4: Do we use postgres?

.eslintrc.js
dadolhay: I would remove this because if the cypress replace PR will have to, it will cause a conflict. Easier to do here and now

.eslintrc.js
dadolhay: same as above

.prettierignore
dadolhay: ^^

.prettierignore
dadolhay: ^^

.eslintignore
```diff
@@ -0,0 +1,5 @@
...
+/mysql-data
```
dadolhay, line 5: /mysql-data does not contain js/ts code, no need to add it

dadolhay (changes requested):

dadolhay (approved):
This will cause some conflicts for sure, but at least the code becomes consistent with the new rules

.eslintignore
```diff
@@ -1,5 +1,3 @@
 /node_modules
 /build
 /public/build
-/postgres-data
-/mysql-data
```
shawnyu5: Why do we want to lint the sql data folder?

I get weird errors if the sql data folder is linted

image

shawnyu5: Same issue with `npm run format`

dadolhay: Only js, ts jsx and tsx files should be ever linted. Give me a sec

dadolhay (approved):

.eslintignore
shawnyu5: We have other file types in the repo, like markdown and yaml. I think we should be excluding the stuff we dont want to lint / format, rather than including everything we would like to

shawnyu5: theres also a `.js.map`

shawnyu5:
Still the same error
image

.eslintignore
dadolhay: The issue is, that that folder it references was created by docker, and it cannot scan the folder for js/ts/jsx/tsx files so it fails.
You are right, because of this mysql-data has to be added to eslintignore and prettierignore

dadolhay (dismissed):

Ririio:
@shawnyu5 do you mind approving it if there's no more mistakes?

shawnyu5:
A hack we can do, what I've been doing. For the format script, we can change it to `prettier --write * .*`. That seems to carry on despite the error.

Ririio:
> A hack we can do, what I've been doing. For the format script, we can change it to `prettier --write * .*`. That seems to carry on despite the error.

what does it do?

Ririio:
> A hack we can do, what I've been doing. For the format script, we can change it to `prettier --write * .*`. That seems to carry on despite the error.

what does it do?

dadolhay:
https://github.com/prettier/prettier/issues/11568

dadolhay (approved):
Since this is not a new issue, I think the prettier problem should be handled in a different ticket. I think this is good to merge

shawnyu5:
I dont think we want to merge with a broken prettier script. Other people will have a pretty big headache formatting their code....

humphd:
Closing in favour of #94

Issue #50, opened 2023-01-28
Create deployment webhook and script


humphd:
As part of #43, we need a webhook that can be triggered via GitHub Actions, and which causes our staging or production deployments to update the Docker image tag they use.

The way we'll do this is to use [webhook](https://github.com/adnanh/webhook).  We'll run it on port `8000`.  GitHub Actions will do a `POST` to `https://mycustomdomain.senecacollege.ca/hooks/deploy` which can trigger a `deploy` webhook (i.e., that would be its `id`).  We'd send the new tag to use in the body.  The webhook would then run a shell script that passes the new tag via a variable (e.g., environment variable).

I have a [demo of this working](https://github.com/Seneca-CDOT/starchart/wiki/Docker).  However, it's not secured at all.  We should add some kind of security token that the server expects to receive, etc.

This issue is about:

1. Creating the `hooks.json` file to configure the webhook(s)
2. Creating the shell script we need to run to update the containers to use new images

humphd:
We could also consider running this webhook in Docker; but we'd have to mount the system's docker socket file so it can trigger the Docker update in the host from within the container.

PR #57, opened 2023-01-30
ci(webhooks): dockerized webhooks with tokens

416rehman:
# Description

This is a direct PR for #50 and provides a dockerized webhook service, along with simple token-based authentication.

A custom-made docker service has been added to the `docker-compose` file, called `webhook` and that contains the functionality for the webhook package. This is a Docker-in-Docker service therefore the host-side docker socket is mounted to this container so it can act upon running containers.

## Authentication
An `authenticate.sh` utility script is provided to handle authentication. By default, it will compare the values of the `DEFAULT_AUTH_TOKEN` (set through `docker-compose.yml`) and `HOOK_token` (set through `hooks.yml`) env vars, if the values match, it returns 0, otherwise a non-zero number. This can be used by webhook scripts to achieve authentication, like so
```sh
if .authenticate.sh; then   # the "token" query param must be the same as $DEFAULT_AUTH_TOKEN
        echo "You are in!"
else
        echo "Auth failed"
fi;
```
The authentication token can be overriden for webhooks so it uses a custom token and not the `DEFAULT_AUTH_TOKEN`. This can be done by providing the expected token as a parameter to the `authenticate.sh` utility script.
```sh
# to override: .authenticate.sh <token> <expected_token>
if .authenticate.sh $HOOK_token uniqueT0ken; then   # the "token" query param must be the same as uniqueT0ken
        echo "You are in!"
else
        echo "Auth failed"
fi;
```

# How To Test
1. Clone the repo
2. Run `docker compose up` (this will start the `webhook` service, and others)
3. Invoke the following command ```curl -X POST localhost:8000/hooks/deploy -H "Content-Type: application/json" -d '{"tag":"asdf", "service":"asfd", "image":"asdf", "token":"IAmASecretToken"}'``` (notice the token param is the same as the `DEFAULT_AUTH_TOKEN` env var in the `docker-compose.yml`)
4. Invoke the command ```curl -X POST localhost:8000/hooks/heartbeat -H "Content-Type: application/json" -d '{"token":"AnOverridenToken"}'```, notice how the token is different this time, this is because the heartbeat.sh script overrode the token like so 
```sh
# .authenticate.sh <token> <expected_token>
if ./.authenticate.sh $HOOK_TOKEN AnOverridenToken; then
  main
fi
```

https://user-images.githubusercontent.com/15690606/215421278-97c69c58-a351-45ee-b0bf-df549cf5825f.mp4


sirinoks:
image
Reviewing code at 3 am...

sirinoks:
Seems to not be working..
image

416rehman:
> Seems to not be working.. 

Wow, this was so annoying to figure out.. The line separators are using Windows format I am assuming? those `^M` chars will make the `.sh` scripts unexecutable on Linux 😭
image

sirinoks:
After the latest update, it seems to only run the DB container
image

SerpentBytes:
I tested the PR locally on the macOS environment, and it's workings. 👍🏼

sirinoks (approved):
Works for me now

docker-compose.yml
```diff
@@ -15,5 +15,23 @@ services:
     environment:
       MYSQL_DATABASE: starchart
       MYSQL_USER: starchart
-      MYSQL_PASSWORD: starchart_password 
+      MYSQL_PASSWORD: starchart_password
       MYSQL_ROOT_PASSWORD: root_password
+
+  webhook:
+    # a docker in docker container that installs go, installs webhook, and runs it
+    image: docker:dind
```
humphd, line 23: I think we should write a Dockerfile and build an Alpine image that has:

1. Defines the ENVs for passing in the token
2. [Docker](https://wiki.alpinelinux.org/wiki/Docker).  We'll have to match the version of Docker in the host.
3. [webhook](https://alpine.pkgs.org/3.15/alpine-community-x86_64/webhook-2.8.0-r4.apk.html).  I would pin it to a version vs. use `latest`, so you don't get a surprise when they update and it breaks.
4. copies the deploy script and hooks.json into the image
5. runs the webhook server as its CMD, using port 8000
6. exposes 8000

It's probably worth creating a top-level `webhook/` directory, and you can put your Dockerfile in there, along with the script, JSON, etc.

I also wonder if it's worth doing this in `docker-compose.yml` (i.e., the development compose), or if we should create `production.yml` and do it there.

scripts/hooks/.authenticate.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+# This script checks if the $HOOK_token environment variable is equal to the $DEFAULT_AUTH_TOKEN environment variable.
```
humphd, line 2: When we do this in production, we'll use [Docker Secrets](https://docs.docker.com/engine/swarm/secrets/), which will mean the secret will be mounted in `/run/secrets/webhook_secret_token` (or a filename similar to that).  We'll have to compare the token we get to the value of this file.

You could probably do something [like this](https://stackoverflow.com/questions/11287861/how-to-check-if-a-file-contains-a-specific-string-using-bash).

scripts/hooks/deploy.sh
```diff
@@ -0,0 +1,33 @@
...
+    fi
+
+    if [ -z "${HOOK_service}" ]; then
+      echo "The service is not set. Skipping deployment."
+      exit 1
+    fi
+
+    echo "Updating service ${HOOK_service}..."
+
+    cmd="docker service update --image ${HOOK_image}:${HOOK_tag} ${HOOK_service}"
```
humphd, line 23: We better harden this so it's safe to run.  We can use something like `printf %q` to safely transform the values into something we can put in a command-line arg.  From the [printf man page](https://man7.org/linux/man-pages/man1/printf.1.html):

>        %q     ARGUMENT is printed in a format that can be reused as
>               shell input, escaping non-printable characters with the
>               proposed POSIX $'' syntax.

```sh
docker service update --image "$(printf '%q' "${HOOK_image}:${HOOK_tag}")" "$(printf '%q' "${HOOK_service}")"
```

Also, let's use `ALL_CAPS` for the environment variables.

scripts/hooks/deploy.sh
```diff
@@ -0,0 +1,33 @@
...
+    echo "Running command: ${cmd}"
+
+    eval $cmd
```
humphd, line 26: Either here, or in a follow-up, we should give some thought to what we'll do if this command fails.

scripts/hooks/hooks.yml
```diff
@@ -0,0 +1,19 @@
+- id: deploy
+  execute-command: ./deploy.sh
+  include-command-output-in-response: true
+  pass-environment-to-command:
+      - source: url
```
humphd, line 5: I think we should move all this to the body of the request, so it isn't available to read on the URL in transit.  Best to hide as much as we can via SSL.

humphd (changes requested):
Thanks for starting this.

.gitattributes
```diff
@@ -0,0 +1 @@
+*.sh    text eol=lf
```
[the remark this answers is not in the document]
416rehman: Agreed, I use Linux for all my development. Just wasn't sure if CRLF was needed for other reasons over here.

scripts/hooks/.authenticate.sh
416rehman: Okay I could do this with an order of precedence like this, ranging from highest priority to lowest.
1. token parameter/argument
2. Docker secret `DEFAULT_AUTH_TOKEN`
3. Container env var `DEFAULT_AUTH_TOKEN`

Also, If we want each script to have a different auth token, we can override the `DEFAULT_AUTH_TOKEN` secret/envVar by specifying the expected token as a parameter when calling the `.authenticate.sh` script, like `.authenticate.sh <providedToken> <expectedToken>`

scripts/hooks/deploy.sh
416rehman: Oh this is neat, I didn't know about the `printf %q`.

The environment variables are automatically created from the hooks.json file, the variables were lowercase there, hence the lowercase envVars. Will change them to uppercase in hooks.json

scripts/hooks/deploy.sh
416rehman: We could have some logic to notify an admin of a failed webhook run? Or write out the log for this run to a file and have it accessible through a URL? But that will likely get out of scope of the current pr so I'll leave this for now.

scripts/hooks/hooks.yml
416rehman: Agreed, having secrets in readable URLs irks me haha

docker-compose.yml
humphd: Sure, mounting the `hooks/` directory works.  What if you do:

```
webhook/
  Dockerfile
  hook/
    hooks.json
    *.sh
```

scripts/hooks/.authenticate.sh
humphd: In a follow-up, we could also look at doing something where we generate a secret using a public/private key so we don't have to worry about rotating.  We don't need that here, but it would be more secure in future.

scripts/hooks/deploy.sh
humphd: Docker swarm can rollback, which might be the right first move.

I think pinging an admin is out of scope for this script.  But we should make sure we "do the right thing" if the command fails and not just crash.

.gitattributes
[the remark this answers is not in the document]
416rehman: https://github.com/Seneca-CDOT/starchart/pull/57/files#diff-618cd5b83d62060ba3d027e314a21ceaf75d36067ff820db126642944145393eR1

docker-compose.yml
416rehman: Resolved. The new custom image expects a directory to be mounted to `/hooks` with the `hooks.yaml` file in its root.

Using YAML instead of JSON because its designed to be human readable, and we can have comments.

scripts/hooks/deploy.sh
416rehman: Brought this up in Slack, but `printf %q` is not POSIX compliant, therefore our Busybox alpine image will not be able to run scripts containing this.

https://www.shellcheck.net/wiki/SC3050

scripts/hooks/deploy.sh
416rehman: [Resolved](https://github.com/Seneca-CDOT/starchart/pull/57/files#diff-68418ea51a029f95dee04a42bea2ffced96d0a3bdd351427aa2e0bece3f80320R27). The command invocation now branches into 2 separate paths.

scripts/hooks/hooks.yml
416rehman: [Resolved.](https://github.com/Seneca-CDOT/starchart/pull/57/files#diff-609a3b67b49141c106c5e8c1326bdbca40532edf252cfc87fb7da5cde58af70bR6) All the params are now sourced from payload, instead of the url.

416rehman:
Refer to the updated *How To Test* instructions to test.

sirinoks:
@416rehman My output after reproducing steps
image
I can't tell if what I got is what was supposed to happen, or not.

webhook/Dockerfile
```diff
@@ -0,0 +1,37 @@
+# Exposes webhook on port 8000
```
humphd, line 1: Let's use apk to get the `webhook` package vs. building:

```dockerfile
# Exposes webhook on port 8000
# Mount host docker socket to /var/run/docker.sock
# Mount host webhook directory to /hooks, the hooks.yaml and scripts should be in this directory

FROM alpine:3.17@sha256:f271e74b17ced29b915d351685fd4644785c6d1559dd1f2d4189a5e851ef753a

ARG DOCKER_CLI_VERSION=20.10.21-r2
ARG WEBKOOK_VERSION=2.8.0-r12

WORKDIR /hooks

RUN apk add --no-cache \
    docker-cli=${DOCKER_CLI_VERSION} \
    webhook=${WEBKOOK_VERSION}

ENTRYPOINT ["/usr/local/bin/webhook"]
CMD ["-hooks", "hooks.yaml", "-verbose", "-hotreload", "-port", "8000"]

EXPOSE 8000
```

docker-compose.yml
```diff
@@ -15,5 +15,16 @@ services:
...
       MYSQL_USER: starchart
-      MYSQL_PASSWORD: starchart_password 
+      MYSQL_PASSWORD: starchart_password
       MYSQL_ROOT_PASSWORD: root_password
+
+  webhook:
+    build: ./webhook  # the directory with Dockerfile
+    container_name: webhook
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock # mount the docker socket so we can control docker from inside the container
```
humphd, line 25: Mount this Read-Only add `:ro` to the end like so:

```yml
- /var/run/docker.sock:/var/run/docker.sock:ro
```

humphd (changes requested):

humphd:
I like your suggestion of doing the safe-string-bash stuff in a follow-up.  Can you file an issue on that, so we can avoid doing it here?

sirinoks:
Can't test this, instructions unclear :(

416rehman:
> I like your suggestion of doing the safe-string-bash stuff in a follow-up. Can you file an issue on that, so we can avoid doing it here?

Created https://github.com/Seneca-CDOT/starchart/issues/74

All the request changes have been addressed as well.

humphd:
Filed #76 as a follow-up to secure this webhook.

webhook/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ARG WEBKOOK_VERSION=2.8.0-r12
+
+WORKDIR /hooks
+
+RUN apk add --no-cache \
+    docker-cli=${DOCKER_CLI_VERSION} \
+    webhook=${WEBKOOK_VERSION}
+EXPOSE 8000
+ENTRYPOINT ["/usr/bin/webhook"]
+CMD ["-hooks", "hooks.yaml", "-verbose", "-hotreload", "-port", "8000"]
```
humphd, line 16: When we do this in production, we'll have to remember not to use `-verbose` or `-hotreload`.  I get that they are useful while developing.

webhook/hooks/deploy.sh
```diff
@@ -0,0 +1,39 @@
...
+
+main() {
+    echo "Deploying..."
+
+    if [ -z "${HOOK_tag}" ]; then
+      echo "The tag is not set. Skipping deployment."
+      exit 1
+    fi
+
+    if [ -z "${HOOK_image}" ]; then
```
humphd, line 11: I'm not sure if we should allow this.  Giving the ability to specify arbitrary images seems dangerous to me.  I think we'll only update the main web service, so we can probably remove this until it's needed.

webhook/hooks/hooks.yaml
```diff
@@ -0,0 +1,23 @@
...
+        name: tag
+      - source: payload
+        name: image
+      - source: payload
+        name: service
+      - source: payload
+        name: token
+        envName: HOOK_TOKEN # this indicates this webhook will require authentication (handled in the script for this hook)
+
+- id: heartbeat
```
humphd, line 16: I think you can remove this.

humphd (changes requested):

sirinoks:
After a talk, adding some documentation notes for myself in the future:
@sirinoks 
This microservice will run in our deployment environment (where exactly?). It will listen for a call from GitHub actions for when we release a new version of our software (question - does it only apply to our web service, or is it applicable to all services we will be running?). Once it gets that call, it will start the deployment scripts that will basically replace the version of our current [website] with the new one that we just released.

It is also wrapped in a docker container, which allows for fast and universal activation

- What do the .env variables do?
- What is heartbeat?

humphd:
@416rehman do you want us to finish this for you, or are you going to do more?

humphd:
I just spoke with ITS, and it sounds like they want to run this webhook directly on the master node vs. in Docker.  I can take this over and convert to use the scripts in the VM directly.

webhook/Dockerfile
416rehman: Resolved in new commit.

webhook/hooks/deploy.sh
416rehman: Resolved. You can now specify the TARGET_IMAGE and TARGET_SERVICE inside the script, and the hook only provides the tag param now.

humphd:
> > @416rehman do you want us to finish this for you, or are you going to do more?
> 
> Sorry, been busy with classes, I implemented the changes you requested. Let me know if there is anything else.

OK, thank you.  I totally understand that you've been busy with school.  Keep focusing on that instead of this!

humphd:
Closing in favour of #400, where I'm going to finish this work. Thanks for getting this started and prototyping a solution!

PR #61, opened 2023-01-31
Add dns lib to app with tests

### LINKED ISSUE #27: Initial set up of Route 53 DNS module
This issue will initially set up `Route 53` DNS module to provide basic functionality for the app. 

- Configurability of AWS Route 53 SDK so that it can be used in multiple environments
- Ability to create, update, delete DNS records in hosted zone
- Unit tests for the module
- Add a mock environment for tests

Genne23v:
Resolves #27 

Adding DNS lib and tests. Test uses `moto` Docker container. 
Lib implements following methods 

- `createHostedZone(domain)` returns `hostedZoneId`
- `resetHostedZone(hostedZoneId)`  
- `createRecord(recordType, name, value, hostedZoneId)` returns `requestedId`
- `updateRecord(recordType, name, value, hostedZoneId)` returns `requestedId`
- `deleteRecord(recordType, name, value, hostedZoneId)` returns `requestedId`
- `listRecords(hostedZoneId)` returns an array of `ResourceRecordSet` 
- `getChangeStatus(requestedId)` returns `changeStatus` (i.e. `PENDING` | `INSYNC`)

Also environment can be configured with `AWS_ENDPOINT_URL` in `Route53Client`. If `AWS_ENDPOINT_URL` is not provided, `http://localhost:5050` is used to test against `moto` container.


docker-compose.yml
```diff
@@ -15,5 +14,11 @@ services:
...
       MYSQL_USER: starchart
-      MYSQL_PASSWORD: starchart_password 
+      MYSQL_PASSWORD: starchart_password
       MYSQL_ROOT_PASSWORD: root_password
+
+  dns:
+    container_name: route53_mock
+    image: motoserver/moto:4.1.1
+    ports:
+      - "5050:5000"
```
humphd, line 24: Let's use `5053` to give a nod to this being a Route53 instance.

docker-compose.yml
```diff
@@ -15,5 +14,11 @@ services:
     environment:
       MYSQL_DATABASE: starchart
       MYSQL_USER: starchart
-      MYSQL_PASSWORD: starchart_password 
+      MYSQL_PASSWORD: starchart_password
       MYSQL_ROOT_PASSWORD: root_password
+
+  dns:
```
humphd, line 20: Let's add a comment above this:

```yml
  # Mock Route53 server for testing, see:
  # http://docs.getmoto.org/en/latest/docs/server_mode.html
  #
  # AWS Endpoint URL: http://localhost:5053
  # Dashboard URL: http://localhost:5053/moto-api
  # Reset URL: POST to http://localhost:5053/moto-api/reset
```

app/lib/route53Client.ts
```diff
@@ -0,0 +1,11 @@
+import { Route53Client } from "@aws-sdk/client-route-53";
```
humphd, line 1: I'm not sure that there's value in separating this out into its own file, since you only ever use it in `dns-server.ts`.  Let's move this into that file.

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+      }
+    });
+  } catch (error) {
+    throw new Error(`Error while deleting hosted zone: ${error}`);
+  }
+};
+
+export const createRecord = async (
+  type: RecordType,
+  name: string,
```
humphd, line 70: Let's add a function to validate the `name` callers can pass.  Route53 allows [this set](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DomainNameFormat.html#domain-name-format-hosted-zones), but we can limit it a bit further.  I'd suggest:

```ts
// Only a-z, 0-9, and a hyphen - are valid in names
const isValidName = (name) => /^[a-z0-9-]+$/.test(name)
```

If a name other than this is given, we should throw.

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+          },
+        ],
+      },
+      HostedZoneId: hostedZoneId,
+    });
+    const response = await route53Client.send(command);
+
+    if (!response.ChangeInfo?.Id) {
+      throw new Error(`Could not create a record`);
+    } else {
```
humphd, line 98: No `else` after a `throw` or `return`:

```ts
if(!...) {
  throw ...
}
return ...
```

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+      return response.ChangeInfo?.Id;
+    }
+  } catch (error) {
+    throw new Error(
+      `Something went wrong while creating resource record: ${error}`
```
humphd, line 103: `error.message`

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+    );
+  }
+};
+
+export const updateRecord = async (
+  type: RecordType,
+  name: string,
```
humphd, line 110: Same comments as above: move `hostedZoneId` first, validate `name`

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+              ],
+            },
+          },
+        ],
+      },
+      HostedZoneId: hostedZoneId,
+    });
+    const response = await route53Client.send(command);
+
+    if (!response.ChangeInfo?.Id) {
```
humphd, line 136: Same comments as above

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+    }
+  } catch (error) {
+    throw new Error(
+      `Something went wrong while updating resource record: ${error}`
+    );
+  }
+};
+
+export const deleteRecord = async (
+  type: RecordType,
```
humphd, line 149: Same comments as above: move `hostedZoneId` first, validate `name`

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+          },
+        ],
+      },
+      HostedZoneId: hostedZoneId,
+    });
+
+    const response = await route53Client.send(command);
+
+    if (!response.ChangeInfo?.Id) {
+      throw new Error("Could not delete the record");
```
humphd, line 178: Same comments as above

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+    } else {
+      return response.ChangeInfo?.Id;
+    }
+  } catch (error) {
+    throw new Error(
+      `Something went wrong while deleting resource record: ${error}`
```
humphd, line 184: Same as above

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+    );
+  }
+};
+
+export const listRecords = async (
```
humphd, line 189: I don't think we need this.  Our database will be our query source.

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+};
+
+export enum RecordType {
+  A = "A",
+  AAAA = "AAAA",
+  CNAME = "CNAME",
+  TXT = "TXT",
+}
+
+export type HostedZone = {
```
humphd, line 236: I don't think you're using this, are you?

package.json
```diff
@@ -53,6 +53,7 @@
     "tiny-invariant": "^1.3.1"
   },
   "devDependencies": {
+    "@aws-sdk/client-route-53": "^3.259.0",
```
humphd, line 56: This is a dependency, not a dev dependency

package.json
```diff
@@ -81,6 +82,7 @@
     "eslint-config-prettier": "^8.6.0",
     "eslint-plugin-cypress": "^2.12.1",
     "happy-dom": "^8.1.5",
+    "jest": "^29.4.1",
```
humphd, line 85: cc @Eakam1007, do we have an issue on switching to Jest already?  Looks like some of it is happening here, but likely it will need more work to configure it fully.

test/dns-server.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  beforeAll(async () => {
+    hostedZoneId = await createHostedZone("starchart.com");
+  });
+
+  afterAll(() => resetHostedZone(hostedZoneId));
+
+  test("Hosted zone is created and hosted zone is returned", () => {
+    expect(hostedZoneId.length).toBeGreaterThan(0);
+    expect(/^[A-Z0-9]+/.test(hostedZoneId)).toBe(true);
```
humphd, line 23: I've never seen clear docs on the format of this string.  I'd suggest what you do with the length is enough.  Route53/moto could change the format in the future.  We don't really care what it is, just that we get something back.

test/dns-server.test.ts
```diff
@@ -0,0 +1,112 @@
...
+  });
+
+  test("Create a record in existing hosted zone but throw an error in non existing hosted zone", async () => {
```
humphd, line 26: Split this into two tests (i.e., test one thing per test).

test/dns-server.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      createRecord(
+        RecordType.A,
+        "sub2.starchart.com",
+        "192.168.0.2",
+        "WRONGHOSTEDZONEID"
+      )
+    ).rejects.toThrow();
+  });
+
+  test("Update a record in hosted zone but throw an error in non existing hosted zone", async () => {
```
humphd, line 45: Same comment throughout this file: one thing per test.  Prefer more and smaller test cases.

test/dns-server.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  test("Create a record in existing hosted zone but throw an error in non existing hosted zone", async () => {
+    const changeId = await createRecord(
+      RecordType.A,
+      "sub.starchart.com",
+      "192.168.0.1",
+      hostedZoneId
+    );
+
+    expect(changeId).toMatch(/\/change\/[A-Z0-9]+/);
```
humphd, line 34: According to the [docs](https://docs.aws.amazon.com/Route53/latest/APIReference/API_GetChange.html#API_GetChange_RequestSyntax), I think we should just check that it's a string greater than 1.

test/dns-server.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      updateRecord(
+        RecordType.A,
+        "sub.starchart.com",
+        "192.168.0.3",
+        "WRONGHOSTEDZONEID"
+      )
+    ).rejects.toThrow();
+  });
+
+  test("Delete a record in hosted zone but throw an error in non existing hosted zone", async () => {
```
humphd, line 64: We should add a test where we try to delete a non-existent record and another where we use an incorrect value (e.g., change one digit in the IP address).

vitest.config.ts
```diff
@@ -11,7 +11,10 @@ export default defineConfig({
     globals: true,
     environment: "happy-dom",
     setupFiles: ["./test/setup-test-env.ts"],
-    include: ["./app/**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}"],
+    include: [
```
humphd, line 14: cc @Eakam1007, this config needs work in follow-up.

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+    });
+  } catch (error) {
+    throw new Error(`Error while deleting hosted zone: ${error}`);
+  }
+};
+
+export const createRecord = async (
+  type: RecordType,
+  name: string,
+  value: string,
```
humphd, line 71: We need to do some validation on the value as well.  Depending on the record type, we'll expect different things:

If the `RecordType` is `A` (ipv4) or `AAAA` (ipv6) we could use https://www.npmjs.com/package/is-ip to check it.

If it's a CNAME, we should probably let Route53 decide if it's valid (there's no standard definition of what this should be afaik).  A string of length >= 1 is probably enough

For txt, we could again take any value that is length >= 1

cc @dadolhay for thoughts as well.

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+      return response.ResourceRecordSets;
+    }
+  } catch (error) {
+    throw new Error(
+      `Something went wrong while fetching resource record: ${error}`
+    );
+  }
+};
+
+export const getChangeStatus = async (requestId: string) => {
```
humphd, line 210: `requestId` should be `changeId` I think.  Also, does Route53 require you to strip the leading `change/` from the id or will it work if you use exactly what it responds with on the `ChangeInfo.Id`?

humphd (changes requested):
Nice work on this, @Genne23v!  I've done a first pass through the code and made some suggestions.  I'll read it again after you update.

docker-compose.yml
Genne23v: I like the suggestion!

docker-compose.yml
Genne23v: Will be added

app/lib/route53Client.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const route53Client = new Route53Client({
+  endpoint: process.env.AWS_ENDPOINT_URL || "http://localhost:5050",
+  region: process.env.AWS_REGION || "us-east-1",
+  credentials: {
+    accessKeyId: process.env.AWS_ACCESS_KEY_ID || "",
```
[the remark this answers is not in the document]
Genne23v: I see. I will plan it for future releases.

app/lib/route53Client.ts
Genne23v: Will do that

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
+import {
```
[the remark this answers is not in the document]
Genne23v: Will be updated

test/dns-server.test.ts
Genne23v: Will be updated

test/dns-server.test.ts
Genne23v: Will be addressed

app/lib/dns-server.ts
Genne23v: I thought there would be another interface to validate the input. But I would love to add the validation. Will be added.

app/lib/dns-server.ts
Genne23v: I don't think it works with `/change/` but I will confirm it. Thanks

humphd: If that's the case, we should strip the string prefix like you do with the hosted zone id result.

app/lib/dns-server.ts
humphd: We can validate in the front-end too, but this adds some extra ability to debug faulty data.

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+export const createHostedZone = async (domain: string) => {
+  try {
+    const command = new CreateHostedZoneCommand({
+      Name: domain,
+      CallerReference: new Date().toString(),
+    });
+    const response = await route53Client.send(command);
+
+    if (!response.HostedZone?.Id) {
+      throw new Error("Could create hosted zone upon request");
```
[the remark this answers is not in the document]
Genne23v: Will be updated

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    return response.HostedZone?.Id.replace("/hostedzone/", "");
+  } catch (error) {
+    throw new Error(`Error while creating hosted zone, ${error}`);
+  }
+};
+
+export const resetHostedZone = async (hostedZoneId: string) => {
```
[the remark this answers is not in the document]
Genne23v: Yes, it is only for the test. I don't think this is necessary operation even for the admin. I will move and handle accordingly.

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+  } catch (error) {
+    throw new Error(`Error while deleting hosted zone: ${error}`);
+  }
+};
+
+export const createRecord = async (
+  type: RecordType,
+  name: string,
+  value: string,
+  hostedZoneId: string
```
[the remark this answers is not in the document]
Genne23v: I agree

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+    });
+    const response = await route53Client.send(command);
+
+    if (!response.HostedZone?.Id) {
+      throw new Error("Could create hosted zone upon request");
+    }
+
+    return response.HostedZone?.Id.replace("/hostedzone/", "");
+  } catch (error) {
+    throw new Error(`Error while creating hosted zone, ${error}`);
```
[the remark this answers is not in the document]
Genne23v: I was trying to use `error.message` here. I had to write below due to Typescript error. 
```js
if (error instanceof Error){
      throw new Error(`Error while creating hosted zone, ${error.message}`);
    }
```
Then, I decided to remove it as I see the `error` itself is not complicated. Is there any better way to write this code?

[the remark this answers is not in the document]
humphd: https://fettblog.eu/typescript-typing-catch-clauses/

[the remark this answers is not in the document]
Genne23v: Thank you for the great resource!

package.json
Genne23v: oh, this is a mistake. I probably didn't notice that I was adding `-D` mindlessly. Will fix this.

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+  name: string,
+  value: string,
+  hostedZoneId: string
+): Promise<string> => {
+  try {
+    const command = new ChangeResourceRecordSetsCommand({
+      ChangeBatch: {
+        Changes: [
+          {
+            Action: "CREATE",
```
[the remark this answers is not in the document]
Genne23v: I agree that will be better code.

app/lib/dns-server.ts
```diff
@@ -0,0 +1,239 @@
...
+            },
+          },
+        ],
+      },
+      HostedZoneId: hostedZoneId,
+    });
+    const response = await route53Client.send(command);
+
+    if (!response.ChangeInfo?.Id) {
+      throw new Error(`Could not create a record`);
```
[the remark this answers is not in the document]
Genne23v: Will be updated

app/lib/dns-server.ts
Genne23v: Will address

app/lib/dns-server.ts
Genne23v: I see. I will remove it.

app/lib/dns-server.ts
Genne23v: No, I don't think it's necessary.

app/lib/dns-server.ts
Genne23v: I would like to combine all validation into one function and common error message for both `name` and `value`. Let me know if this is ok.

humphd: I wouldn't, they aren't the same.  Functions should do one thing when possible.  Easier to write tests that way, too.

Genne23v:
@humphd  I thoroughly checked the feedback and updated the code. Please let me know if you have further feedback. Thank you

app/lib/dns.server.ts
```diff
@@ -0,0 +1,193 @@
...
+  endpoint: process.env.AWS_ENDPOINT_URL || "http://localhost:5053",
+  region: process.env.AWS_REGION || "us-east-1",
+  credentials: {
+    accessKeyId: process.env.AWS_ACCESS_KEY_ID || "",
+    secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY || "",
+    sessionToken: process.env.AWS_SESSION_TOKEN,
+  },
+});
+
+export const createHostedZone = async (
```
humphd, line 19: Let's improve the TypeScript for this a bit.  Right now it's giving you untyped responses from the Route53 client. 
 We can fix that.

First, we'll pull in the proper type definition for the `CreateHostedZoneResponse`.  This requires us to `import type ...`.  Then we can type our `response` using `const response: CreateHostedZoneResponse = ...`.  This means that TS knows about the shape and properties of the response, so we can get rid of the `string | undefined` in the return value.

Finally, I've changed the way you were doing the `catch` so that we always return a value or throw.  We'll use proper logging later when we add it in to the server.

Here's how that will look:

```ts
import type { CreateHostedZoneResponse  } from "@aws-sdk/client-route-53";

// ...

export const createHostedZone = async (
  domain: string
) => {
  try {
    const command = new CreateHostedZoneCommand({
      Name: domain,
      CallerReference: new Date().toString(),
    });
    const response: CreateHostedZoneResponse = await route53Client.send(command);

    if (!response.HostedZone?.Id) {
      throw new Error("Missing hosted zone ID in AWS response");
    }
 
    return response.HostedZone.Id.replace("/hostedzone/", "");
  } catch (error) {
    // TODO: Use proper logging...
    console.warn(error);
    throw new Error(`Error while creating hosted zone`);
  }
};
```

We need to follow this pattern for all of the rest below.

app/lib/dns.server.ts
```diff
@@ -0,0 +1,193 @@
...
+};
+
+export const createRecord = async (
+  hostedZoneId: string,
+  type: RecordType,
+  name: string,
+  value: string
+): Promise<string | undefined> => {
+  try {
+    return await upsertRecord(hostedZoneId, type, name, value);
```
humphd, line 47: You don't need `async/await` on this.  You can return the Promise directly.

app/lib/dns.server.ts
```diff
@@ -0,0 +1,193 @@
...
+      throw new Error("Could not get ChangeIno for requested ID");
+    }
+    return response.ChangeInfo?.Status;
+  } catch (error) {
+    throw new Error(`Error occurred while getting change status: ${error}`);
+  }
+};
+
+const isNameValid = (name: string) => {
+  const subdomain = name.replace(".starchart.com", "");
```
humphd, line 170: We shouldn't have anything hard-coded in here.  We'll test the entire name (i.e., get rid of this line).

humphd (changes requested):

app/lib/dns.server.ts
```diff
@@ -0,0 +1,192 @@
+import {
+  Route53Client,
+  CreateHostedZoneCommand,
+  ChangeResourceRecordSetsCommand,
+  GetChangeCommand,
+} from "@aws-sdk/client-route-53";
+import { CreateHostedZoneResponse } from "@aws-sdk/client-route-53";
```
humphd, line 7: This is a `type` import vs. a module.  You should separate it below your module imports with a blank line, and use `import type` vs. `import`:

```ts
import {
  Route53Client,
  CreateHostedZoneCommand,
  ChangeResourceRecordSetsCommand,
  GetChangeCommand,
} from "@aws-sdk/client-route-53";

import type { CreateHostedZoneResponse } from "@aws-sdk/client-route-53";
```

Also, do you not need the other response types as well?  That is, for all the other command types you are executing?  I think the issue is that `.send(command)` doesn't return the specific type, and you have to set it when declaring the response.

app/lib/dns.server.ts
```diff
@@ -0,0 +1,192 @@
...
+    }
+  }
+};
+
+export const upsertRecord = async (
+  hostedZoneId: string,
+  type: RecordType,
+  name: string,
+  value: string
+): Promise<string | undefined> => {
```
humphd, line 64: You haven't followed the same pattern I showed you above with `CreateHostedZone` (i.e., how you do the error typing, using the specific type from AWS).  Can you please fix all of your methods in here to use the same pattern I showed you above?

humphd (changes requested):

app/lib/dns.server.ts
```diff
@@ -0,0 +1,200 @@
+import {
+  Route53Client,
+  CreateHostedZoneCommand,
+  ChangeResourceRecordSetsCommand,
+  GetChangeCommand,
+} from "@aws-sdk/client-route-53";
+import {
```
humphd, line 7: Use `import type` and also move this to its own separate line, below the other regular imports.

app/lib/dns.server.ts
```diff
@@ -0,0 +1,200 @@
...
+    const response: CreateHostedZoneResponse = await route53Client.send(
+      command
+    );
+
+    if (!response.HostedZone?.Id) {
+      throw new Error("Missing hosted zone ID in AWS response");
+    }
+    return response.HostedZone?.Id.replace("/hostedzone/", "");
+  } catch (error) {
+    if (error instanceof Error) {
```
humphd, line 39: Drop this if, get rid of `error.message` in line 41:

```ts
} catch(error) {
  // TODO: fix with better logging...
  console.warn(error);
  throw new Error(`Error while creating hosted zone`);
}
```

Please make this change in **all** code below.

app/lib/dns.server.ts
```diff
@@ -0,0 +1,200 @@
...
+      CallerReference: new Date().toString(),
+    });
+    const response: CreateHostedZoneResponse = await route53Client.send(
+      command
+    );
+
+    if (!response.HostedZone?.Id) {
+      throw new Error("Missing hosted zone ID in AWS response");
+    }
+    return response.HostedZone?.Id.replace("/hostedzone/", "");
```
humphd, line 37: Drop the `?` here, it's not needed, due to check in 34.

I already mentioned all this in previous reviews.

app/lib/dns.server.ts
```diff
@@ -0,0 +1,200 @@
...
+    throw new Error(`Error occurred while getting change status: ${error}`);
+  }
+};
+
+const isNameValid = (name: string) => {
+  return /^[a-z0-9-]+.[a-z]+.[a-z]+$/.test(name);
+};
+
+const isValueValid = (type: RecordType, value: string) => {
+  if (type === RecordType.A) {
```
humphd, line 183: Move your enum up in the file and/or change this to `'A'`

humphd (changes requested):

humphd:
@Genne23v before I re-review this, can you rebase on current `main`?

app/lib/dns.server.ts
humphd: This hasn't been done yet.  You shouldn't need to check the type of error, since we won't use the `error.message`.

humphd: You also need to do the same for all other methods in this file.

app/lib/dns.server.ts
```diff
@@ -0,0 +1,203 @@
...
+    if (error instanceof Error) {
+      // TODO: fix with better logging...
+      console.warn(error);
+      throw new Error(`Error occurred while getting change status: ${error.message}`);
+    }
+  }
+};
+
+const isNameValid = (name: string) => {
+  return /^[a-z0-9-]+.[a-z]+.[a-z]+$/.test(name);
```
humphd, line 189: It's valid if you include the trailing `.` as well (i.e., canonical name version).  Add a `\.?$` to the end

humphd (changes requested):

test/dns.server.test.ts
```diff
@@ -0,0 +1,187 @@
...
+  createHostedZone,
+  createRecord,
+  upsertRecord,
+  deleteRecord,
+  getChangeStatus,
+  RecordType,
+} from "../app/lib/dns.server";
+
+describe("DNS server lib function test", () => {
+  let hostedZoneId: string | undefined;
```
humphd, line 11: Let's use `string` vs. `string | undefined` which should allow you to get rid of the `!` on all your uses of `hostedZoneId` below.

test/dns.server.test.ts
```diff
@@ -0,0 +1,187 @@
...
+  afterAll(() => resetHostedZone());
+
+  test("Hosted zone is created and hosted zone is returned", () => {
+    expect(hostedZoneId?.length).toBeGreaterThan(0);
+    expect(/^[A-Z0-9]+/.test(hostedZoneId!)).toBe(true);
+  });
+
+  test("createRecord() creates a record in existing hosted zone and returns changeId", async () => {
+    const changeId = await createRecord(
+      hostedZoneId!,
```
humphd, line 26: You can drop the `!` here once you remove the `| undefined` on line 11.

Do the same throughout this file.

humphd (changes requested):
This is almost done!  One more TS improvement and we can land it.  Great work

humphd (changes requested):
Code looks great.

Prettier is unhappy:

```
Checking formatting...
[warn] docker-compose.yml
```

Can you fix?

humphd (approved):
👍 

Great work @Genne23v!

Genne23v:
> 👍
> 
> Great work @Genne23v!

Thank you! Must have been exhausting review. Sorry about that.

humphd:
Don't apologize, this was a big piece of code to get added, so it's natural to take so many iterations.

A few things we can learn from this:

You are always doing a single commit (probably you are amending?).  This is a good practice, but not during review.  After I read your code, and ask for changes, I want to see what you changed between one review and the next.  During the review process, you'd be better to add more commits, then squash at the end.

When you push new commits you want me to review, re-flag me for review so I know it's been updated without having to go and re-read it.

Make sure you do all the changes your reviewer requests before asking for another review.

I'm thrilled to see the Route53 stuff happening in 0.1.  Thanks for pushing so hard on this.

Genne23v:
> 

I felt that review would have been easier too if I keep the commits. I wasn't sure if I should keep it or not. But I learned the lesson, it will be a better review next time. Thank you for all the detailed comments and support.

humphd:
> > 
> 
> I felt that review would have been easier too if I keep the commits. I wasn't sure if I should keep it or not. But I learned the lesson, it will be a better review next time. Thank you for all the detailed comments and support.

You're doing great, it's a learning process!

humphd:
Rebase should fix the TypeCheck fail.

humphd (approved):

sfrunza13 (approved):
Looks good as far as I can tell. 👏

Genne23v:
> 
Thank you for the review!

Issue #34, opened 2023-01-25
Define Charka UI theme


Myrfion:
Since we are using Charka UI I think we will have to define a theme for it. One of the first things that we could put there could be the red colour as a primary colour as it was mentioned in wiki:
```
Use colorScheme='red' for ChakraUI components to match Seneca's red colour theme, see 
https://chakra-ui.com/docs/components/button/usage#usage
```
Charka also provides functionality to override styles of its own components like `Button` in the theme and in my opinion we should use theming as much as possible in order to keep all or almost all styles in one place. 

Read more on Charka UI theming here: https://chakra-ui.com/docs/styled-system/theme

humphd:
@Myrfion when you make a PR, you need to use "Fixes #34" in the description so it links the PR with the Issue, and auto-closes it.  I think #100 fixed this, correct?  If so, please close this.

Myrfion:
Fixed in #100

PR #75, opened 2023-01-31
Setup Chakra UI

### LINKED ISSUE #24: Setup Chakra UI for Remix
Building on top of #7, we would need to setup Chakra UI with Remix.
See [this Chakra UI docs page](https://chakra-ui.com/getting-started/remix-guide) for more information on setting up with Remix.

Eakam1007:
Fixes #24 

Added and setup Chakra UI. 
This does not involve the addition of a theme as there is a separate issue for it: #34 

Also changed the index page to use Chakra UI components (with increased font sizes) so that this can be tested:
- Before
  
  image

- After
  image

### Steps for testing
This assumes you already have starchart [setup locally](https://github.com/Seneca-CDOT/starchart/blob/41c1e750936c604fb797b87288dc04a36992b43f/CONTRIBUTING.md#development)

- `npm install`
- `npm run dev`
- navigate to `localhost:3000` and see that the app works and the index page has the updated styles


humphd (changes requested):
This looks good.

Can you modify something in the https://github.com/Seneca-CDOT/starchart/blob/main/app/routes/index.tsx to use ChakraUI Components?  Maybe switch to:

- https://chakra-ui.com/docs/components/heading
- https://chakra-ui.com/docs/components/text

That will make it easier for others to test that this works.

Myrfion (approved):
Agree with @humphd on adding some simple Charka components on `index.tsx` to show everybody else how it works, other than that LGTM

Eakam1007:
I changed the index page to use the components you linked. I also added some styles to change the font size so the change should be visible now

humphd (approved):

dadolhay (approved):

PR #78, opened 2023-02-01
feat: Add class for let's encrypt management

### LINKED ISSUE #17: Implement reusable code for node-acme let's encrypt account login
Since certificate generation will be a multi-step async process, we need an easy and reliable way of creating an "acme.client"

dadolhay:
Adds a class (to be extended later) that internally instantiates an acme client with the appropriate config.

Closes #17


CONTRIBUTING.md
```diff
@@ -32,7 +32,7 @@ Optional:
 $ cp .env.example .env
 
 # Start MySQL in docker
-$ docker compose up -d
+$ docker-compose up -d
```
humphd, line 35: Docker bundles `compose` into `docker` now, so you can skip using the Python version.

resources/LetsEncrypt.js
```diff
@@ -0,0 +1,38 @@
+import acme from "acme-client";
```
humphd, line 1: Let's write this in TypeScript, and move it to `app/lib/lets-encrypt.server.ts`, similar to what @Genne23v is doing in #27.

humphd (changes requested):

CONTRIBUTING.md
dadolhay: Interesting, under ubuntu 22.04 that command was unknown. I'll remove this change then

humphd: Not all Linux distros install this for some reason, see https://docs.docker.com/compose/install/linux/

resources/LetsEncrypt.js
Genne23v: I just learned that we should name the file in a certain way in Remix. In case, you don't find this link in my long threaded code review. [Remix docs on server code and bundles](https://remix.run/docs/en/v1/pages/gotchas#server-code-in-client-bundles)

humphd (changes requested):
This file needs to be in `app/lib/lets-encrypt.server.ts` so it's reachable via the rest of the Remix backend.

humphd:
Also, the lock file needs an update.

dadolhay:
> This file needs to be in `app/lib/lets-encrypt.server.ts` so it's reachable via the rest of the Remix backend.

Ah, I renamed it, created the folder and forgot to move... my bad, sorry
Rebased to lastest origin/main, should have no conflicts now

app/lib/lets-encrypt.server.ts
```diff
@@ -0,0 +1,39 @@
+import acme from "acme-client";
+import type { Client as AcmeClient } from "acme-client";
+import { secrets } from "docker-secret";
+
+const { LETSENCRYPT_ACCOUNT_PRIVATE_KEY_PEM } = secrets ?? {};
```
humphd, line 5: `LETS_ENCRYPT_...`

humphd (changes requested):
One last small thing, this is good after that.

app/lib/lets-encrypt.server.ts
dadolhay: Okay, updated

humphd (approved):

Genne23v (approved):
It looks good to go

PR #85, opened 2023-02-02
Add playwright

### LINKED ISSUE #25: Add playwright for end-to-end testing

Eakam1007:
Resolves #25 

- Removed cypress
- Updated scripts to remove cypress references
- Added playwright
- Updated e2e test scripts to run playwright
- Added basic test for the index page
- Added end-to-end test job to ci workflow
- Removed mocks from scripts and mocks folder (unused)

### Steps to run tests:
Assuming you already have project setup as per [CONTRIBUTING.md](https://github.com/Seneca-CDOT/starchart/blob/main/CONTRIBUTING.md#development):  
- `npm install`
- `npm run docker`
- `npm run test:e2e:run`


.github/workflows/ci.yaml
```diff
@@ -126,7 +126,7 @@ jobs:
         run: npm run docker
 
       - name: Seed database
-        run: npm run setup
+        run: DATABASE_URL=mysql://starchart:starchart_password@localhost:3306/starchart npm run setup
```
humphd, line 129: I think we should do this instead:

```yaml
run: |
  cp .env.example .env
  npm run setup
```

This way we can make sure that any other environment variables we expect will also be there.

package.json
```diff
@@ -21,12 +21,12 @@
     "db:studio": "prisma studio",
     "setup": "run-s db:generate db:push db:seed",
     "start": "cross-env NODE_ENV=production node ./build/server.js",
-    "start:mocks": "cross-env NODE_ENV=production node --require ./mocks --require dotenv/config ./build/server.js",
+    "start:mocks": "cross-env NODE_ENV=development node --require dotenv/config ./build/server.js dotenv_config_path=.env.example",
```
humphd, line 24: Let's not do this.  The dev should have their own env.

package.json
```diff
@@ -21,12 +21,12 @@
...
     "test": "vitest",
-    "test:e2e:dev": "start-server-and-test dev http://localhost:3000 \"npx cypress open\"",
+    "test:e2e:dev": "cross-env PORT=3000 start-server-and-test dev http://localhost:3000 \"playwright test\"",
```
humphd, line 26: We'll have to make this `8080` so it works with what's on `main`, see https://github.com/Seneca-CDOT/starchart/commit/3417e4a47fe4eaee2650537e7eeed736c1c9e0a7#diff-1ba718c1eb8aa39cd20c2562d92523068c734d75f54655e97d652b992d9b4259R57

playwright.config.ts
```diff
@@ -0,0 +1,107 @@
...
+    // {
+    //   name: 'Google Chrome',
+    //   use: {
+    //     channel: 'chrome',
+    //   },
+    // },
+  ],
+
+  /* Folder for test artifacts such as screenshots, videos, traces, etc. */
+  // outputDir: 'test-results/',
```
humphd, line 98: Why not include this?  Is this the default?

playwright.config.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  /* Run your local dev server before starting the tests */
+  // webServer: {
```
humphd, line 101: If you don't need to do this here, remove it.

humphd (changes requested):

playwright.config.ts
Eakam1007: This is the default that was auto generated. I haven't looked at it yet since I was trying to fix the ci

playwright.config.ts
Eakam1007: This was me testing a bunch of things. I will remove it

.github/workflows/ci.yaml
Eakam1007: Thanks for this! Didn't consider doing this

humphd (changes requested):
@Eakam1007 can you please squash these 37 commits into a single commit and rebase on current `main`?  We just updated the Prettier setup, which will affect this.

sfrunza13 (dismissed):
I think this looks good

playwright.config.ts
```diff
@@ -0,0 +1,95 @@
...
+  reporter: 'html',
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
+    actionTimeout: 0,
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    baseURL: `http://localhost:${process.env.PORT}`,
+
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: 'on-first-retry',
```
humphd, line 36: Add:

```
video: 'on-first-retry',
```

.github/workflows/ci.yaml
```diff
@@ -107,6 +107,71 @@ jobs:
             # cache node modules using the same key as restore.
           key: ${{ steps.node_modules-cache-restore.outputs.cache-primary-key }}
 
+  E2E-Test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    services:
+      mysql:
+        image: mysql
```
humphd, line 115: Will it let you specify an image tag?

```
image: mysql:8
```

That will match what we have in our setup.  If you can't, it's fine.

humphd (changes requested):
Two small things, and this is ready.  Awesome stuff.

.github/workflows/ci.yaml
```diff
@@ -107,6 +107,71 @@ jobs:
             # cache node modules using the same key as restore.
           key: ${{ steps.node_modules-cache-restore.outputs.cache-primary-key }}
 
+  E2E-Test:
+    timeout-minutes: 60
```
shawnyu5, line 111: This seems like a super long time out... Something like 5 to 10 mins seems more reasonable?

Eakam1007: 10 minutes seems a bit low to me. Maybe 15 minutes would be good?

shawnyu5: the default is 360 mins? Wow.... However long we set this if something goes wrong in the job and it decides to hang. it will wait that long before killing it, which means our CI will also hang for that amount of time. 10 to 15 ish mins sounds good I suppose

package.json
```diff
@@ -20,12 +20,12 @@
     "db:studio": "prisma studio",
     "setup": "run-s db:generate db:push db:seed",
     "start": "cross-env NODE_ENV=production node ./build/server.js",
-    "start:mocks": "cross-env NODE_ENV=production node --require ./mocks --require dotenv/config ./build/server.js",
+    "start:mocks": "cross-env NODE_ENV=production node --require dotenv/config ./build/server.js",
```
shawnyu5, line 23: What is the expected behavior of this? Currently I get `Unexpected Server Error` at `localhost:8080`, with these logs:

```
TypeError: (0 , import_jsx_dev_runtime.jsxDEV) is not a function
    at /home/shawn/starchart/app/entry.server.tsx:27:9
    at new Promise (<anonymous>)
    at handleRequest (/home/shawn/starchart/app/entry.server.tsx:21:10)
    at handleDocumentRequestRR (/home/shawn/starchart/build/server.js:39728:24)
    at requestHandler (/home/shawn/starchart/build/server.js:39591:22)
    at /home/shawn/starchart/build/server.js:40748:26
GET / 500 - - 65.736 ms
```

humphd: We don't even use these mocks, should we get rid of this?

Eakam1007: I don't get this error:
image

Could you please list the exact steps you did when you got this?

humphd (approved):
This is fantastic.  Nice sleuthing on the GitHub Actions Service Containers.

package.json
shawnyu5: 1. Check out your PR
2. run `npm i`
3. run `npm run start:mocks`
4. navigate to `locahost:8080`. See `Unexpected Server Error`
As well as see this in the terminal:
image

shawnyu5 (approved):
Looks great!

PR #92, opened 2023-02-02
feat: Create and recall let's encrypt orders, extract challenges

### LINKED ISSUE #18: Implement code that creates an "order" and retrieves "authorizations" with node-acme
We need to place an order with let's encrypt, and then retreive the "authorizations" or list of challenges that we need to complete After this is done, we can retreive the "keyauthorizations" for each challenge

Since this is an async proccess, we need to store both the "order", "authorizations" and each "keyauthorization" in the DB

The next step is to add the dns records to pass the challenges

Based on my research, all three object types should be easily serializable

dadolhay:
Closes #18

### DIFF
--- a/app/lib/lets-encrypt.server.ts
+++ b/app/lib/lets-encrypt.server.ts
@@ -71,9 +71,6 @@ class LetsEncrypt {
       if (!process.env.LETS_ENCRYPT_ACCOUNT_EMAIL)
         throw new Error('The env LETS_ENCRYPT_ACCOUNT_EMAIL is missing');
 
-      if (!process.env.LETS_ENCRYPT_ACCOUNT_EMAIL)
-        throw new Error('The env LETS_ENCRYPT_ACCOUNT_EMAIL is missing');
-
       this.#accountKey = LETS_ENCRYPT_ACCOUNT_PRIVATE_KEY_PEM;
       this.#directoryUrl = acme.directory.letsencrypt.production;
     } else {
@@ -113,8 +110,13 @@ class LetsEncrypt {
         ({ type }) => type === 'dns-01'
       ) as AcmeDnsChallenge;
 
+      if (!selectedChallenge)
+        throw new Error(
+          'The authorization object does not contain dns-01 type challenge. This should never happen.'
+        );
+
       // Get key for challenge
-      const keyAuthorization = await this.#client!.getChallengeKeyAuthorization(selectedChallenge!);
+      const keyAuthorization = await this.#client!.getChallengeKeyAuthorization(selectedChallenge);
 
       const challengeBundle: ChallengeBundle = {
         domain: `_acme-challenge.${authorization.identifier.value}`,
@@ -125,10 +127,17 @@ class LetsEncrypt {
     });
 
     this.#challengeBundles = await Promise.all(promises);
-
-    console.log(JSON.stringify(this.#challengeBundles));
   };
 
+  /**
+   * Create an order with let's encrypt using the account that was
+   * loaded in the initialize step
+   *
+   * @param {string} rootDomain The domain the certificate will be
+   * issued to (cert covers this root domain and one level of subdomains
+   * AKA. wildcard)
+   * @returns {Promise<this>} Promise of current object. Useful for chaining
+   */
   createOrder = async (rootDomain: string) => {
     if (!this.#client || !this.#account)
       throw new Error('You need to initialize the instance first');
@@ -146,6 +155,13 @@ class LetsEncrypt {
     return this;
   };
 
+  /**
+   * When we create an order, we get back a resource URL. We use that URL
+   * to recall the order from let's encrypt
+   *
+   * @param {string} url resource url of the order
+   * @returns {Promise<this>} Promise of current object. Useful for chaining
+   */
   recallOrder = async (url: string) => {
     if (!this.#client || !this.#account)
       throw new Error('You need to initialize the instance first');
--- a/app/lib/lets-encrypt.server.ts
+++ b/app/lib/lets-encrypt.server.ts
@@ -1,8 +1,35 @@
 import acme from 'acme-client';
-import type { Client as AcmeClient } from 'acme-client';
+import type {
+  Client as AcmeClient,
+  Order as AcmeOrder,
+  Authorization as AcmeAuthorization,
+} from 'acme-client';
+import type {
+  Account as AcmeAccount,
+  DnsChallenge as AcmeDnsChallenge,
+} from 'acme-client/types/rfc8555';
 import { secrets } from 'docker-secret';
 
+interface ChallengeBundle {
+  domain: string;
+  value: String;
+}
+
 const { LETS_ENCRYPT_ACCOUNT_PRIVATE_KEY_PEM } = secrets ?? {};
+
+/**
+ * Usage examples
+ *
+ * new LetsEncrypt()
+ *  .initialize()
+ *  .then((le) => le.createOrder("xyz.com"))
+ *  .then((le) => console.log(le.challengeBundles));
+ *
+ * new LetsEncrypt()
+ *  .initialize()
+ *  .then((le) => le.recallOrder(orderUrl))
+ *  .then((le) => console.log(le.challengeBundles));
+ */
 
 class LetsEncrypt {
   #client?: AcmeClient;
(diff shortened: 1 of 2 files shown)


app/lib/lets-encrypt.server.ts
```diff
@@ -29,8 +87,93 @@ class LetsEncrypt {
...
+     *
+     * Later, we can use the granted authorizations in our certificate
+     */
+    const promises = this.#authorizations.map(async (authorization) => {
+      const selectedChallenge = authorization.challenges.find(
+        ({ type }) => type === 'dns-01'
+      ) as AcmeDnsChallenge;
+
+      // Get key for challenge
+      const keyAuthorization = await this.#client!.getChallengeKeyAuthorization(selectedChallenge!);
```
humphd, line 117: Instead of `selectedChallenge!`, can we throw earlier if it's really missing?  I realize it's not likely, but given that it's possible, from the point of view of the types, we should deal with it.

humphd (changes requested):
This is looking really good.  I'm excited to see it coming together.

app/lib/lets-encrypt.server.ts
```diff
@@ -1,21 +1,79 @@
...
+   */
+  #registerReloadAccount = async () => {
+    // https://datatracker.ietf.org/doc/html/rfc8555#section-7.3.1
+
+    const account = await this.#client!.createAccount({
+      termsOfServiceAgreed: true,
+      contact: [`mailto:${process.env.LETS_ENCRYPT_ACCOUNT_EMAIL}`],
+    });
+
+    if (account.status !== 'valid') throw new Error(`Acme account is "${account.status}"`);
```
Genne23v, line 61: I didn't know `prettier` allows `if` statement without curly brackets. To me, line break after `if` is more readable.

Genne23v (changes requested):

app/lib/lets-encrypt.server.ts
```diff
@@ -1,21 +1,79 @@
...
+
   initialize = async () => {
     if (process.env.NODE_ENV === 'production') {
       if (!LETS_ENCRYPT_ACCOUNT_PRIVATE_KEY_PEM)
         throw new Error('The docker secret LETS_ENCRYPT_ACCOUNT_PRIVATE_KEY_PEM is missing');
 
+      if (!process.env.LETS_ENCRYPT_ACCOUNT_EMAIL)
+        throw new Error('The env LETS_ENCRYPT_ACCOUNT_EMAIL is missing');
+
+      if (!process.env.LETS_ENCRYPT_ACCOUNT_EMAIL)
```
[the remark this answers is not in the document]
dadolhay: Yep, thanks, I missed that

app/lib/lets-encrypt.server.ts
```diff
@@ -29,8 +87,93 @@ class LetsEncrypt {
...
+        domain: `_acme-challenge.${authorization.identifier.value}`,
+        value: keyAuthorization,
+      };
+
+      return challengeBundle;
+    });
+
+    this.#challengeBundles = await Promise.all(promises);
+
+    console.log(JSON.stringify(this.#challengeBundles));
```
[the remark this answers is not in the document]
dadolhay: I removed the console.

What type of test you think we should do? I think it would make most sense to do an integration test against a dockerized let's encrypt when we have those parts ironed out. If we wanted to do a unit test, we'd have to mock out / stub a large amount of api calls. Still possible, but less than optimal imo.

app/lib/lets-encrypt.server.ts
```diff
@@ -29,8 +87,93 @@ class LetsEncrypt {
...
+  };
+
+  createOrder = async (rootDomain: string) => {
```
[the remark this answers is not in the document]
dadolhay: Added standard doc block

app/lib/lets-encrypt.server.ts
```diff
@@ -29,8 +87,93 @@ class LetsEncrypt {
...
+        { type: 'dns', value: `*.${rootDomain}` },
+      ],
+    });
+
+    await this.#extractChallenges();
+
+    return this;
+  };
+
+  recallOrder = async (url: string) => {
```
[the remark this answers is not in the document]
dadolhay: added standard doc block

app/lib/lets-encrypt.server.ts
dadolhay: You are right. Updated!

app/lib/lets-encrypt.server.ts
dadolhay: I actually like to omit the code block for one liner statements that stop execution of the code (next, break, return, throw) I think it looks cleaner

app/lib/lets-encrypt.server.ts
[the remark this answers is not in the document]
humphd: I agree, we should do integration tests with the Pebble test container.  We'll do it in follow-ups vs. here.

app/lib/lets-encrypt.server.ts
humphd: When I was working with Mozilla on Firefox and Gecko, we used to regularly see bugs happen due to code like this:

```js
// Original code
if(whatever)
  doSomething();
```

Someone would make a change, like so:

```js
// Modified code
if(whatever)
  doSomething();
  doSomethingElse();
```

This obviously breaks, but only some of the time!  Mozilla's engineers got to the point where they decided it wasn't worth the hassle of having people accidentally break code by forgetting to wrap multi-line blocks.

The rule became, "always brace your ifs" and I've done it ever since.  This pattern will never break:

```js
// Original code
if(whatever) {
  doSomething();
}

// Modified code
if(whatever) {
  doSomething();
  doSomethingElse();
}
```

You'd think, "no one will ever make this mistake!"  But I've seen some of the best devs in the world do it.  As you get older, you start to appreciate simple code practices more.

app/lib/lets-encrypt.server.ts
```diff
@@ -29,8 +84,112 @@ class LetsEncrypt {
...
+     * to be passed for the authorization to be granted
+     *
+     * Later, we can use the granted authorizations in our certificate
+     */
+    const promises = this.#authorizations.map(async (authorization) => {
+      const selectedChallenge = authorization.challenges.find(
+        ({ type }) => type === 'dns-01'
+      ) as AcmeDnsChallenge;
+
+      if (!selectedChallenge)
```
humphd, line 113: For the future, we have a module in the Blue Stack to help with this: [tiny-invariant](https://github.com/alexreardon/tiny-invariant).  You'd use it like this:

```ts
import invariant from 'tiny-invariant';
...
invariant(selectedChallenge, 'The authorization object does not contain dns-01 type challenge. This should never happen.');
```

It's a nice utility to have, since this pattern happens all the time with TS.  After your call to `invariant(foo)` you can use `foo` without worry.

humphd (approved):
This looks good.  I've made a couple of suggestions, and agree with @Genne23v re: bracing-ifs.  I think the code is fine as is, too.  Your call.

Genne23v (approved):

app/lib/lets-encrypt.server.ts
dadolhay: I agree, and do exactly that for everything except for commands stopping the code execution (next, continue, break, return, throw).

If we have no strong convention, I'd keep using it as is, but if we decide that we want to prevent these kinds of errors, and braces are always mandatory, we should enforce it with an eslint rule:
https://eslint.org/docs/latest/rules/curly

As always, I'll yield to the convention if we decide to do this.

BTW, I'd also suggest adding this:
https://eslint.org/docs/latest/rules/no-console
In the very rare cases when we actually do want to use console, we can turn it off with `//eslint-disable-next-line no-console`

app/lib/lets-encrypt.server.ts
dadolhay: I have used invariant before, but does it also satisfy the typescript built in null check? I.e.,
```
if (!a) return;
  // I can use a after this without specifying a!
```
Does this hold true for invariant too?

humphd: Yes, it's perfect for this.

app/lib/lets-encrypt.server.ts
humphd: https://github.com/Seneca-CDOT/starchart/issues/102

app/lib/lets-encrypt.server.ts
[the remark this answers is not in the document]
dadolhay: Perfect, thank you!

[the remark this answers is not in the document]
humphd: Added Pebble for this in https://github.com/Seneca-CDOT/starchart/pull/103

PR #100, opened 2023-02-03
Initialize Chakra Theme

Myrfion:
### What I did?
I created the folder `theme` where we gonna have all theme-related files. For now, it has `index.ts` which imports all other files and initializes the theme itself. Another file is `colors.ts` where I added the Seneca colours in spectra from 100 to 900 and that colour I called `brand` (this is how chakra does it in the example in the doc so probably this is the way to go in terms of naming). I also created the button on `index.ts` just to show how the component that uses the theme looks like 

### How to test?
Just run the app locally and check the front end, you should see the button with very familiar color


humphd:
@Myrfion can you please rebase on `main` to get rid of the conflict you have, and also post a screenshot?

Myrfion:
### Log in screen (button has a brand color scheme): 
image


### Log out screen (button has a brand color scheme): 
image

@humphd here it is

app/theme/index.ts
```diff
@@ -0,0 +1,6 @@
+import { extendTheme } from '@chakra-ui/react';
+import colors from './colors';
+
+const theme = extendTheme({ colors });
```
humphd, line 4: Should we set the default `colorTheme` prop for Buttons, etc. as well (i.e., so you don't have to remember to do it each time)?  See https://github.com/chakra-ui/chakra-ui/discussions/2840

humphd (approved):

SerpentBytes:
@Myrfion I was able to test it locally, and it works. Looks good to me. 👍🏼

SerpentBytes (dismissed):

app/theme/index.ts
Eakam1007: I agree, this would be really helpful

Eakam1007 (dismissed):
Could look into setting a default color theme for different components, but looks good

app/theme/index.ts
humphd: We can do it in a follow-up, @Myrfion/@Ririio if one of you wants to file it.

app/theme/components.ts
```diff
@@ -0,0 +1,14 @@
+const components = {
```
humphd, line 1: Should we apply it across the board? https://chakra-ui.com/docs/styled-system/customize-theme#theme-extension-withdefaultcolorscheme

Eakam1007: These colors are some variation of red. Do we want all our components to have a theme based on red? 
We could also use the `withDefaultColorScheme` syntax but only for specific components. I think that is cleaner than the current syntax

Myrfion: Just made an update by applying it the way @humphd mentioned. Since we are using red as our primary colour, it makes sense to apply it to all components at once.

Eakam1007 (approved):
Looks good

humphd (approved):
This is great

PR #122, opened 2023-02-03
Switch from Pino to Winston for logging

humphd:
Pino doesn't want to work properly in Remix with esbuild.  So I've switched to using Winston instead.


SerpentBytes (approved):
This seems to fix the problem previously caused by `pino` as the `dev` script runs successfully, and I am able to navigate to `localhost:8080`.

dadolhay (approved):

Eakam1007 (approved):

PR #125, opened 2023-02-04
Configure Nodemailer transport with OAuth2

### LINKED ISSUE #33: Configure nodemailer
We will use nodemailer to send users notifications and need to configure it. It should be configured to use in different environments (product, staging, development etc)

SerpentBytes:
## Description
This PR resolves #33 by configuring Nodemailer transport using the `Outlook365` service provider. 

### Additional Information
- Since in Transport `options` I am using the `service` property, as opposed to `host`, I do not have to specify `port` and `secure` props.
- Nodemailer does not have many examples using Outlook/Exchange, so I am not sure if I have the suitable properties specified in the `auth` object. We need `user` which would be the email address we would use to send notification emails to our clients
- Looking at [Microsoft Documentation](https://learn.microsoft.com/en-us/exchange/client-developer/legacy-protocols/how-to-authenticate-an-imap-pop-smtp-application-by-using-oauth#get-an-access-token), we would at the very least require `CLIENT_ID` and `ACCESS_TOKEN`.


.env.example
```diff
@@ -7,7 +7,14 @@ SESSION_SECRET="super-duper-s3cret"
...
-
 # One of: trace, debug, info, warn, error, fatal, silent
 LOG_LEVEL=debug
+
+PORT=8080
+
+# Nodemailer configuration
+NODEMAILER_USER="no-reply@senecacollege.ca"
+
+NODEMAILER_CLIENT_ID="secret-client-id"
```
humphd, line 18: The client id and access token should be secrets we pull from Docker engine.  Take a look at how @dadolhay did it for Let's Encrypt in https://github.com/Seneca-CDOT/starchart/blob/main/app/lib/lets-encrypt.server.ts#L18.  You can do the same here.

This will read the values at runtime from a file that is mounted into the container by Docker.

app/lib/nodemailer.server.ts
```diff
@@ -0,0 +1,21 @@
+import { createTransport } from 'nodemailer';
```
humphd, line 1: I wonder if we should rename this to `app/lib/notifications.server.ts` and export a `send()` method only.  This way we can keep the details about nodemailer hidden from the rest of the code.

Basically, we'd take what you're doing here and add:

```ts
// https://nodemailer.com/usage/
export default function send(data) {
  const transport = initializeTransport();
  logger.debug('Sending notification', data);
  return transport.send(data);
}
```

humphd (changes requested):
This is looking good.  I think we're going to have to test this on staging to really understand if it works properly, but the code seems sane.

I've made an API suggestion.

app/lib/nodemailer.server.ts
SerpentBytes: Agreed. We shouldn't expose `initializeTransport.`.

app/lib/notifications.server.ts
```diff
@@ -0,0 +1,35 @@
...
+    throw new Error('Missing Nodemailer environment variables, skipping nodemailer configuration');
+  }
+  const transporter = createTransport({
+    service: 'Outlook365',
+    auth: {
+      type: 'OAuth2',
+      user: NODEMAILER_USER, // email address for notifications
+      clientId: NODEMAILER_CLIENT_ID,
+      accessToken: NODEMAILER_ACCESS_TOKEN,
+      expires: 1484314697598,
```
shawnyu5, line 19: What time format is this? Epoch? Seems like epoch translates into `Thu, 23 Jan 49006 01:13:18`.

SerpentBytes: This is the "default" expiration time in milliseconds. I should give it a more realistic value. `expires` accepts value in ms.

shawnyu5: ahh okay make sense

.env.example
```diff
@@ -7,7 +7,10 @@ SESSION_SECRET="super-duper-s3cret"
 # https://letsencrypt.org/docs/expiration-emails/
 LETS_ENCRYPT_ACCOUNT_EMAIL="nx@senecacollege.ca"
 
-PORT=8080
```
shawnyu5, line 10: Why remove this? Is this no longer needed?

.env.example
```diff
@@ -7,7 +7,10 @@ SESSION_SECRET="super-duper-s3cret"
...
-
 # One of: trace, debug, info, warn, error, fatal, silent
 LOG_LEVEL=debug
+
+PORT=8080
```
SerpentBytes, line 13: @shawnyu5, it's here.

shawnyu5: Oh my bad

SerpentBytes: Accidentally deleted it but added it back in one of the commits.

.env.example
SerpentBytes: see my other comment.

shawnyu5 (approved):

app/lib/notifications.server.ts
```diff
@@ -0,0 +1,35 @@
+import { createTransport, type SendMailOptions } from 'nodemailer';
+import { secrets } from 'docker-secret';
+import logger from './logger.server';
+
+const { NODEMAILER_USER } = process.env;
+const { NODEMAILER_CLIENT_ID, NODEMAILER_ACCESS_TOKEN } = secrets ?? {};
+
+const initializeTransport = () => {
+  if (!(process.env.NODEMAILER_USER && NODEMAILER_CLIENT_ID && NODEMAILER_ACCESS_TOKEN)) {
+    throw new Error('Missing Nodemailer environment variables, skipping nodemailer configuration');
```
dadolhay, line 10: nit: 
```suggestion
  if (!NODEMAILER_USER || !NODEMAILER_CLIENT_ID || !NODEMAILER_ACCESS_TOKEN) {
      throw new Error('Missing Nodemailer environment variables or docker secrets. Skipping nodemailer configuration');
```

dadolhay (approved):

dadolhay (approved):

app/lib/notifications.server.ts
```diff
@@ -0,0 +1,37 @@
...
+    );
+  }
+  const transporter = createTransport({
+    service: 'Outlook365',
+    auth: {
+      type: 'OAuth2',
+      user: NODEMAILER_USER, // email address for notifications
+      clientId: NODEMAILER_CLIENT_ID,
+      accessToken: NODEMAILER_ACCESS_TOKEN,
+      expires: 1484314697598,
```
shawnyu5, line 21: Should this be lowered to something saner?

SerpentBytes: Would 3 or 5 days be more appropriate?

SerpentBytes: Scaled down to 60hrs, for now.

humphd (changes requested):
Looks good.  One small issue to fix.

I'm not sure about the expiry on the token.  It will likely be dictated by the server.  However, we won't hold a transporter for a long time (i.e., just long enough to send), so getting a new one each time is fine.

SerpentBytes:
> Looks good. One small issue to fix.
> 
> I'm not sure about the expiry on the token. It will likely be dictated by the server. However, we won't hold a transporter for a long time (i.e., just long enough to send), so getting a new one each time is fine.

Removed the optional `expires` field is the latest commit.

humphd (approved):

dadolhay (dismissed):

humphd (dismissed):

humphd:
Delete your lock file, run `npm install` again, and push up another commit.

humphd (changes requested):
This doesn't seem right.  You're breaking tests all of a sudden, and have downgraded the node types version.  Can you rebase again and fix your lock file versions?

package.json
```diff
@@ -30,11 +30,11 @@
     "prepare": "husky install"
   },
   "dependencies": {
+    "@aws-sdk/client-route-53": "^3.261.0",
```
shawnyu5, line 33: Please avoid making cosmetic changes to files.

.env.example
shawnyu5, line 13: Same here. Add the new stuff to this file. Do not move things around

package.json
dadolhay: I disagree, npm alphabetizes these, so we should follow the convention. Plus, it gets auto moved in certain cases.

dadolhay (approved):

humphd (approved):

PR #128, opened 2023-02-04
ci: cache playwright browser

### LINKED ISSUE #120: ci: cache playwright browser
Every action run should not install the playwright browser. That takes way too long. Look into how we can cache the browser to make the CI job run faster.

shawnyu5:
Closes #120

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: ['main']
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   Build:
     permissions: write-all
@@ -140,12 +144,13 @@ jobs:
         id: playwright-version
         run: echo "PLAYWRIGHT_VERSION=$(node -e "console.log(require('./package-lock.json').dependencies['@playwright/test'].version)")" >> $GITHUB_ENV
 
-      - name: Cache playwright binaries
-        uses: actions/cache@v3
-        id: playwright-cache
+      - name: Restore Playwright browser cache
+        id: playwright-cache-restore
+        uses: actions/cache/restore@v3
         with:
-          path: |
-            ~/.cache/ms-playwright
+          # path to playwright browser
+          path: ~/.cache/ms-playwright
+          # cache browser with playwright version
           key: ${{ runner.os }}-playwright-${{ env.PLAYWRIGHT_VERSION }}
 
       - name: Install Playwright's dependencies only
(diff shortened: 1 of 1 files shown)


shawnyu5:
It is possible to cache the playwright browser it self, however the browser requires system dependencies, that can not be cached easily. Refer to this run here: https://github.com/shawnyu5/starchart/actions/runs/4093992063

The alternative is to install the browser it self through `npx`, and system dependencies through the package manager. Which still takes too long. Refer to this run here: https://github.com/shawnyu5/starchart/actions/runs/4093992063

Going to experiment with using a docker container with playwright installed to run the tests

shawnyu5:
Hmm, using [this](https://hub.docker.com/_/microsoft-playwright?tab=description) docker container, that contains playwright to run our CI means we would need to install docker inside the container every time. Which also involves installing stuff, and is no better than installing playwright it self...

shawnyu5:
Since the CI run can take a little bit of time, a `concurrency` group is added to cancel previous workflow runs if a new workflow is triggered.

For example, if 2 commits were pushed in close succession. Rather than kicking off 2 separate workflow runs, the first workflow run would be cancelled in favor of the second run. Refer to these 2 workflow runs:

First commit triggering a workflow run, cancelled by the next commit: https://github.com/Seneca-CDOT/star chart/actions/runs/4094766475

Second commit, pushed right after the previous commit. Triggering another workflow run, cancelling the first: https://github.com/Seneca-CDOT/star chart/actions/runs/4094767877

shawnyu5:
The CI run will attempt to restore playwright browser from cache. If it is successful, it will install system dependencies. If its unsuccessful, it will install playwright browser it self along with its dependencies

.github/workflows/ci.yaml
```diff
@@ -58,7 +60,6 @@ jobs:
           prettier_options: --check .
 
   Unit-Tests:
-    permissions: write-all
```
Eakam1007, line 61: Just curious, what did this code do before? Is this not needed?

Eakam1007 (dismissed):
Looks good! It is really nice to see faster tests

Just wondering; maybe this could be addressed in another issue; is it possible to cache the dependencies as well?

.github/workflows/ci.yaml
shawnyu5: This was needed with the weird hack of canceling parallel jobs. Forgot to clean it up in the other PR.

shawnyu5:
The dependencies are system dependencies from what I understand, it's not a single binary, like the browser, or a single directory like `node_modules`. So I dont think its fully possible to cache those.

shawnyu5:
If you look at the logs [here](https://github.com/Seneca-CDOT/starchart/actions/runs/4100467684/jobs/7071323094#step:9:1), the dependencies are being installed through the package manager... I don't know if it's possible to cache those...

shawnyu5:
Cleaned up some commit history

Eakam1007 (approved):

humphd (approved):
Very nice!

PR #133, opened 2023-02-05
Dependabot: ignore major updates

### LINKED ISSUE #33: Configure nodemailer

### LINKED ISSUE #71: Update Dependabot settings to ignore Major version updates
## Description
Dependabot helps scan for vulnerabilities in the node packages used by Starchart. It's configured to run a scan and raise  PRs, which requires manual review and merge. Currently, it's also doing `major` updates, which is risky and should be handled by developers. We only want Dependabot to do `minor` or `patch` level version updates in PR.

SerpentBytes:
## Description
This PR resolves #71 by ignoring `major` version updates. Dependabot will be restricted from raising PR for riskier updates, such as migrating from version `0.X.X` to `2.X.X` for any dependency. 

#### Resource
- https://github.blog/changelog/2021-05-21-dependabot-version-updates-can-now-ignore-major-minor-patch-releases/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,3 +16,7 @@ updates:
     open-pull-requests-limit: 5
     # disable auto rebasing
     rebase-strategy: 'disabled'
+
+    ignore:
+      - dependency-name: '*'
+        update-types: ['version-update:semver-major']


humphd (approved):
I think this is sane, and should save us some grief.  We can do our own major version bumps manually.  Thanks for this.

sfrunza13 (approved):
Good idea

PR #152, opened 2023-02-07
Update Nodemailer production transport options for testing

### LINKED ISSUE #149: Modify Nodemailer module for testing in production environment
## Description

@humphd on Slack
 >we should modify your code a bit so that it can accept OAuth2 info or username/password in production, and for now we'll use the latter

Update the Nodemailer module production code to accept username/password for the initial testing of the current notification functionality.

SerpentBytes:
## Description
This PR resolves #149 by updating the `production` transport options with the following settings:
- host: `smtp.office365.com`
- port: `587`
- auth (accepts username, pass)
- requireTLS: `true  // forces STARTTLS`

### DIFF
--- a/app/lib/notifications.server.ts
+++ b/app/lib/notifications.server.ts
@@ -5,26 +5,26 @@ import logger from './logger.server';
 import type { SendMailOptions } from 'nodemailer';
 
 const { NOTIFICATIONS_EMAIL_USER, NODE_ENV, MAILHOG_SMTP_PORT } = process.env;
-const { NODEMAILER_CLIENT_ID, NODEMAILER_ACCESS_TOKEN } = secrets ?? {};
+const { NOTIFICATIONS_USERNAME, NOTIFICATIONS_PASSWORD } = secrets ?? {};
 
 const initializeTransport = () => {
   if (!NOTIFICATIONS_EMAIL_USER) {
     throw new Error('Missing Nodemailer user. Skipping nodemailer configuration');
   }
   if (NODE_ENV === 'production') {
-    if (!NODEMAILER_CLIENT_ID || !NODEMAILER_ACCESS_TOKEN) {
+    if (!NOTIFICATIONS_USERNAME || !NOTIFICATIONS_PASSWORD) {
       throw new Error(
-        'Missing Nodemailer environment variables or docker secrets. Skipping nodemailer configuration'
+        'Missing NOTIFICATIONS_USERNAME and/or NOTIFICATIONS_PASSWORD env variables. Skipping nodemailer configuration'
       );
     }
     return createTransport({
-      service: 'Outlook365',
+      host: 'smtp.office365.com',
+      port: 587,
       auth: {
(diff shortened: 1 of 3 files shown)


app/lib/notifications.server.ts
```diff
@@ -4,27 +4,28 @@ import logger from './logger.server';
...
-    if (!NODEMAILER_CLIENT_ID || !NODEMAILER_ACCESS_TOKEN) {
+    if (!NOTIFICATIONS_USERNAME || !NOTIFICATIONS_PASSWORD) {
       throw new Error(
         'Missing Nodemailer environment variables or docker secrets. Skipping nodemailer configuration'
       );
     }
     return createTransport({
-      service: 'Outlook365',
+      host: 'smtp.office365.com',
+      port: Number(NOTIFICATIONS_SMTP_PORT || '587'),
```
humphd, line 23: This will always be `587` in production, let's not bother with a variable to configure it.

app/lib/notifications.server.ts
```diff
@@ -4,27 +4,28 @@ import logger from './logger.server';
...
 
 const initializeTransport = () => {
   if (!NOTIFICATIONS_EMAIL_USER) {
     throw new Error('Missing Nodemailer user. Skipping nodemailer configuration');
   }
   if (NODE_ENV === 'production') {
-    if (!NODEMAILER_CLIENT_ID || !NODEMAILER_ACCESS_TOKEN) {
+    if (!NOTIFICATIONS_USERNAME || !NOTIFICATIONS_PASSWORD) {
       throw new Error(
         'Missing Nodemailer environment variables or docker secrets. Skipping nodemailer configuration'
```
humphd, line 18: I would improve this error message: tell them which variables are actually missing, so it's easier to debug startup crashes.

.env.example
```diff
@@ -14,4 +14,5 @@ LOG_LEVEL=debug
 
 # Nodemailer config
 NOTIFICATIONS_EMAIL_USER="no-reply@senecacollege.ca"
+NOTIFICATIONS_SMTP_PORT=587
```
humphd, line 17: We probably don't need this, see my comment later.

humphd (changes requested):
Nice, this is looking good.  A few small things.

SerpentBytes:
> Nice, this is looking good.  A few small things.

I was thinking of expanding this PR by implementing tests that would work with MailHog. That way, we would know the current configuration of nodemailer is working as intended.

dadolhay (approved):

sfrunza13 (dismissed):

humphd:
@SerpentBytes looks like this is good to merge.

SerpentBytes:
> @SerpentBytes looks like this is good to merge.

merging is blocked despite have 2 approvals.

shawnyu5 (approved):

humphd (approved):

humphd:
Should be good now, I'm not sure why it blocked on my review.

PR #154, opened 2023-02-07
feat: Implement authoritative DNS check for acme challenge

### LINKED ISSUE #20: Implement code that verifies a challenge
We need to poll the dns server and wait until the new TXT records are propagated, then notify let's encrypt about the challenge completion.

It the answer is positive, we can mark the specific challenge as completed in the database.

Once a challenge been verified, the appropriate DNS record can be removed.

When all challenges for the requests are done, we can do the next step and actually create the certificate

dadolhay:
Closes #20

### DIFF
--- a/app/lib/lets-encrypt.server.ts
+++ b/app/lib/lets-encrypt.server.ts
@@ -1,3 +1,5 @@
+import dnsPromises from 'dns/promises';
+import type { Resolver as PromisesResolver } from 'dns/promises';
 import acme from 'acme-client';
 import type {
   Client as AcmeClient,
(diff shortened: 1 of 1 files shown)


app/lib/lets-encrypt.server.ts
```diff
@@ -10,6 +12,62 @@ import type {
...
+  let found = false;
+
+  // loop until we find a soa, each time drop a subdomain from `domainName`
+  while (domainName.includes('.') && !found) {
+    try {
+      await dnsPromises.resolveSoa(domainName);
+      found = true;
+    } catch (e) {
+      // remove one level, i.e., c.b.a.com => b.a.com
+      domainName = domainName.replace(/^.*?\./, '');
```
humphd, line 31: Why the `?` after `*` here?

Would it not be better to use `/^[^.]*\./`?

app/lib/lets-encrypt.server.ts
```diff
@@ -10,6 +12,62 @@ import type {
...
+    }
+  }
+
+  if (!found) {
+    // We have reached the TLD, resolution failed
+    throw new Error(
+      `No authoritative DNS server found for ${incomingDomain} or any of it's parents`
```
humphd, line 38: s/it's/its/

humphd (changes requested):
This looks pretty good.  A few questions and small things to fix.

Another question I have in general: there's a lot of places this code can/will throw.  Are we going to leave-out exception handling for most internal calls, and leave it to the caller to deal with?  Is there anything in this code you want to log (debug logs) or errors we should be surfacing with more meaningful error messages?

app/lib/lets-encrypt.server.ts
dadolhay: The ? makes * to not be "greedy", so it will match as few characters as possible, making the subsequent \. be the first dot in the string (if there was no `?` * would gulp up everything until the last '.' in the domain).

The effect is the same to what you suggested

app/lib/lets-encrypt.server.ts
dadolhay: Updated, thanks for catching it!

dadolhay:
> This looks pretty good. A few questions and small things to fix.
> 
> Another question I have in general: there's a lot of places this code can/will throw. Are we going to leave-out exception handling for most internal calls, and leave it to the caller to deal with? Is there anything in this code you want to log (debug logs) or errors we should be surfacing with more meaningful error messages?

If anything goes wrong, we do want this to fail causing the queue job to restart, and the exceptions contain meaningful error messages that can be caught and re-thrown by the logic of the queue worker.

To help with maintainability, I'm trying to limit the scope of this class, to only work with let's encrypt, and have no knowledge at all about SQL or the queue system.

At the moment I don't see the need to add additional debug logging, but if you think there are places that we would benefit on keeping an extra eye on, I'd be happy to add them (error handlig and logging should still be only happening in the queue worker code)

app/lib/lets-encrypt.server.ts
humphd: Sounds good, thanks for the explanation.

humphd:
> At the moment I don't see the need to add additional debug logging, but if you think there are places that we would benefit on keeping an extra eye on, I'd be happy to add them (error handlig and logging should still be only happening in the queue worker code)

OK, let's continue with this approach, and keep an eye on things as we add tests.  Layering in logging later is often much harder.

app/lib/lets-encrypt.server.ts
```diff
@@ -46,6 +104,27 @@ class LetsEncrypt {
...
+    key,
+  }: {
+    domain: string;
+    key: string;
+  }): Promise<boolean> => {
+    const resolver = await getAuthoritativeResolverForDomain(domain);
+
+    const txtRecords = (await resolver.resolveTxt(domain))
+      // Flatten array of arrays
+      .reduce((acc, txtArr) => [...acc, ...txtArr], []);
```
humphd, line 123: Same question here, can we use `.flat()` instead to improve clarity?

app/lib/lets-encrypt.server.ts
```diff
@@ -10,6 +12,62 @@ import type {
...
+ * @returns {Promise<Resolver>} domain
+ */
+const getAuthoritativeResolverForDomain = async (domain: string): Promise<PromisesResolver> => {
+  const hostedZone = await getHostedZoneForDomain(domain);
+
+  const nsRecords = await dnsPromises.resolveNs(hostedZone);
+  // Get all ipv4 addresses for all ns records (there may be multiple ipv4 for each NS)
+  const nsAddrIpv4Arr = (await Promise.all(nsRecords.map((nsName) => dnsPromises.resolve4(nsName))))
+    // Flatten array of arrays
+    .reduce((acc, ipv4Arr) => [...acc, ...ipv4Arr], []);
```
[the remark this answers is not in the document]
dadolhay: Hm, I actually didn't know about that, thank you!

humphd (approved):

sfrunza13 (approved):

PR #155, opened 2023-02-08
fix(ci): e2e test timeout len and cache name

shawnyu5:
- Set timeout Len for E2E test to 30 mins
- Fix the key name which playwright browser is cached under

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -100,7 +100,7 @@ jobs:
           key: ${{ steps.node_modules-cache-restore.outputs.cache-primary-key }}
 
   E2E-Tests:
-    timeout-minutes: 15
+    timeout-minutes: 30
     runs-on: ubuntu-latest
     services:
       mysql:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -140,8 +140,7 @@ jobs:
       - name: Get installed Playwright version
         id: playwright-version
         run: |
-          echo "PLAYWRIGHT_VERSION=$(npm ls @playwright/test --json | jq --raw-output '.dependencies["@playwright/test"].version')"
-          echo $PLAYWRIGHT_VERSION >> $GITHUB_ENV
+          echo "PLAYWRIGHT_VERSION=$(npm ls @playwright/test --json | jq --raw-output '.dependencies["@playwright/test"].version')" >> $GITHUB_ENV
```
Eakam1007, line 143: I looked at the test run, and this works. I am confused why it wasn't working before. Could you please explain what changed here that made it work?

Eakam1007 (approved):

.github/workflows/ci.yaml
shawnyu5: Before, the cache under the name was `Linux-playwright-`, missing the version number. I just messed up injecting the env var into the github environment

Example [here](https://github.com/Seneca-CDOT/starchart/actions/runs/4118747950/jobs/7111645350#step:8:16)

SerpentBytes (approved):

.github/workflows/ci.yaml
Eakam1007: Oh yea, I saw that. I am just confused about how changing this code fixes it

shawnyu5: Ah. It seems like gh needs the full shell syntax of assigning a value to a variable. ie `VAR=hello`. Before, I was doing creating the variable (`VAR=hello`), and putting the value of that variable(`hello`) into gh env. It's just `hello` floating around in there.

idk if this is a great way of explaining it

Eakam1007: Ah, I see. Thanks for the explanation!

Issue #114, opened 2023-02-03
Design certificate explanation UI and write text


humphd:
Our system lets users create SSL Certificates that they can use in various ways.  Many students won't know what this means, or how to do it.  We should add a page with text/screenshots/links to help explain what they are and how to use them in a few common scenarios.

humphd:
This could also include info on how to do your own HTTP-Challenge using certbot and your own server (e.g., if you need to auto-renew).

sirinoks:
@Ririio Why did you remove front end? Does this not require a front end page/component?

Ririio:
> @Ririio Why did you remove front end? Does this not require a front end page/component?

Oh no I did it without thought cause I mistook this from another issue

Eakam1007:
Sorry, is this completed? We still have a page on staging linking to this issue: https://mycustomdomain-dev.senecacollege.ca/certificate/instructions
<img width="959" alt="image" src="https://user-images.githubusercontent.com/67077705/229979911-7636e381-f12f-48d5-93e8-3677685eb713.png">

Ririio:
> @SerpentBytes  Sorry, is this completed? We still have a page on staging linking to this issue: https://mycustomdomain-dev.senecacollege.ca/certificate/instructions
> 
> <img width="959" alt="image" src="https://user-images.githubusercontent.com/67077705/229979911-7636e381-f12f-48d5-93e8-3677685eb713.png">
> 
> 
> 
> 

We are adding this to another issue, forgot which number

Eakam1007:
Could you please link it here for reference?

Ririio:
> Could you please link it here for reference?

#293

PR #156, opened 2023-02-08
Add skeleton pages for user dashboard routes

### LINKED ISSUE #138: Create basic pages and routes for user dashboard pages
As @humphd mentioned in one of our meetings, it would be nice to have the routes setup for different pages so that the UI design could be implemented easily in parallel. The pages served from these routes do not need to have anything more than a simple heading stating what the page is.

This issue focuses on adding the following routes (based on #36):

- /domains - view and manage domains
- /domains/:id - edit subdomain
- /domains/new - create new subdomain
- /certificate - view, renew, or request a certificate (A user can only have one certificate)
- /certificate/instructions - help page about certificates (see #114)

Eakam1007:
Resolves #138 

Added basic pages for the following routes (as mentioned in the issue):
- `/domains`
- `/domains/new`
- `/domains/:id`
- `/certificate`
- `/certificate/instructions`

Also added a component for the header that will display on top of these pages. See image [here](https://github.com/Seneca-CDOT/starchart/pull/156#issuecomment-1421903942)

These are basic pages which can be styled later based on the [figma](https://www.figma.com/file/XAcnMIC1y4qumWinSalHI9/Starchart-UI?node-id=2105%3A2109&t=mPPiqAA5dItrRWGB-1) designs

### Steps to test
- Setup and run starchart locally
- Navigate to the above routes and check that they work. They will only have one simple heading text (irrelevant as this will be changed later anyway)


app/routes/certificates/instructions.tsx
```diff
@@ -0,0 +1,21 @@
+import { Heading, Text, Link } from '@chakra-ui/react';
```
humphd, line 1: What are the rules about using Chakra UI Link vs. Remix Link?  I wonder if we should prefer the router's link component?

humphd (approved):
This is great.  One question.

app/routes/certificates/instructions.tsx
Ririio: You require an `as` tag for when you're using a router library. I'm using Chakra for my other project, and it was suggested I rename `Link` as `ReachLink` for chakra component as implied in their [doc](https://chakra-ui.com/docs/components/link)

Ririio:
Should we have a separate heading page so we don't have to recycle it for each route?

app/routes/certificates/instructions.tsx
Eakam1007: Oh, I simply used a `Link` here to render a `<a>` element which links to #114

Eakam1007:
> Should we have a separate heading page so we don't have to recycle it for each route?

You mean a separate component for the header that appear on the [figma](https://www.figma.com/file/XAcnMIC1y4qumWinSalHI9/Starchart-UI?node-id=2105%3A2109&t=mPPiqAA5dItrRWGB-1) designs?

image

Ririio:
> > Should we have a separate heading page so we don't have to recycle it for each route?
> 
> You mean a separate component for the header that appear on the [figma](https://www.figma.com/file/XAcnMIC1y4qumWinSalHI9/Starchart-UI?node-id=2105%3A2109&t=mPPiqAA5dItrRWGB-1) designs?
> 
> image

yeah... I was wondering how you'll be handling displaying the header across all the routes?

Eakam1007:
> yeah... I was wondering how you'll be handling displaying the header across all the routes?

Hmm, yea I hadn't thought of that. I will need to take a look to understand the best way to do this

Eakam1007:
I have added a common header component for both `domains` and `certificates` routes

Ririio (approved):
Looks good to me

humphd (changes requested):

app/routes/certificates.tsx
```diff
@@ -0,0 +1,14 @@
+import { Outlet } from '@remix-run/react';
+
+import Header from '~/components/header';
+
+export default function CertificatesRoute() {
+  return (
+    <div>
+      <Header />
```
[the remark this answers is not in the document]
Ririio: would rendering it at the top affect the login page?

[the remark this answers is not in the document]
humphd: True, need to put it in the right spot.  We could also decide to always show the header, and remove the user info from the right.  Remix has [`useMatches()`](https://remix.run/docs/en/v1/hooks/use-matches) that lets you get data loaded in any route layer, which you can use to get a user object or something without passing it around via props.  We could use this to know if the user is logged in or not, and change how we show things.

app/routes/__index.tsx
```diff
@@ -0,0 +1,13 @@
+import { Outlet } from '@remix-run/react';
+import Header from '~/components/header';
+
+export default function Index() {
+  return (
+    <div>
```
Ririio, line 6: I recall we're using chakra component as much as possible, should we be using the `Box` component instead of div?

app/routes/certificates.tsx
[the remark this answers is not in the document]
Eakam1007: There is also a way to do this purely based on the routes folder structure. I found [this useful blog](https://dev.to/pckilgore/remix-sharing-your-index-layout-with-sub-routes-2856) about it
I have changed it so the header will be display on all the routes except login

app/routes/__index.tsx
Eakam1007: I can change this to a box, sure

app/routes/__index/certificate/index.tsx
```diff
@@ -0,0 +1,11 @@
+import { Heading } from '@chakra-ui/react';
+
+export default function CertificateIndexRoute() {
+  return (
+    <div>
```
Ririio, line 5: To all files that uses `div` we should replace them to use `Box` if we're following conventions

Ririio (changes requested):

app/routes/__index/certificate/index.tsx
Eakam1007: I don't think there is any point to changing them in the new pages since they will be replaced when they are updated based on the figma designs

Ririio (approved):
Looks good to me

dadolhay (approved):
LGTM

app/routes/certificates.tsx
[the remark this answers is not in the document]
humphd: Neat, I didn't know about this.

humphd (approved):

PR #157, opened 2023-02-08
Improve and secure Dockerfile for production

### LINKED ISSUE #116: Improve the main Dockerfile for production
There is a [Dockerfile](https://github.com/Seneca-CDOT/starchart/blob/main/Dockerfile) already in the tree for the Remix app.  It needs to be improved for production:

- Use node 18.  They use `bullseye-slim`, and it would be good to see if we can make this work with `-alpine` (not a huge problem, but would save some space)
- Pin our image to an image SHA
- Don't use `npm` for the init process, use `node`, use tini docker-compose.yml (i.e., `init: true`)
- Add an `EXPOSE` for the port we use: `8080`
- Don't run as the `root` user
- Prefer `npm ci` vs. `npm install`
- Add a HEALTHCHECK to monitor the container's health
- Update `.dockerignore` to make sure it's not including things we don't want (e.g., `mysql-data/`, `img/`, etc)
- Add https://github.com/hadolint/hadolint-action to our CI to lint our Dockerfiles

There's probably more, but this is a good start.

humphd:
Fixes #116

This updates the Dockerfile from the Blue Stack.  It improves a bunch of things in terms security, size, optimization, and best practices.

You can test it by adding the following `service` to `docker-compose.yml` file:

```yml
  starchart:
    build: .
    # NOTE: the container will crash and restart a bunch of times while MySQL comes up
    restart: always
    ports:
      - 8080:8080
    environment:
      DATABASE_URL: 'mysql://starchart:starchart_password@database:3306/starchart'
      SESSION_SECRET: 'super-duper-secret!!!!!!'
```

You should now be able to go to http://localhost:8080.  If you run `docker ps` you should see the `starchart` container, and its status should eventually say `healthy`.

I've also added a CI step to lint the Dockerfile.

While I was doing this I notice that the `invariant()` calls are having the string value stripped in production builds.  In some cases this won't matter (e.g., where we catch the exception), but during so during startup sucks, since you lose info.  I've changed to manually throw the error + message.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,52 +1,65 @@
-# base node image
-FROM node:16-bullseye-slim as base
+FROM node:18-bullseye-slim@sha256:e2fbe082615911b184e192b05c55e7e38460a2c24c88d92e8c122ea0175fbe56 as base
 
-# set for base and all layer that inherit from it
-ENV NODE_ENV production
+ARG CURL_VERSION=7.74.* \
+  OPENSSL_VERSION=1.1.* \
+  TINI_VERSION=0.19.*
+  
+# hadolint ignore=SC2086
+RUN apt-get update && \
+  apt-get install -y --no-install-recommends \
+  curl=${CURL_VERSION} \
+  openssl=${OPENSSL_VERSION} \
+  tini=${TINI_VERSION} \
+  && rm -rf /var/lib/apt/lists/*
 
-# Install openssl for Prisma
-RUN apt-get update && apt-get install -y openssl
+###############################################################################
 
 # Install all node_modules, including dev dependencies
 FROM base as deps
 
-WORKDIR /myapp
+WORKDIR /app
+COPY package*.json .npmrc ./
+RUN npm ci --include=dev --ignore-scripts
 
-ADD package.json package-lock.json package-lock.json package-lock.json package-lock.json package-lock.json .npmrc ./
-RUN npm install --production=false
+###############################################################################
 
-# Setup production node_modules
+# Remove any non-production dependencies
 FROM base as production-deps
 
-WORKDIR /myapp
+WORKDIR /app
+COPY --from=deps /app/node_modules ./node_modules
+COPY package*.json .npmrc ./
+RUN npm prune --omit=dev
 
-COPY --from=deps /myapp/node_modules /myapp/node_modules
-ADD package.json package-lock.json package-lock.json package-lock.json package-lock.json package-lock.json .npmrc ./
-RUN npm prune --production
+###############################################################################
 
-# Build the app
+# Build the app and generate the prisma client
 FROM base as build
 
-WORKDIR /myapp
+WORKDIR /app
+COPY --from=deps /app/node_modules ./node_modules
+COPY . .
+RUN npx prisma generate \
+  && npm run build
 
-COPY --from=deps /myapp/node_modules /myapp/node_modules
+###############################################################################
 
-ADD prisma .
-RUN npx prisma generate
+# Deploy the built app on top of the production deps, run as non-root
+FROM base as deploy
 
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -1,52 +1,65 @@
-# base node image
-FROM node:16-bullseye-slim as base
+FROM node:18-bullseye-slim@sha256:e2fbe082615911b184e192b05c55e7e38460a2c24c88d92e8c122ea0175fbe56 as base
 
-# set for base and all layer that inherit from it
-ENV NODE_ENV production
+ARG CURL_VERSION=7.74.0-1.3+deb11u5 \
+  OPENSSL_VERSION=1.1.1n-0+deb11u4 \
```
dadolhay, line 4: Is there a reason we are not using openssl 3 instead of 1.1.1?

humphd: I was pinning the version `apt-get` available in this image:

```sh
$ apt list -a openssl
Listing... Done
openssl/stable-security 1.1.1n-0+deb11u4 arm64
openssl/stable 1.1.1n-0+deb11u3 arm64
```

dadolhay: From this output it seems to me that they basically only allow the latest version to be installed (the last two builds of the same version is listed above).

I would suggest only pinning the version 1.1.1 (if 3.0.0 is not available), because the letter is the security patch version, and if a new, critical bug comes out in openssl, it's probably best to let the apt system install the latest version that's still compatible (Also, probably they'd make it so, that 1.1.1n is unavailable, and they release a new, 1.1.1o, making this install step fail)

humphd: Sure, so I could do this:

```
OPENSSL_VERSION=1.1.\*
CURL_VERSION=7.74.\*
TINI_VERSION=0.19.\*
```

Unless you have a better suggestion?

dadolhay: If this works, I think this is a good solution 👍

humphd:
Updated the versions to allow patch/security fixes.  Good suggestion @dadolhay.

dadolhay (approved):

SerpentBytes:
@humphd, Does it matter if the SESSION_SECRET in docker-compose is not consistent with the variable with the same name in .env.example file?

humphd:
@SerpentBytes for testing, it should not.  Only one value will get used.  This code above is only to make it easy to test, and you can change it to whatever you want.

Dockerfile
```diff
@@ -1,52 +1,65 @@
-# base node image
-FROM node:16-bullseye-slim as base
-
-# set for base and all layer that inherit from it
-ENV NODE_ENV production
-
-# Install openssl for Prisma
-RUN apt-get update && apt-get install -y openssl
+FROM node:18-bullseye-slim@sha256:e2fbe082615911b184e192b05c55e7e38460a2c24c88d92e8c122ea0175fbe56 as base
```
shawnyu5, line 1: Is there a reason we pin to a commit?

sfrunza13: If I remember from CCP David likes to do this because he knows exactly which version of the image will work for the app and doesn't want it to be left to the latest for example because that is subject to potentially breaking changes.

sfrunza13 (approved):

SerpentBytes (approved):

Dockerfile
humphd: Best practices for deploying Docker images includes pinning your dependencies so you have reproducible environments.  If you deploy to staging with version 0.0.1 and then by the time you deploy to production it's 0.0.2, you are not deploying the same software.

It *should* be true that people only do breaking changes with a major update; but "should" isn't what happens.  So you end-up having your production boxes breaking because one version of something is slightly off.

The answer is to pin your images, dependencies, modules, etc.

Issue #143, opened 2023-02-06
Update seed.ts, for Record, Certificate, and Challenge tables


cychu42:
We should add the ability to seed Record, Certificate, and Challenge tables, for future testing.

humphd:
It would be great to make sure we 5 or so Records made of various types (A, CNAME, etc) with good examples so that @Myrfion, @Ririio, and @Eakam1007 working on front-end have things to use in their implementation.

Ririio:
I have a question regarding the "ports" property. With the user being able to add as many ports as they want, should the value in the database be an array as to a string?

Issue #153, opened 2023-02-07
Fix/change the user schema's column names for records and certificate


cychu42:
In User, the column for records is currently named `subdomians`, and we have decided to use the term record over subdomain. It should be changed to `record` for consistency, and it also has typo. See #79 
Secondly, `Certificates` shouldn't be plural since there's only one certificate we are tracking for each user.

I'm thinking we can stick to singular names to keep things easier, especially if we ever decide to change relationship types later.

cychu42:
The `privkey` column in `certificate` model can be changed to `privateKey` to be clearer.

PR #160, opened 2023-02-08
Add seeding for Record, Certificate, and Challenge

cychu42:
Closes: #143 
Let me know if you think more example data should be included, as well as whether the examples need changes.

## Changes
- Code for seeding Record, Certificate, and Challenge tables in `prisma/seed.ts`.
- Code for cleaning all existing data in the database in `prisma/seed.ts`.

## Testing
1. Set up the project normally to see if seeding works
or
2. run `npx prisma db seed` in a working copy of the project to test it

### DIFF
--- a/prisma/seed.ts
+++ b/prisma/seed.ts
@@ -8,11 +8,16 @@ async function seed() {
   const firstName = 'Starchart';
   const lastName = 'Developer';
   const email = 'dev@starchart.invalid';
+  let recordExpDate = new Date();
+  recordExpDate.setMonth(recordExpDate.getMonth() + 6); // a record expires after 6 months
+  let certExpDate = new Date();
+  certExpDate.setDate(certExpDate.getDate() + 90); // certificate expires after 90 days
 
-  // cleanup the existing database
-  await prisma.user.delete({ where: { username } }).catch(() => {
-    // no worries if it doesn't exist yet
-  });
+  // cleanup the existing database; no worries if it doesn't exist yet
+  await prisma.record.deleteMany().catch(() => {});
+  await prisma.challenge.deleteMany().catch(() => {});
+  await prisma.certificate.deleteMany().catch(() => {});
+  await prisma.user.deleteMany().catch(() => {});
 
   await prisma.user.create({
     data: {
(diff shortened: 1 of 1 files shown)


prisma/seed.ts
```diff
@@ -23,6 +25,118 @@ async function seed() {
     },
   });
 
+  await prisma.record.createMany({
+    data: [
+      // A record
+      {
+        id: 1,
```
humphd, line 32: Based on https://github.com/Seneca-CDOT/starchart/blob/main/prisma/schema.prisma#L23, you shouldn't need to provide an `id`, it should automatically assign a new one.

prisma/seed.ts
```diff
@@ -23,6 +25,118 @@ async function seed() {
...
+        username: username,
+        name: 'OSD700',
+        type: `A`,
+        value: '172.5.251.0',
+        description: 'For final porject.',
```
humphd, line 37: Typo in `project`

prisma/seed.ts
```diff
@@ -23,6 +25,118 @@ async function seed() {
     },
   });
 
+  await prisma.record.createMany({
+    data: [
+      // A record
+      {
+        id: 1,
+        username: username,
+        name: 'OSD700',
```
humphd, line 34: Prefer lowercase for all `name` values`

prisma/seed.ts
```diff
@@ -23,6 +25,118 @@ async function seed() {
     },
   });
 
+  await prisma.record.createMany({
+    data: [
+      // A record
+      {
+        id: 1,
+        username: username,
```
humphd, line 33: When the key/value are the same, list it once: `username,`

prisma/seed.ts
```diff
@@ -23,6 +25,118 @@ async function seed() {
...
+        createdAt: new Date('2022-03-15T01:01:01'),
+        expiresAt: new Date('2022-09-15T12:59:59'),
+        status: `pending`,
+      },
+    ],
+  });
+
+  await prisma.certificate.create({
+    data: {
+      id: 1,
```
humphd, line 110: Same comments here.  Let Prisma do the automatic fields for you.

humphd (changes requested):

prisma/seed.ts
cychu42: Ah, thanks.

prisma/seed.ts
cychu42: I see. Thank you.

prisma/seed.ts
humphd: Based on the call today, sounds like you want this for foreign key.  Resolving.

prisma/seed.ts
```diff
@@ -23,6 +25,118 @@ async function seed() {
...
+      username: username,
+      subject: `CN=*.${username}.example.com`,
+      certificate:
+        '-----BEGIN CERTIFICATE-----ApfFCv0O65TYkp5jEWSlU8PhKYD43nXA=-----END CERTIFICATE-----',
+      orderUrl: `orderURL.example.com`,
+      privkey:
```
SerpentBytes, line 116: Maybe follow camelCase convention here to be consistent with the rest of the code.

cychu42: OK, done.

cychu42:
Thank you for the feedbacks. I did a number of changes:
- I would like to keep the specified `id`s to avoid situations where there’s is an issue referencing them for foreign keys. Otherwise. auto incrementing will just keep incrementing the number every time someone runs the seeding command.
- For dates, I changed dates of creation to be now. Expiration dates are a certain time from now. 6 months for records, and 90 days for certificates.
- The challenge part is removed as Denes suggested.
- `orderUrl` is also changed as suggested.

SerpentBytes:
@cychu42 is privkey supposed to be privKey?

humphd (changes requested):
Looks good, other than the `privKey` vs. `privkey` change @SerpentBytes wants.

cychu42:
> @cychu42 is privkey supposed to be privKey?

It's supposed to be `privkey`, as it's what's in the schema. However, I see that it should probably be `privKey` or even `privateKey`. I originally took it from the [Data Architecture](https://github.com/Seneca-CDOT/starchart/wiki/Data-Architecture) page when making the schema.
I think the later (`privateKey`) makes more sense as it will be clearer to new readers.

If we are going to change it, I can add it as a suggestion under another PR to change the schema, like #153. Just to keep scope of issues separate.

humphd (approved):

dadolhay (approved):

SerpentBytes (approved):

PR #161, opened 2023-02-08
Add BullMQ + Redis and prototype notifications queue

humphd:
This is an initial attempt to add [BullMQ](https://docs.bullmq.io/) and [Redis](https://redis.io/) (i.e., [ioredis](https://github.com/luin/ioredis)) to our code.

I've done the most basic thing possible vs. making this so it will scale.  At the moment we have:

- `app/queues/*` which holds all of our queues.
- `app/queues/notifications.server.ts` which is our first queue, dealing with sending user notifications asynchronously.  We'll add more as we go, but this can provide a template.
- The `Queue` and `Worker` are both co-located in the same file, and in the same process.  Down the road, we can break the worker into its own process or container(s).  But I didn't want to do this at first, since it will complicate things significantly.

The way this queue would be used is like this.  Imagine we want to send a message to a user.  The code would be:

```ts
import notificationsQueue from '~/queues/notifications.server';

...

// Send a new user a welcome message. Create a unique job id for this job
const jobId = `${user.email}-welcome`;
// Add the job to the queue, passing the data it will need. We have to wait for
// the job to get put into Redis, but not for the email to be sent.
await notificationsQueue.add(jobId, {
  emailAddress: user.email,
  subject: "Welcome to my.custom.domain!",
  message: "We're thrilled to welcome you to my.custom.domain...",
});
```

The job can now be handled by the `Worker`, which will attempt to send it at some point in the near future.  If it fails, it will retry the job 3 more times, waiting progressively longer each time.  We can configure all of this.

I don't have anywhere in the code to hook into this yet, so it's not easy to test.


app/lib/redis.server.ts
```diff
@@ -0,0 +1,28 @@
+import Redis from 'ioredis';
+
+import type { Redis as RedisType, RedisOptions } from 'ioredis';
+
+let redis: RedisType;
+
+declare global {
+  var __redis: RedisType | undefined;
```
dadolhay, line 8: Why do we do this global, if we also later export redis in line 5?

app/lib/redis.server.ts
```diff
@@ -0,0 +1,28 @@
...
+const redisOptions: RedisOptions = {
+  maxRetriesPerRequest: null,
+  enableReadyCheck: false,
+};
+
+// this is needed because in development we don't want to restart
+// the server with every change, but we want to make sure we don't
+// create a new connection to the Redis with every change either.
+if (process.env.NODE_ENV === 'production') {
+  redis = new Redis(process.env.REDIS_URL || '', redisOptions);
```
dadolhay, line 20: Will we have a dedicated redis that survives between releases?

docker-compose.yml
```diff
@@ -9,14 +9,25 @@ services:
...
+  redis:
+    image: redis:7.0.8-alpine3.17
+    container_name: redis
+    command: ['redis-server']
+    ports:
+      # Only for dev
+      - 6379:6379
+    volumes:
+      # Persist data to ./redis-data. Remove if you want to clear the db.
+      - ./redis-data:/data
```
dadolhay, line 29: So we are capable of persisting data on disk also with redis? (I always thought that redis is in-memory only)

app/queues/notifications.server.ts
```diff
@@ -0,0 +1,55 @@
...
+});
+
+notificationsQueue.on('error', (err) => {
+  logger.warn('notifications queue error', err);
+});
+
+/**
+ * A worker for processing a notification job, sending an email to the user
+ */
+export const notificationsWorker = new Worker<NotificationsQueueData>(
```
dadolhay, line 36: How do you register this worker into BullMQ? Where is BullMQ initialized?

app/lib/redis.server.ts
humphd: NOTE: this code is coming from https://github.com/remix-run/examples/tree/main/bullmq-task-queue.

I think they do it this way so that they only put it on the global if not in production.  So it's module-level unless it needs to get elevated to the global.

app/lib/redis.server.ts
humphd: We can run our own instance on the cluster, yes.  So when we re-deploy, it will remain.

docker-compose.yml
humphd: Yes.  You are correct that Redis is in-memory.  However, you have two options for persisting data:

1. append-only file
2. regular snapshots

app/queues/notifications.server.ts
humphd: Unless you tell it not to, a `Worker` starts processing as soon as it gets created.  So in this case, the notifications worker will begin waiting for jobs to appear.

app/lib/redis.server.ts
dadolhay: Perfect!

docker-compose.yml
dadolhay: Great!

app/lib/redis.server.ts
dadolhay: Okay, so thos only affects hot reload, Thanks!

app/queues/notifications.server.ts
dadolhay: Okay, so I guess, importing it in the first place causes the global initialization, and the new Worker constructor does also registers it into the global system. Interestint implementation. Thanks for all the explanations!

dadolhay (dismissed):
PR is still in draft, but it looks great!

humphd:
I added graceful shutdown for the workers and redis.

This won't build yet, I need to get the nodemailer stuff landed first, and then I'll iterate on this some more.

Thanks for the first review @dadolhay, appreciated!

humphd:
I've updated this to use the work @TD just landed for notifications.  I also added a `/dev` page we can use to create test UI and added a button to send a notification.

Once you click the button, it adds a job to the notifications queue.  This will get picked up by the worker and an email sent.  You can check it in MailHog via http://localhost:8025:

Screenshot 2023-02-09 at 1 45 18 PM

To make this work, I also had to fix our [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) policy in [helmetjs](https://helmetjs.github.io/).  It breaks hot-reloads in the UI due to unsafe inline script injection.  We don't want that in production, but in development it's nice so the app responds to code changes in real-time.

What I'm doing here on the `/dev` page could be copied by other people, who need to try out UI-triggered actions (e.g., @Genne23v with DNS, @dadolhay with Certs).

humphd:
Added a fix for the node EventEmitter max listeners warning you get when using Redis/BullMQ:

```
(node:35293) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 17 error listeners added to [Commander]. Use emitter.setMaxListeners() to increase limit
```

By default, node.js warns if an event emitter has more than 10 listeners; however, the way Redis works, you need way more than that all the time.  The recommended fix for this by the BullMQ maintainer is to bump the default value up.  In Telescope, we've used 32 as our default and not had any issues.

SerpentBytes (approved):
Seems to work on my end.

dadolhay (approved):

SerpentBytes (approved):

PR #162, opened 2023-02-08
Add interface to DNS and DB

### LINKED ISSUE #151: Add wrapper to connect DB and DNS module
`starchart` needs a wrapper for client `app/lib/client.server.ts` to manipulate DB and DNS module together. It needs following methods

- `createUserDomain(hostedZoneId, username, name, type, value)` => Create a record in Route53 and DB
- `updateUserDomain(hostedZoneId, username, name, type, value)` => Update a record in Route53 and DB
- `deleteUserDomain(hostedZoneId, username, name, type, value)` => Delete a record in Route53 and DB
- `getUserDomains(username)` => Get all user's domains 
- `getUserDomainInfo(username, name, type, value)` => Get all metadata of matching record from DB
- `updateUserDomainInfo(username, id, ...props)` => Update user's domain meta info
- `isExpired(id)` => Check if a record is expired
- `isBeingExpired(id)` => Check if notification needs to be sent to user for the record

Feel free to add anything if I miss something here. Thanks

Genne23v:
Closes #151 
I wrote a long detailed change summary here, but I hit a wrong button and lost it all. This is not a client interface as I thought this is too big to be a direct interface to client. Please let me know if I took a wrong approach. 

- `createUserDomain()` => Create a record in Route 53 and DB and return created record. If `type`, `name`, `value` exists, it declines to add it to DB. 
- `updateUserDomain()` 
- `deleteUserDomain()`
- `removeIfExpired()` => Remove a record when it's expired
- `isDomainBeingExpired()`
- `isDomainExpired()`

Quick high level review would be much appreciated!

### DIFF
--- a/app/lib/dns.server.ts
+++ b/app/lib/dns.server.ts
@@ -158,7 +158,7 @@ export const getChangeStatus = async (changeId: string) => {
 };
 
 const isNameValid = (name: string) => {
-  return /^[a-z0-9-]+.[a-z]+.[a-z]+\.?$/.test(name);
+  return /^[a-z0-9-]+.[a-z0-9-]+.[a-z]+.[a-z]+\.?$/.test(name);
 };
 
 const isValueValid = (type: RecordType, value: string) => {
--- a/app/lib/dns.server.ts
+++ b/app/lib/dns.server.ts
@@ -1,15 +1,24 @@
+import logger from './logger.server';
 import {
   Route53Client,
   CreateHostedZoneCommand,
   ChangeResourceRecordSetsCommand,
   GetChangeCommand,
 } from '@aws-sdk/client-route-53';
+import { isIPv4, isIPv6 } from 'is-ip';
+
 import type {
   CreateHostedZoneResponse,
   ChangeResourceRecordSetsResponse,
   GetChangeResponse,
 } from '@aws-sdk/client-route-53';
-import { isIPv4, isIPv6 } from 'is-ip';
+import type { RecordType } from '@prisma/client';
+
+if (process.env.NODE_ENV === 'production') {
+  if (!process.env.AWS_ROUTE53_HOSTED_ZONE_ID) {
+    throw new Error('AWS_ROUTE53_HOSTED_ZONE_ID is missing');
+  }
+}
 
 export const route53Client = new Route53Client({
   endpoint: process.env.AWS_ENDPOINT_URL || 'http://localhost:5053',
@@ -20,13 +29,6 @@ export const route53Client = new Route53Client({
     sessionToken: process.env.AWS_SESSION_TOKEN,
   },
 });
-
-export enum RecordType {
-  A = 'A',
-  AAAA = 'AAAA',
-  CNAME = 'CNAME',
-  TXT = 'TXT',
-}
 
 export const createHostedZone = async (domain: string) => {
   try {
@@ -41,33 +43,21 @@ export const createHostedZone = async (domain: string) => {
     }
     return response.HostedZone.Id.replace('/hostedzone/', '');
   } catch (error) {
-    // TODO: fix with better logging...
-    console.warn(error);
+    logger.warn(error);
     throw new Error(`Error while creating hosted zone`);
   }
 };
 
-export const createRecord = (
-  hostedZoneId: string,
-  type: RecordType,
-  name: string,
-  value: string
-) => {
+export const createRecord = (type: RecordType, name: string, value: string) => {
   try {
-    return upsertRecord(hostedZoneId, type, name, value);
+    return upsertRecord(type, name, value);
   } catch (error) {
-    // TODO: fix with better logging...
-    console.warn(error);
+    logger.warn(error);
     throw new Error(`Error occurred while creating resource record`);
   }
 };
 
-export const upsertRecord = async (
-  hostedZoneId: string,
-  type: RecordType,
-  name: string,
-  value: string
-) => {
+export const upsertRecord = async (type: RecordType, name: string, value: string) => {
   try {
     if (!isNameValid(name)) {
       throw new Error('Invalid name provided');
@@ -94,7 +84,7 @@ export const upsertRecord = async (
           },
         ],
       },
-      HostedZoneId: hostedZoneId,
+      HostedZoneId: process.env.AWS_ROUTE53_HOSTED_ZONE_ID,
     });
     const response: ChangeResourceRecordSetsResponse = await route53Client.send(command);
 
@@ -103,18 +93,12 @@ export const upsertRecord = async (
     }
     return response.ChangeInfo.Id;
   } catch (error) {
-    // TODO: fix with better logging...
-    console.warn(error);
+    logger.warn(error);
     throw new Error(`Error occurred while updating resource record: ${error}`);
   }
 };
 
-export const deleteRecord = async (
-  hostedZoneId: string,
-  type: RecordType,
-  name: string,
-  value: string
-) => {
+export const deleteRecord = async (type: RecordType, name: string, value: string) => {
   try {
     if (!isNameValid(name)) {
       throw new Error('Invalid name provided');
@@ -141,7 +125,7 @@ export const deleteRecord = async (
           },
         ],
       },
-      HostedZoneId: hostedZoneId,
+      HostedZoneId: process.env.AWS_ROUTE53_HOSTED_ZONE_ID,
     });
 
     const response: ChangeResourceRecordSetsResponse = await route53Client.send(command);
@@ -151,8 +135,7 @@ export const deleteRecord = async (
     }
     return response.ChangeInfo.Id;
   } catch (error) {
-    // TODO: fix with better logging...
-    console.warn(error);
+    logger.warn(error);
     throw new Error(`Error occurred while deleting resource record`);
   }
 };
@@ -169,8 +152,7 @@ export const getChangeStatus = async (changeId: string) => {
     }
     return response.ChangeInfo.Status;
   } catch (error) {
-    // TODO: fix with better logging...
-    console.warn(error);
+    logger.warn(error);
     throw new Error(`Error occurred while getting change status`);
   }
 };
--- a/app/lib/domains.server.ts
+++ b/app/lib/domains.server.ts
@@ -1,0 +1,138 @@
+import * as dns from './dns.server';
+import { RecordStatus } from '@prisma/client';
+import { prisma } from '~/db.server';
+
+import type { RecordType, Record } from '@prisma/client';
+
+interface DnsRecord {
+  type: RecordType;
+  name: string;
+  value: string;
+}
+
+interface DomainRecord {
+  id?: Record['id'];
+  username?: Record['username'];
+  description?: Record['description'];
+  course?: Record['course'];
+  ports?: Record['ports'];
+}
+
+type Domain = DnsRecord & DomainRecord;
+
+export const createUserDomain = async (data: Domain) => {
+  if (!data.username) {
+    throw new Error('Username is not provided');
+  }
+
+  const existingRecords = await prisma.record.count({
+    where: {
+      name: data.name,
+      type: data.type,
+      value: data.value,
+    },
+  });
+
+  if (existingRecords > 0) {
+    throw new Error('Record already exists');
+  }
+
+  const [, result] = await Promise.all([
+    dns.createRecord(data.type, data.name, data.value),
+    prisma.record.create({
+      data: {
+        username: data.username,
+        type: data.type,
+        name: data.name,
+        value: data.value,
+        expiresAt: setMonthsFromNow(6),
+        status: RecordStatus.pending,
+      },
+    }),
+  ]);
+
+  if (!result) {
+    throw new Error('Could not create a record in DB');
+  }
+  return result;
+};
+
+export const updateUserDomain = async (data: Domain) => {
+  const [, result] = await Promise.all([
+    dns.upsertRecord(data.type, data.name, data.value),
+    prisma.record.update({
+      where: {
+        id: data.id,
+      },
+      data: {
+        name: data.name,
+        type: data.type,
+        value: data.value,
+        description: data.description,
+        course: data.course,
+        ports: data.ports,
+        expiresAt: setMonthsFromNow(6),
+        status: RecordStatus.pending,
+      },
+    }),
+  ]);
+
+  if (!result) {
+    throw new Error('Could not update the record in DB');
+  }
+  return result;
+};
+
+export const deleteUserDomain = async (data: Domain) => {
+  const [, result] = await Promise.all([
+    dns.deleteRecord(data.type, data.name, data.value),
+    prisma.record.delete({
+      where: {
+        id: data.id,
+      },
+    }),
+  ]);
+
+  if (!result) {
+    throw new Error('Could not delete the record in DB');
+  }
+  return result;
+};
+
+export const removeIfExpired = async (data: Domain) => {
+  const recordFromId = await prisma.record.findUnique({
+    where: {
+      id: data.id,
+    },
+  });
+
+  if (!recordFromId) {
+    throw new Error('Record with provided ID does not exist');
+  }
+
+  if (isExpired(recordFromId.expiresAt)) {
+    const [, result] = await Promise.all([
+      dns.deleteRecord(data.type, data.name, data.value),
+      prisma.record.delete({
+        where: {
+          id: data.id,
+        },
+      }),
+    ]);
+
+    if (!result) {
+      throw new Error('Could not delete the record in DB');
+    }
+    return result;
+  }
+};
+
+export const isExpired = (expiresAt: Date) => expiresAt.getTime() < new Date().getTime();
+
+export const willExpireIn = (expiresAt: Date, months: number = 1) =>
+  new Date(expiresAt.setMonth(expiresAt.getMonth() - months)).getTime() < new Date().getTime();
+
+export const setMonthsFromNow = (months: number) =>
+  new Date(new Date().setMonth(new Date().getMonth() + months));
+
+export type { Domain };
--- a/test/unit/dns.server.test.ts
+++ b/test/unit/dns.server.test.ts
@@ -4,17 +4,17 @@ import {
   upsertRecord,
   deleteRecord,
   getChangeStatus,
-  RecordType,
 } from '~/lib/dns.server';
+import { RecordType } from '@prisma/client';
 
 describe('DNS server lib function test', () => {
   let hostedZoneId: string;
+  const rootDomain = 'starchart.com';
 
   beforeAll(async () => {
-    hostedZoneId = await createHostedZone('starchart.com');
+    hostedZoneId = await createHostedZone(rootDomain);
+    process.env.AWS_ROUTE53_HOSTED_ZONE_ID = hostedZoneId;
   });
-
-  afterAll(() => resetHostedZone());
 
   test('Hosted zone is created and hosted zone is returned', () => {
     expect(hostedZoneId.length).toBeGreaterThan(0);
@@ -22,107 +22,64 @@ describe('DNS server lib function test', () => {
   });
 
   test('createRecord() creates a record in existing hosted zone and returns changeId', async () => {
-    const changeId = await createRecord(
-      hostedZoneId,
-      RecordType.A,
-      'my-domain.starchart.com',
-      '192.168.0.1'
-    );
+    const changeId = await createRecord(RecordType.A, `my-domain.${rootDomain}`, '192.168.0.1');
     expect(changeId?.length).toBeGreaterThan(1);
   });
 
-  test('createRecord() throws an error when invalid hosted zone ID or name or value ', async () => {
+  test('createRecord() throws an error when name or value is provided ', async () => {
     await expect(
-      createRecord('WRONGHOSTEDZONEID', RecordType.A, 'sub2.starchart.com', '192.168.0.2')
+      createRecord(RecordType.A, `invalid_domain.${rootDomain}`, '192.168.0.1')
     ).rejects.toThrow();
+    await expect(createRecord(RecordType.A, `sub2.${rootDomain}`, '192-168-0-1')).rejects.toThrow();
     await expect(
-      createRecord(hostedZoneId, RecordType.A, 'invalid_domain.starchart.com', '192.168.0.1')
-    ).rejects.toThrow();
-    await expect(
-      createRecord(hostedZoneId, RecordType.A, 'sub2.starchart.com', '192-168-0-1')
-    ).rejects.toThrow();
-    await expect(
-      createRecord(hostedZoneId, RecordType.AAAA, 'sub2.starchart.com', '192.168.0.1')
+      createRecord(RecordType.AAAA, `sub2.${rootDomain}`, '192.168.0.1')
     ).rejects.toThrow();
   });
 
   test('upsertRecord() updates an existing record in hosted zone and returns changeId ', async () => {
-    await createRecord(hostedZoneId, RecordType.A, 'sub.starchart.com', '192.168.0.2');
+    await createRecord(RecordType.A, `sub.${rootDomain}`, '192.168.0.2');
     const changeId = await upsertRecord(
-      hostedZoneId,
       RecordType.AAAA,
-      'sub.starchart.com',
+      `sub.${rootDomain}`,
       '2001:db8:3333:4444:5555:6666:7777:8888'
     );
     expect(changeId?.length).toBeGreaterThan(1);
   });
 
-  test('upsertRecord() throws an error when non existing hosted zone ID or invalid name or value was provided', async () => {
+  test('upsertRecord() throws an error when invalid name or value is provided', async () => {
     await expect(
-      upsertRecord('WRONGHOSTEDZONEID', RecordType.A, 'sub.starchart.com', '192.168.0.3')
+      upsertRecord(RecordType.A, `invalid.domain.${rootDomain}`, '192.168.0.2')
     ).rejects.toThrow();
-    await expect(
-      upsertRecord(hostedZoneId, RecordType.A, 'invalid.domain.starchart.com', '192.168.0.2')
-    ).rejects.toThrow();
-    await expect(
-      upsertRecord(hostedZoneId, RecordType.A, 'sub.starchart.com', '192-168-0-2')
-    ).rejects.toThrow();
-    await expect(
-      upsertRecord(hostedZoneId, RecordType.CNAME, 'sub.starchart.com', '')
-    ).rejects.toThrow();
+    await expect(upsertRecord(RecordType.A, `sub.${rootDomain}`, '192-168-0-2')).rejects.toThrow();
+    await expect(upsertRecord(RecordType.CNAME, `sub.${rootDomain}`, '')).rejects.toThrow();
   });
 
   test('deleteRecord() deletes an existing record in hosted zone and returns changeId', async () => {
-    await createRecord(hostedZoneId, RecordType.A, 'to-be-deleted.starchart.com', '192.168.0.0');
-    const changeId = await deleteRecord(
-      hostedZoneId,
-      RecordType.A,
-      'to-be-deleted.starchart.com',
-      '192.168.0.0'
-    );
+    await createRecord(RecordType.A, 'to-be-deleted.starchart.com', '192.168.0.0');
+    const changeId = await deleteRecord(RecordType.A, `to-be-deleted.${rootDomain}`, '192.168.0.0');
 
     expect(changeId.length).toBeGreaterThan(1);
   });
 
   test('deleteRecord() throws an error when attempting to delete non existing record', async () => {
     await expect(
-      deleteRecord(hostedZoneId, RecordType.A, 'not-exist.starchart.com', '192.168.0.1')
+      deleteRecord(RecordType.A, `not-exist.${rootDomain}`, '192.168.0.1')
     ).rejects.toThrow();
   });
 
   test('deleteRecord() throws an error when invalid hosted zone ID or invalid value is provided', async () => {
     await expect(
-      deleteRecord(
-        'WRONGHOSTEDZONEID',
-        RecordType.A,
-        'wrong-hosted-zone.starchart.com',
-        '192.168.0.2'
-      )
+      deleteRecord(RecordType.A, `wrong-hosted-zone.${rootDomain}`, '192.168.0.2')
     ).rejects.toThrow();
     await expect(
-      deleteRecord(hostedZoneId, RecordType.AAAA, 'wrong-hosted-zone.starchart.com', '192.168.0.2')
+      deleteRecord(RecordType.AAAA, `wrong-hosted-zone.${rootDomain}`, '192.168.0.2')
     ).rejects.toThrow();
   });
 
   test('Get the status of record update using changeId', async () => {
-    const changeId = await upsertRecord(
-      hostedZoneId,
-      RecordType.A,
-      'sub1.starchart.com',
-      '192.168.0.2'
-    );
+    const changeId = await upsertRecord(RecordType.A, `sub1.${rootDomain}`, '192.168.0.2');
     const status = await getChangeStatus(changeId!);
 
     expect(status).toEqual('INSYNC');
   });
 });
-
-const resetHostedZone = async () => {
-  try {
-    fetch('http://localhost:5053/moto-api/reset', {
-      method: 'POST',
-    });
-  } catch (error) {
-    throw new Error(`Error occurred while deleting records in hosted zone: ${error}`);
-  }
-};
--- a/test/unit/dns.server.test.ts
+++ b/test/unit/dns.server.test.ts
@@ -10,6 +10,7 @@ import { RecordType } from '@prisma/client';
 describe('DNS server lib function test', () => {
   let hostedZoneId: string;
   const rootDomain = 'starchart.com';
+  const studentId = 'jdo12';
 
   beforeAll(async () => {
     hostedZoneId = await createHostedZone(rootDomain);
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -50,6 +50,7 @@
     "helmet": "^6.0.1",
     "is-ip": "^5.0.0",
     "isbot": "^3.6.5",
+    "random-words": "^1.2.1",
```
humphd, line 53: I don't think we need a dependency for this.  Can you use something like:

```ts
const randomWord = (function () {
  let seed = Date.now();
  return () => String(seed++);
}());
```

app/models/record.server.ts
```diff
@@ -0,0 +1,98 @@
+import type { Record, User } from '@prisma/client';
+import { prisma } from '../db.server';
+
+export const createRecordInDB = async (
+  username: User['username'],
+  type: Record['type'],
+  name: Record['name'],
+  value: Record['value'],
+  createdAt: Record['createdAt'],
```
humphd, line 9: `createdAt` and `updatedAt` should be handled internally by Prisma.  You shouldn't pass them in or set them manually.  Same applies throughout this PR (please look for other places with the same problem and fix).

app/lib/client.server.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  const result = await createRecordInDB(
+    username,
+    type,
+    name,
+    value,
+    new Date(),
+    new Date(),
+    new Date(new Date().setMonth(new Date().getMonth() + 6)),
```
humphd, line 57: Make a function to better document this:

```ts
const dateFromNow = (months: number) => new Date(new Date().setMonth(new Date().getMonth() + months));
```

app/lib/client.server.ts
```diff
@@ -0,0 +1,179 @@
...
+  createRecordInDB,
+  updateRecordInDB,
+  deleteRecordInDB,
+  findRecordById,
+  doesRecordExistInDB,
+} from '../models/record.server';
+import { RecordStatus } from '@prisma/client';
+
+export const createUserDomain = async (
+  hostedZoneId: string,
```
humphd, line 13: Throughout this PR, you're using multiple args per function.  As this number increases, positional args become more difficult to use.  Consider a function like this:

```ts
doSomething(true, true, 1, 5, false, true);
```

A better approach is to pass a single `Object` and name all the params.

Another suggestion I have: provide a default value for `hostedZoneId` using whatever is in the environment.  In testing, you can override it, but in normal code, we can omit passing it on the Object of args, and it will do the right thing.  Do this throughout this PR.

app/lib/client.server.ts
```diff
@@ -0,0 +1,179 @@
...
+  name: string,
+  value: string
+) => {
+  const changeId = await createRecord(hostedZoneId, type, name, value);
+
+  if (!changeId) {
+    throw new Error('Could not get ChangeID from creating a record in Route 53');
+  }
+
+  const response = await getChangeStatus(changeId);
```
humphd, line 25: I wonder if this is useful or not?  The change will almost certainly take more than 1 call to `getChangeStatus()`, so it might be better to set it to `PENDING` by default, and to this status update in a background task?

app/lib/client.server.ts
```diff
@@ -0,0 +1,179 @@
...
+      break;
+    case 'INSYNC':
+      status = RecordStatus.active;
+      break;
+    default:
+      status = RecordStatus.error;
+      break;
+  }
+
+  const existingRecords = await doesRecordExistInDB(type, name, value);
```
humphd, line 44: This seems like it should be done first.  If you're going to error on the record already existing, why not fail early?

app/models/record.server.ts
```diff
@@ -0,0 +1,98 @@
+import type { Record, User } from '@prisma/client';
```
humphd, line 1: I wonder if this file is achieving anything.  Maybe we'd be better to use Prisma directly in your other file, since all you're doing here is calling the Prisma functions with the passed arguments (i.e., you're not adding a lot of value that we couldn't do directly in other code).

app/lib/client.server.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  const result = await updateRecordInDB(
+    id,
+    type,
+    name,
+    value,
+    description,
+    course,
+    ports,
+    new Date(),
```
humphd, line 110: You can see this TS error is showing the problem with positional args:

```
Expected 7 arguments, but got 10.
```

app/models/record.server.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export const deleteRecordInDB = async (id: Record['id']) => {
+  return await prisma.record.delete({
+    where: {
+      id,
+    },
+  });
+};
+
+export const getRecordFromDB = async (id: Record['id']) => {
```
humphd, line 84: `getRecordById()`

app/models/record.server.ts
```diff
@@ -0,0 +1,98 @@
...
+  return await prisma.record.findUnique({
+    where: {
+      id,
+    },
+  });
+};
+
+export const getUserRecordsFromDB = async (username: User['username']) => {
```
humphd, line 92: `getRecordsByUsername()`

app/models/record.server.ts
```diff
@@ -0,0 +1,98 @@
...
+      value,
+      createdAt,
+      updatedAt,
+      expiresAt,
+      status,
+    },
+  });
+};
+
+export const findRecordById = async (id: Record['id']) => {
```
humphd, line 28: Isn't this the same as what you have below for `getRecordFromDB()`?

humphd (changes requested):
Nice work @Genne23v.  It's great how you're working through hard problems, creating APIs, and writing tests.

I've done a first pass through this code.  I'll do more after you've had a chance to fix these.

Genne23v:
> I thought this might be a lot of rework. It's good to hear that I'm on the right path. I will thoroughly review the whole code again to reduce the amount of review.

humphd:
> I thought this might be a lot of rework. It's good to hear that I'm on the right path. I will thoroughly review the whole code again to reduce the amount of review.

I think it might be useful to hash our APIs in Issues before you dive in, since it means more work for you to change things around.  However, I'm also pleased to see you pushing so hard on this DNS code.

I do want to change a bunch of things here, but we can do it starting from what you have.  I'd avoid changing the tests too much at this point, until we figure out the right API.

app/lib/client.server.ts
```diff
@@ -0,0 +1,179 @@
+import { createRecord, upsertRecord, deleteRecord, getChangeStatus } from './dns.server';
```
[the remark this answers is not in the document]
Genne23v: I would like to suggest `domain-record-wrapper.server` or `domain-wrapper.server` if previous one is too long.

app/lib/client.server.ts
```diff
@@ -0,0 +1,179 @@
...
+import type { RecordType } from './dns.server';
```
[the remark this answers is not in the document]
Genne23v: Will be rearranged

app/models/record.server.ts
```diff
@@ -0,0 +1,98 @@
+import type { Record, User } from '@prisma/client';
+import { prisma } from '../db.server';
+
+export const createRecordInDB = async (
```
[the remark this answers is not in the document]
Genne23v: The reason I added `InDB` is due to name conflict with DNS module. What I can do is to change DB function names without `InDB` and update DNS functions like `createDomainRecord()` or `createDomain` instead. Please let me know if my suggestion helps.

package.json
Genne23v: oh, I didn't think about it. I will remove the dependency.

app/models/record.server.ts
Genne23v: Should have discussed. Will be all updated.

app/lib/client.server.ts
Genne23v: It is not a good looking code either

app/lib/client.server.ts
Genne23v: I thought same. I wasn't sure how this should be handled. I will remove these for now.

app/lib/client.server.ts
Genne23v: Yes, right. Will be rearranged

app/models/record.server.ts
Genne23v: Will be removed

app/lib/client.server.ts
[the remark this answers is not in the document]
humphd: What about `domains.server.ts`?  The other code is about "records" and this module is about user domains.

app/models/record.server.ts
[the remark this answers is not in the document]
humphd: Maybe you shouldn't import the methods, but import the module as a whole: `dns.createRecord` vs. `db.createRecord()`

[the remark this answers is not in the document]
Genne23v: oh, thanks so much. This is a great technique that I never thought of using it!

app/lib/client.server.ts
Genne23v: I wanted to ask if `hostedZone` could be multiple. I guess it won't be our concern. Will be updated accordingly.

app/models/record.server.ts
Genne23v: It's true that I started with direct `prisma` call. If we do that way, I think we would barely need any models. I like removing my model since I could reduce the code. Will remove it.

app/lib/dns.server.ts
```diff
@@ -10,14 +11,28 @@ import type {
...
 import { isIPv4, isIPv6 } from 'is-ip';
+import { secrets } from 'docker-secret';
+
+declare global {
+  var __hostedZoneId__: string;
+}
+
+const {
+  AWS_ROUTE53_HOSTED_ZONE_ID,
+  AWS_ENDPOINT_URL,
```
humphd, line 22: This doesn't need to be a secret.

app/lib/dns.server.ts
```diff
@@ -10,14 +11,28 @@ import type {
...
 } from '@aws-sdk/client-route-53';
 import { isIPv4, isIPv6 } from 'is-ip';
+import { secrets } from 'docker-secret';
+
+declare global {
+  var __hostedZoneId__: string;
+}
+
+const {
+  AWS_ROUTE53_HOSTED_ZONE_ID,
```
humphd, line 21: This isn't a secret either.  You can keep it on the env.

app/lib/dns.server.ts
```diff
@@ -10,14 +11,28 @@ import type {
   GetChangeResponse,
 } from '@aws-sdk/client-route-53';
 import { isIPv4, isIPv6 } from 'is-ip';
+import { secrets } from 'docker-secret';
+
+declare global {
```
humphd, line 16: I don't think we need to do this, using `process.env` is fine for this.

humphd: What you could do is have this module throw if the hosted zone id isn't configured in the env, and then assume it exists for the rest of the methods below.

app/lib/dns.server.ts
```diff
@@ -41,33 +56,21 @@ export const createHostedZone = async (domain: string) => {
     }
     return response.HostedZone.Id.replace('/hostedzone/', '');
   } catch (error) {
-    // TODO: fix with better logging...
-    console.warn(error);
+    logger.error(error);
```
humphd, line 59: Reserve `error` for fatal errors, and use `warn` for situations that we can recover from (like these).

app/lib/domains.server.ts
```diff
@@ -0,0 +1,148 @@
...
+import { prisma } from '~/db.server';
+import type { RecordType } from './dns.server';
+
+interface DNSProp {
+  type: RecordType;
+  name: string;
+  value: string;
+}
+
+interface DBProp {
```
humphd, line 12: You can use the `Record` type created by Prisma.  I exposes a TypeScript definition.

```ts
```

app/lib/domains.server.ts
```diff
@@ -0,0 +1,148 @@
+import * as dns from './dns.server';
+import { RecordStatus } from '@prisma/client';
+import { prisma } from '~/db.server';
+import type { RecordType } from './dns.server';
+
+interface DNSProp {
```
humphd, line 6: `DnsRecord`?

app/lib/domains.server.ts
```diff
@@ -0,0 +1,148 @@
...
+  id?: number;
+  username?: string;
+  description?: string;
+  course?: string;
+  ports?: string;
+  expiresAt?: Date;
+  status?: RecordStatus;
+}
+
+export const createUserDomain = async (dnsProp: DNSProp, dbProp: DBProp) => {
```
humphd, line 22: You could combine these two together using an [Intersection Type](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#intersection-types):

```ts
export const createUserDomain = async (data: Record & DnsRecord) => {
```

Now `data` will have to satisfy both types, meaning that you have to include everything in one object.  This might make it easier to pass the data down.

humphd: Same comment for the rest of this file.

humphd (changes requested):

app/lib/dns.server.ts
Genne23v: Will be updated

app/lib/domains.server.ts
Genne23v: oh, this is like a magic! Thank you

app/lib/domains.server.ts
Genne23v: `Record type` definition forces me to add all the properties to the data. 
```js
    const data: Record = {
      type: RecordType.A,
      name: domain,
      value,
      id: 1,
      username,
      description,
      course,
      ports,
      createdAt: new Date(),
      updatedAt: new Date(),
      expiresAt: changeMonthFromNow(6),
      status: RecordStatus.pending,
    };
```
Is it the way it should be or is there a way that I can make properties optional?

app/lib/dns.server.ts
Genne23v: Should I remove `global var`? I still have to create a hosted zone dynamically and supply `hostedZoneId` to the module for the test. I haven't found a way to feed `hostedZoneId` other than using `global var`. Would you advise how I achieve it?

app/lib/domains.server.ts
humphd: You should be able to do optional args, if the types are actually optional.  Consider:

```ts
type A = {
   one: string;
   two?: string;
}

type B = {
   three: string;
   four?: string;
}

type C = A & B;

let c: C = {
   one: 'one',
   three: 'three'
};
```

humphd: Oh, you mean the Prisma type:

```
model Record {
  id          Int          @id @default(autoincrement())
  username    String
  name        String
  type        RecordType
  value       String
  description String?
  course      String?
  ports       String?
  createdAt   DateTime     @default(now())
  updatedAt   DateTime     @updatedAt
  expiresAt   DateTime
  status      RecordStatus
  user        User         @relation(fields: [username], references: [username])
}
```

Here you can see that only a few are optional.

humphd: The other things you can do that might help:

- Create a type that has all the properties as optional: https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype
- Pick the specific properties you want, but not the others: https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys
- Omit the specific properties you want, but not the others: https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys

Using these, you should be able to use a mix of intersect, pick, omit to produce what you need.

app/lib/dns.server.ts
humphd: Yeah, I'd remove it.  Probably you're right, we should always pass it in as an arg (first arg) to all these functions at this level, and at higher levels of the API, use the value on the environment.

Genne23v: I've been struggling with this concept. I will push my update soon. I hope I understand the concept here.

Genne23v: I've been trying not to use `global var` while removing unnecessary parameter, but it seems not doable for me now. Here's my thoughts. 

`createUserDomain()` in `domains.server` calls `createRecord()`. So both `domains`, `dns` level functions need `hostedZoneId` as it's generated dynamically in testing. 

As per knowledge now, I can use `global var` and remove `hostedZoneId` parameter in all functions or add the parameter to both `dns` and `domains` level. Please let me know if I'm missing something here.

Genne23v: oh, I thought `process.env` is constant. So much learning from real world problem solving. Thank you!

app/lib/domains.server.ts
```diff
@@ -0,0 +1,146 @@
...
+      type: data.type,
+      value: data.value,
+    },
+  });
+
+  if (existingRecords.length > 0) {
+    throw new Error('Record already exists');
+  }
+
+  const changeId = await dns.createRecord(data.type, data.name, data.value);
```
humphd, line 33: You don't seem to be using this value, so maybe don't bother doing it here?

app/lib/domains.server.ts
```diff
@@ -0,0 +1,146 @@
...
+    throw new Error('Username is not provided');
+  }
+
+  const result = await prisma.record.create({
+    data: {
+      username: data.username,
+      type: data.type,
+      name: data.name,
+      value: data.value,
+      expiresAt: changeMonthFromNow(6),
```
humphd, line 49: I don't understand "change" in this name.  `setMonthsFromNow()`?

app/lib/domains.server.ts
```diff
@@ -0,0 +1,146 @@
...
+import type { RecordType } from '@prisma/client';
+
+interface DnsRecord {
+  type: RecordType;
+  name: string;
+  value: string;
+}
+
+interface Record {
+  id?: number;
```
humphd, line 13: I would use values from the `Record` type that Prisma has, to match them: `Record['id']` etc.

app/lib/domains.server.ts
```diff
@@ -0,0 +1,146 @@
...
+    });
+
+    if (!result) {
+      throw new Error('Could not delete the record in DB');
+    }
+    return result;
+  }
+};
+
+export const isDomainExpired = (expiresAt: Date) => {
```
humphd, line 135: This name is confusing, since you aren't passing a domain, but rather a date.  `isExpired`?

app/lib/domains.server.ts
```diff
@@ -0,0 +1,146 @@
...
+  return expiresAt.getTime() < new Date().getTime();
+};
+
+export const isDomainBeingExpired = (expiresAt: Date) => {
```
humphd, line 139: `willExpire(date, months=1)` maybe?

package.json
humphd: This should get removed, it looks like you still have it?

app/lib/domains.server.ts
```diff
@@ -0,0 +1,146 @@
+import * as dns from './dns.server';
+import { RecordStatus } from '@prisma/client';
+import { prisma } from '~/db.server';
+import type { RecordType } from '@prisma/client';
+
+interface DnsRecord {
```
humphd, line 6: Throughout this file, you use `Record & DnsRecord`.  I wonder if you should declare a type and use it, then export?

```ts
export type Domain = Record & DnsRecord;
```

Then all your functions below can take `(domain: Domain)`

humphd (changes requested):

app/lib/domains.server.ts
```diff
@@ -0,0 +1,146 @@
...
+interface Record {
+  id?: number;
+  username?: string;
+  description?: string;
+  course?: string;
+  ports?: string;
+}
+
+export const createUserDomain = async (data: Record & DnsRecord) => {
+  const existingRecords = await prisma.record.findMany({
```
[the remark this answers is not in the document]
Genne23v: Yes, this is definitely better.

app/lib/domains.server.ts
Genne23v: oh, I should remove them.

app/lib/domains.server.ts
Genne23v: I think that's a better name

app/lib/domains.server.ts
Genne23v: Will be updated

app/lib/domains.server.ts
Genne23v: Yeah, I think it's a better design to add `months` variable

humphd:
Let me know when I should re-review this.

Genne23v:
@humphd Yes, please review my updated PR.

humphd:
> @humphd Yes, please review my updated PR.

When you want someone to re-review, you have to click this refresh button:

Screenshot 2023-02-10 at 3 04 46 PM

app/lib/domains.server.ts
```diff
@@ -0,0 +1,138 @@
...
+      type: data.type,
+      value: data.value,
+    },
+  });
+
+  if (existingRecords > 0) {
+    throw new Error('Record already exists');
+  }
+
+  await dns.createRecord(data.type, data.name, data.value);
```
humphd, line 39: You might as well run these two operations at the same time, since they don't depend on each other:

```ts
const [result,] = await Promise.all([
  // Add to database
  prisma.record.create({
    data: {
      username: data.username,
      type: data.type,
      name: data.name,
      value: data.value,
      expiresAt: setMonthsFromNow(6),
      status: RecordStatus.pending,
    },
  }),
  // Add to Route53
  dns.createRecord(data.type, data.name, data.value);
])
```

Question: can the `dns.createRecord()` throw?  If so, we should be careful about that too.

app/lib/domains.server.ts
```diff
@@ -0,0 +1,138 @@
...
+  });
+
+  if (!result) {
+    throw new Error('Could not create a record in DB');
+  }
+  return result;
+};
+
+export const updateUserDomain = async (data: Domain) => {
+  await dns.upsertRecord(data.type, data.name, data.value);
```
humphd, line 59: Same pattern here, and probably everywhere below: you can run the db + dns async operations at the same time using `Promise.all()`

app/lib/domains.server.ts
```diff
@@ -0,0 +1,138 @@
...
+    });
+
+    if (!result) {
+      throw new Error('Could not delete the record in DB');
+    }
+    return result;
+  }
+};
+
+export const isExpired = (expiresAt: Date) => {
```
humphd, line 125: When you only return an expression, don't include a body:

```ts
export const isExpired = (expiresAt: Date) => expiresAt.getTime() < new Date().getTime();
```

app/lib/domains.server.ts
```diff
@@ -0,0 +1,138 @@
...
+  return expiresAt.getTime() < new Date().getTime();
+};
+
+export const willExpireIn = (expiresAt: Date, months: number = 1) => {
```
humphd, line 129: Same thing here:

```ts
export const willExpireIn = (expiresAt: Date, months: number = 1) =>
  new Date(expiresAt.setMonth(expiresAt.getMonth() - months)).getTime() < new Date().getTime()
```

humphd (commented):
This is really close!  A few more simplifications and optimizations.

app/lib/domains.server.ts
Genne23v: It looks simpler and cleaner

app/lib/domains.server.ts
Genne23v: It throws. In this case, is this how I should handle? 
```js
const [result, ] = Promise.all([p1.catch((error) => logger.warn(error)), p2.catch((error) => logger.warn(error))]);
```

humphd: Yeah, for the `dns.createRecord()` add a `.catch()` like you did to log the problem.

The alternative is to not deal with any of these errors, and to just let them throw.  The caller would have to `try/catch`.

Genne23v: I would like to go with the alternative. Otherwise, I have to remove many `throw` scenario tests.

humphd: Sounds good.

test/unit/dns.server.test.ts
```diff
@@ -4,125 +4,82 @@ import {
...
-    );
+    const changeId = await createRecord(RecordType.A, `my-domain.${rootDomain}`, '192.168.0.1');
     expect(changeId?.length).toBeGreaterThan(1);
   });
 
-  test('createRecord() throws an error when invalid hosted zone ID or name or value ', async () => {
-    await expect(
-      createRecord('WRONGHOSTEDZONEID', RecordType.A, 'sub2.starchart.com', '192.168.0.2')
-    ).rejects.toThrow();
+  test('createRecord() throws an error when name or value is provided ', async () => {
```
humphd, line 29: I don't understand this.  Why does it through if the name or value is provided?  Can we clarify what this test is actually testing?

test/unit/dns.server.test.ts
```diff
@@ -4,125 +4,82 @@ import {
...
+  test('upsertRecord() throws an error when invalid name or value is provided', async () => {
     await expect(
-      upsertRecord(hostedZoneId, RecordType.A, 'invalid.domain.starchart.com', '192.168.0.2')
-    ).rejects.toThrow();
-    await expect(
-      upsertRecord(hostedZoneId, RecordType.A, 'sub.starchart.com', '192-168-0-2')
-    ).rejects.toThrow();
-    await expect(
-      upsertRecord(hostedZoneId, RecordType.CNAME, 'sub.starchart.com', '')
+      upsertRecord(RecordType.A, `invalid.domain.${rootDomain}`, '192.168.0.2')
```
humphd, line 51: What makes this one invalid?

test/unit/domains.server.test.ts
```diff
@@ -0,0 +1,221 @@
...
+  deleteUserDomain,
+  willExpireIn,
+  isExpired,
+  removeIfExpired,
+  updateUserDomain,
+} from '~/lib/domains.server';
+import { prisma } from '~/db.server';
+import { RecordType } from '@prisma/client';
+
+import type { Domain } from '../../app/lib/domains.server';
```
humphd, line 14: `from '~/lib/domains.server';`

test/unit/domains.server.test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+describe('Client module function test', () => {
```
humphd, line 16: "Client" isn't a great name here.  `domains module`?

test/unit/domains.server.test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+    const result = await createUserDomain(data);
+
+    expect(result.name).toEqual(domain);
+    expect(result.type).toEqual('A');
+    expect(result.value).toEqual(value);
+    expect(result.createdAt).not.toBeNaN();
+    expect(result.updatedAt.getTime()).toEqual(result.createdAt.getTime());
+    expect(result.expiresAt.getTime()).toBeGreaterThan(result.createdAt.getTime());
+    expect(result.status.length).toBeGreaterThan(0);
```
humphd, line 52: We should be able to check for one of the known status values here, no?

test/unit/domains.server.test.ts
```diff
@@ -0,0 +1,221 @@
...
+  });
+
+  test('createUserDomain() throws when record with name, type, value already exists', async () => {
```
humphd, line 55: If your test relies on pre-existing state, that's a red flag.  You should create the state you expect inside this test (i.e., create the record first, then try to create it again).

test/unit/domains.server.test.ts
```diff
@@ -0,0 +1,221 @@
...
+    const updatedResult = await updateUserDomain(data);
+
+    expect(updatedResult.type).toEqual('AAAA');
+    expect(updatedResult.value).toEqual(updatedValue);
+    expect(updatedResult.description).toEqual(description);
+    expect(updatedResult.course).toEqual(course);
+    expect(updatedResult.ports).toEqual(ports);
+    expect(updatedResult.updatedAt.getTime()).toBeGreaterThan(result.createdAt.getTime());
+    expect(updatedResult.expiresAt.getTime()).toBeGreaterThan(result.updatedAt.getTime());
+    expect(result.status.length).toBeGreaterThan(0);
```
humphd, line 115: same comment here: check against known status values.

test/unit/domains.server.test.ts
```diff
@@ -0,0 +1,221 @@
...
+      type: RecordType.A,
+      name: `delete.${rootDomain}`,
+      value: '192.168.0.3',
+      id: 0,
+    };
+
+    await expect(deleteUserDomain(data)).rejects.toThrow();
+  });
+
+  test('isDomainExpired() deletes an expired record in Route 53 and DB ', async () => {
```
humphd, line 166: This is confusing.  What is `isDomainExpired()`?

test/unit/domains.server.test.ts
```diff
@@ -0,0 +1,221 @@
...
+  test('isDomainBeingExpired() returns true when expiresAt is less than a month from now', () => {
+    const todayPlusHalfMonth = setMonthsFromNow(0.5);
+    const todayPlusTwoMonth = setMonthsFromNow(2);
+
+    expect(willExpireIn(todayPlusHalfMonth)).toBeTruthy();
+    expect(willExpireIn(todayPlusTwoMonth)).toBeFalsy();
+  });
+});
+
+const resetHostedZone = () => {
```
humphd, line 213: Move this to the top of the file

humphd (changes requested):
Almost there...

test/unit/dns.server.test.ts
Genne23v: It was `invalid hostedZoneId or name...` I removed `invalid` by mistake. Sorry fro confusion.

test/unit/dns.server.test.ts
Genne23v: Regex expression allows only `-`. So `invalid.domain` should be `invalid-domain`. And I think valid domain should be `domain.studentName.starchart.com`. I will update the regex.

test/unit/domains.server.test.ts
Genne23v: It showed me an error when I use `~` before. Will be updated.

test/unit/domains.server.test.ts
Genne23v: Will be updated. It should be `removeIfExpired`

test/unit/domains.server.test.ts
Genne23v: Initially it should always return `pending` as `status` value is like default when it's created in DB.

app/lib/dns.server.ts
```diff
@@ -169,14 +152,13 @@ export const getChangeStatus = async (changeId: string) => {
...
     return response.ChangeInfo.Status;
   } catch (error) {
-    // TODO: fix with better logging...
-    console.warn(error);
+    logger.warn(error);
     throw new Error(`Error occurred while getting change status`);
   }
 };
 
 const isNameValid = (name: string) => {
```
humphd, line 160: Add a comment above this function that indicates what our criteria is for a valid name in plain language (i.e., so people don't have to parse your regex to understand it.

Also, according to [Route53's docs](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DomainNameFormat.html), we should be doing this:

> For domain name registration, a domain name can contain only the characters a-z, 0-9, and - (hyphen). You can't specify a hyphen at the beginning or end of a label....Amazon Route 53 stores alphabetic characters as lowercase letters (a-z), regardless of how you specify them: as uppercase letters, lowercase letters, or the corresponding letters in escape codes.

So from this I think we want:

- can't start or end with `-`
- can use `A-Za-z0-9-` (upper case is fine, but we'll convert to lower)
- can have multiple sections like this separated by `.`

What about this (untested):

```ts
/^[a-z0-9]([a-z0-9-]+\.?)+[a-z0-9]$/i
```

app/lib/domains.server.ts
```diff
@@ -0,0 +1,138 @@
...
+
+interface DomainRecord {
+  id?: Record['id'];
+  username?: Record['username'];
+  description?: Record['description'];
+  course?: Record['course'];
+  ports?: Record['ports'];
+}
+
+type Domain = DnsRecord & DomainRecord;
```
humphd, line 21: Move your `export` here:

```ts
export type Domain = DnsRecord & DomainRecord;
```

You can define and export it all at once.

humphd (approved):
Few small things, but it looks great.  If you want to move this stuff to new issues so we don't block this, that's fine too.

app/lib/dns.server.ts
dadolhay: I would probably use an external library for domain name validation, if we want to validate it ourselves, I.e.,
https://www.npmjs.com/package/validator ==> isFQDN

Reason: it is extremely hard to reliably validate a domain name, specially if I take stuff like internaltional characters in the domain name (i.e. punycode coded domain segment) and the regexes that does so are usually a mile long

humphd: Agree, it's hard to validate all possible valid domains.  I'd argue we should further limit this (e.g., I don't think we should allow punycode etc. for our use case, since it has the potential to open us to abuse).

app/lib/dns.server.ts
```diff
@@ -21,13 +30,6 @@ export const route53Client = new Route53Client({
...
 });
 
-export enum RecordType {
-  A = 'A',
-  AAAA = 'AAAA',
-  CNAME = 'CNAME',
-  TXT = 'TXT',
-}
-
 export const createHostedZone = async (domain: string) => {
```
dadolhay, line 33: Will we ever use this fn?

app/lib/dns.server.ts
dadolhay: there are so many things that is hard to validate i.e.two dashes are invalid, two dots next to each other is invalid, Do we want to offer multi level subdomains, i.e. `a.b.studentid.senecastudent.ca`, because if we do, that's not covered with the wildcard SSL cert IIRC, underscoreas also dowss have limitations IIRC

I'd probably change the signature of this:

const fn = {{subdomain ...}} => {
  // validate a single level of subdomain
  const fqdn = `${subdomain}.${process.env.PARENT_DOMAIN}`
  //  create subdomain
}

app/lib/dns.server.ts
humphd: In tests and dev only

app/lib/dns.server.ts
humphd: Forcing them to fit in the pattern `{name}.{username}.senecastudent.ca` is probably good, agree.  And let's keep the list of what's allowed in `name` to something pretty small: `a-zA-Z0-9-`.

dadolhay: If we allow `-` in the name, we need to also make sure to not start with it, and don't have multiple in a sequence like `--`

We will also need _ because of the let's encrypt challenge:
`_acme-challenge.{username}.senecastudent.ca`

Genne23v: I opened a new issue. #185 I would like to close this this PR now. @dadolhay I will soon write another PR to improve regex part.

humphd: Perfect.  Good idea to separate the regex bit so we don't block all this good work on one issue.

dadolhay (approved):
LGTM

PR #164, opened 2023-02-09
feat: verify ACME challenges and complete order

### LINKED ISSUE #21: Implement code that finishes the certificate generation process
When all challenges for the request are done, we can do the next step and actually create the certificate.

- Create a CSR
- Finalize the order
- get the certificate

- Store the appropriate data in the DB (encryption may be needed here)

dadolhay:
This PR:
- Alters the code to use the `Pebble docker container`'s mock let's encrypt server in development and testing
- Completely turns off ssl verification when `node-acme-client` communicates with the development `pebble` server (no need to inject extra CA into node)
- Alters the challenge bundle to return other challenge attributes, like `state`
- Implements `verifyChallenges` instance method that asks the ACME provider to complete each challenge
- implements `completeOrder` instance method that finalizes the order and retreives the certificate after all challenges have been completed

With these additional features, I successfully completed an order and obtained a multi-part certificate from the mock server

Closes #21


shawnyu5:
Can you change the title of the pr to something more descriptive?

dadolhay:
Sure, sorry, my bad

humphd (changes requested):
This is great.  Your PR description makes it sound like you're missing files in this change?

One question about a line in the docs I want clarified.

app/lib/lets-encrypt.server.ts
```diff
@@ -264,6 +302,93 @@ class LetsEncrypt {
...
+
+    if (['expired', 'revoked', 'deactivated', 'valid'].includes(this.#order.status)) {
+      // Do not try if the order already reached a final status
+      // https://www.rfc-editor.org/rfc/rfc8555.html#section-7.1.6
+      throw new Error(`Order found to be in the following final state: ${this.#order.status}`);
+    }
+
+    /**
+     * Validate challenges that are in the `pending` or `invalid` state
+     * It is possible retry failed challenges
```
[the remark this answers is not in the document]
dadolhay: I'm not sure how better word it. If for any reason, a challenge fails (I.e, the dns record was not added / propagated to the acme provider's DNS resolver), we have the ability to ask the provider to re-try to validate the challenge. I linked the appropriate parts of the RFC.

Could you please elaborate, what would you like to see here?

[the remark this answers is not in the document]
humphd: OK, understood now.  I'm tripping over the missing 'to.'  It should be : "It is possible TO retry failed challenges"

dadolhay:
@humphd you mean this?
> Completely turns off ssl verification when node-acme-client communicates with the development pebble server (no need to inject extra CA into node)

that part is here:
https://github.com/Seneca-CDOT/starchart/pull/164/files#diff-7e3d8f4b0066f749d73746330c16a2fcfc75298640f9eb1a39e9e81863371ba6R198

I'm reaching into the acme's default client (an instantiated axios), and overwriting it's https client with the one that does not require SSL signature to match.

This is officially supported by both `node-acme-client` and `axios` for this very reason. This way we don't need to make system wide changes to the CA list, only acme client is affected, and only in a non-production environment

humphd:
It was you saying "Alters the code to use the Pebble docker container's mock let's encrypt server in development and testing" which made me think you had altered other files.  While I'm thinking of this, do you want to delete the CA .pem file I added before, since we won't need it now?

app/lib/lets-encrypt.server.ts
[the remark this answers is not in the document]
dadolhay: yep, that's a typo, fixing

dadolhay:
sure!

app/lib/lets-encrypt.server.ts
```diff
@@ -324,7 +324,7 @@ class LetsEncrypt {
 
     /**
      * Validate challenges that are in the `pending` or `invalid` state
-     * It is possible retry failed challenges
+     * It is possible to retry failed challenges
```
humphd, line 327: Prettier will be upset with that trailing space at the end

dadolhay: yeah, used the web editor, not sure how that got there. Fixed now

app/lib/lets-encrypt.server.ts
```diff
@@ -87,6 +91,30 @@ const { LETS_ENCRYPT_ACCOUNT_PRIVATE_KEY_PEM } = secrets ?? {};
...
+ *
+ * const isChallengeSucceeded = await LetsEncrypt
+ *   .verifyChallenge({
+ *     domain: '_acme-challenge.foo.com',
+ *     key: '12345'
+ *   })
+ */
+
+/**
+ * Order
```
humphd, line 103: I love how you included this in the docs to show what it will be like, nice idea.

app/lib/lets-encrypt.server.ts
```diff
@@ -160,7 +188,16 @@ class LetsEncrypt {
       // For testing and local development, let's use an ad-hoc generated key
 
       this.#accountKey = (await acme.crypto.createPrivateKey()).toString();
-      this.#directoryUrl = acme.directory.letsencrypt.production;
+      this.#directoryUrl = 'https://127.0.0.1:14000/dir';
```
humphd, line 191: We'll have to make this configurable in a follow-up (I know you know this, just documenting).

humphd (approved):

app/lib/lets-encrypt.server.ts
dadolhay: When I keep checking the same thing over and over again, I usually include it, thinking someone else might have the same problem later

app/lib/lets-encrypt.server.ts
dadolhay: Sure, we can make an ENV for example

sfrunza13 (approved):
I have spent a while looking at this and at the [node-acme](https://github.com/publishlab/node-acme-client/blob/master/docs/client.md#AcmeClient+createOrder) and [rfc](https://www.rfc-editor.org/rfc/rfc8555#section-7.4) docs so that I could better understand it because I haven't really worked with this before. That being said the ideas in this code were evident even before that because it is well commented and written already, I went and read up again on the whole flow however because I am a noob.

So in this PR you did exactly what you set out to do. You added to the challengeBundle so you could use properties from AcmeDnsChallenge (and further the ChallengeAbstract interface), you used those attributes to check that all challenges pass and if not (I am not clear here but I think) you begin async retry on pending and invalid ones, and once order status is ready the completeOrder method will finalize the order and get the certificate.

You have also changed to a diff axios client that doesnt require ssl for ease of testing. David already mentioned the directoryUrl and I'm sure as he mentioned you already know.

I really have barely read up on this stuff (maybe 1hr and change) and can kind of follow along with the entirety of this lets-encrypt server you have written so I have nothing more to add, it's really good.

dadolhay:
Hi @sfrunza13, Thanks for the kind words

It is usually a good practice to build stuff layer to layer, so this class only focuses on communicating with Let's Encrypt.
The next piece will be to integrate this with the queue system. To do this, we need to have four stages, and three general outcome of each stage:

Stages:
- Create an order and add the challenges to DNS
- Wait for DNS to propagate (wait as needed. This is a static method, we don't communicate with the LE api to save on the rate limit)
- Request challenge verification from let's encrypt, wait for LE result and **wait / retry as needed** This step has to be able to fail and re-run
- Order completion, DNS cleanup

Outcomes:
- Success (self explanatory, next stage can commence)
- Defer (the current stage has soft-failed, I.e., DNS has not yet been propagated), so let's wait a bit and retry what we have to (some challenges may fail and some may complete for the same cert, only retry the failed ones)
- Failure (This is a hard fault, we cannot go ahead, have to completely abort the process)

So I made this code with the above in mind for example, if step 3 finds it that a challenge is still pending, it will simply return false, indicating a Defer, but if the order is in an incorrect state (the whole order failed), there is no way to continue, so it will throw an exception.

We will be able to handle / catch these in the queue worker, and depending of the above three outcomes, we can decide on what to do

This way the actual certificate generation logic, and the proccess management of it is completely separate, helps to separate the responsibilities and create a cleaner, more readable code.

humphd:
> Hi @sfrunza13, Thanks for the kind words
> 
> It is usually a good practice to build stuff layer to layer, so this class only focuses on communicating with Let's Encrypt. The next piece will be to integrate this with the queue system. To do this, we need to have four stages, and three general outcome of each stage:
> 
> Stages:
> 
> * Create an order and add the challenges to DNS
> * Wait for DNS to propagate (wait as needed. This is a static method, we don't communicate with the LE api to save on the rate limit)
> * Request challenge verification from let's encrypt, wait for LE result and **wait / retry as needed** This step has to be able to fail and re-run
> * Order completion, DNS cleanup
> 
> Outcomes:
> 
> * Success (self explanatory, next stage can commence)
> * Defer (the current stage has soft-failed, I.e., DNS has not yet been propagated), so let's wait a bit and retry what we have to (some challenges may fail and some may complete for the same cert, only retry the failed ones)
> * Failure (This is a hard fault, we cannot go ahead, have to completely abort the process)
> 
> So I made this code with the above in mind for example, if step 3 finds it that a challenge is still pending, it will simply return false, indicating a Defer, but if the order is in an incorrect state (the whole order failed), there is no way to continue, so it will throw an exception.
> 
> We will be able to handle / catch these in the queue worker, and depending of the above three outcomes, we can decide on what to do
> 
> This way the actual certificate generation logic, and the proccess management of it is completely separate, helps to separate the responsibilities and create a cleaner, more readable code.

@dadolhay seems like the start of a great blog post &lt;hint /&gt; :)

dadolhay:
@humphd 
I was already in the process of copying XD

Issue #163, opened 2023-02-09
UI Implementation for Login and Header


Ririio:
I'll be working on implementing the UI for login and header based off the [figma](https://www.figma.com/file/XAcnMIC1y4qumWinSalHI9/Starchart-UI?node-id=2105%3A2109&t=ZfBYe62w5FXC70BW-1) template. I'll be working on combining all the routes for the domain and certificates based on the skeleton page provided from #138

Eakam1007:
Hey, could you please explain what you mean by combining the routes? I am a bit confused

humphd:
@Ririio you should be able to build the UI based on the skeleton routes @Eakam1007 just landed.

Ririio:
> Hey, could you please explain what you mean by combining the routes? I am a bit confused

Oh I mean linking all the routes on the header component

Issue #46, opened 2023-01-28
Add a user domain management page


Eakam1007:
- This page will allow users to view and manage their existing domains. It should also include a link to create a new domain. 

- The page should be build using [Chakra UI components](https://chakra-ui.com/docs/components) as much as possible.

- This should be build with responsive design in mind, so the page can scale to different screen sizes.

- The page should use minimal css styles of its own. Use theme (as per #34) as much as possible instead. This is to avoid having to manually fix colors for dark mode later.

This should be done after a design has been created as per #36 .

Eakam1007:
Closing as duplicate of #137

PR #166, opened 2023-02-10
Header & Login - UI Implementation

Ririio:
This pr is based off #163 

# Implemented the UI for header #

image

- Clicking the Custom Domains takes you to the 'domains` route
- Clicking Certificate takes you to certificate route
- Clicking `My.Custom.Domains` takes you to the home page (not sure if we need a home page as discussed during one of the previous meeting)
- A dropdown menu now displays the user's first and last name, and a sign out button

# UI Login Design #

image

The 'lock' icon will be replaced with the official icon for the website

### DIFF
--- a/app/components/header.tsx
+++ b/app/components/header.tsx
@@ -1,11 +1,75 @@
-import { Text, Center } from '@chakra-ui/react';
+import {
+  Text,
+  Heading,
+  Avatar,
+  Flex,
+  HStack,
+  Menu,
+  MenuButton,
+  MenuList,
+  MenuItem,
+  IconButton,
+} from '@chakra-ui/react';
+import { TriangleUpIcon, LockIcon } from '@chakra-ui/icons';
+import { Link, useFetcher } from '@remix-run/react';
+
+import { useUser } from '~/utils';
 
 export default function Header() {
+  const user = useUser();
+  const fetcher = useFetcher();
   return (
-    <header>
-      <Center>
-        <Text fontSize="2xl">Hi! I am the header</Text>
-      </Center>
-    </header>
+    <Flex
+      as="header"
+      backgroundColor="brand.500"
+      height="auto"
+      alignItems="center"
+      shadow="2xl"
+      fontSize="18"
+      paddingY="2"
+    >
+      <HStack width="50%" color="white" justifyContent="flex-start" gap="10" marginLeft="10">
+        <Link to={{ pathname: '/domains' }}>
+          <Flex alignItems="center">
+            <TriangleUpIcon marginRight="2" />
+            <Text>Domains</Text>
+          </Flex>
+        </Link>
+
+        <Link to={{ pathname: '/certificate' }}>
+          <Flex alignItems="center">
+            <LockIcon marginRight="2" />
+            <Text>Certificate</Text>
+          </Flex>
+        </Link>
+      </HStack>
+      <Heading as="h1" size="xl" color="white">
+        My.Custom.Domain
+      </Heading>
+      <Flex
+        width="50%"
+        justifyContent="flex-end"
+        alignItems="center"
+        color="white"
+        gap="5"
+        marginRight="10"
+      >
+        <Text>{user.username}</Text>
+        <Menu>
+          <MenuButton
+            as={IconButton}
+            icon={<Avatar size="sm" />}
+            style={{ backgroundColor: 'transparent' }}
+          />
+          <MenuList color="black">
+            <MenuItem onClick={() => fetcher.submit({}, { method: 'post', action: '/logout' })}>
+              <Text fontSize="sm" color="brand.500">
+                Sign Out
+              </Text>
+            </MenuItem>
+          </MenuList>
+        </Menu>
+      </Flex>
+    </Flex>
   );
 }
--- a/app/routes/__index/index.tsx
+++ b/app/routes/__index/index.tsx
@@ -1,9 +1,9 @@
-import { Heading, Text, Button } from '@chakra-ui/react';
-import { json } from '@remix-run/node';
-import { Form, useLoaderData } from '@remix-run/react';
-import { requireUsername } from '~/session.server';
-
+import { Heading, Text } from '@chakra-ui/react';
 import type { LoaderArgs } from '@remix-run/node';
+import { json } from '@remix-run/node';
+
+import { requireUsername } from '~/session.server';
+import { useUser } from '~/utils';
 
 export const loader = async ({ request }: LoaderArgs) => {
   const username = await requireUsername(request);
@@ -12,18 +12,14 @@ export const loader = async ({ request }: LoaderArgs) => {
 };
 
 export default function IndexRoute() {
-  const data = useLoaderData<typeof loader>();
-
+  const user = useUser();
   return (
     <div>
       <Heading as="h1" size="3xl" noOfLines={1}>
         Starchart
       </Heading>
       <Text fontSize="xl">Start making your own custom domains and certificates today!</Text>
-      <Text>Welcome {data.username}</Text>
-      <Form action="/logout" method="post">
-        <Button type="submit">Logout</Button>
-      </Form>
+      <Text>Welcome {user.username}</Text>
     </div>
   );
 }
--- a/app/routes/login.tsx
+++ b/app/routes/login.tsx
@@ -1,8 +1,9 @@
-import { Heading, Text, Button } from '@chakra-ui/react';
-import { Form } from '@remix-run/react';
-import { createUserSession } from '~/session.server';
-
+import { Heading, Text, Button, Center, Grid, GridItem, VStack, Flex } from '@chakra-ui/react';
+import { LockIcon } from '@chakra-ui/icons';
 import type { ActionArgs } from '@remix-run/node';
+import { Form } from '@remix-run/react';
+
+import { createUserSession } from '~/session.server';
 
 export const action = async ({ request }: ActionArgs) => {
   return createUserSession({
@@ -15,10 +16,31 @@ export const action = async ({ request }: ActionArgs) => {
 
 export default function Login() {
   return (
-    <Form method="post">
-      <Heading>Login</Heading>
-      <Text>You will be logged in (a session will be created for you)</Text>
-      <Button type="submit">Login</Button>
-    </Form>
+    <Center height="container.md" marginTop="16">
+      <Grid gap="12">
+        <GridItem width="lg">
+          <VStack marginTop="20" spacing="10">
+            <Heading size="2xl" color="brand.500">
+              My.Custom.Domain
+            </Heading>
+            <Text fontSize="3xl">Simple, Secure, DNS for Seneca</Text>
+          </VStack>
+        </GridItem>
+        <GridItem border="solid" borderRadius="2xl" borderColor="brand.500">
+          <VStack height="2xs">
+            <Flex width="100px" height="100px" marginTop="10">
+              <LockIcon color="grey" boxSize="100%" />
+            </Flex>
+            <Flex flex={1} alignItems="center" justifyContent="center">
+              <Form method="post">
+                <Button type="submit" width="xs">
+                  Sign In
+                </Button>
+              </Form>
+            </Flex>
+          </VStack>
+        </GridItem>
+      </Grid>
+    </Center>
   );
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@aws-sdk/client-route-53": "^3.261.0",
+        "@chakra-ui/icons": "^2.0.17",
         "@chakra-ui/react": "^2.4.9",
         "@emotion/react": "^11.10.5",
         "@emotion/server": "^11.10.0",
@@ -3215,6 +3216,18 @@
       "integrity": "sha512-RpA1X5Ptz8Mt39HSyEIW1wxAz2AXyf9H0JJ5HVx/dBdMZaGMDJ0HyyPBVci0m4RCoJuyG1HHG/DXJaVfUTVAeg==",
       "dependencies": {
         "@chakra-ui/shared-utils": "2.0.5"
+      },
+      "peerDependencies": {
+        "@chakra-ui/system": ">=2.0.0",
+        "react": ">=18"
+      }
+    },
+    "node_modules/@chakra-ui/icons": {
+      "version": "2.0.17",
+      "resolved": "https://registry.npmjs.org/@chakra-ui/icons/-/icons-2.0.17.tgz",
+      "integrity": "sha512-HMJP0WrJgAmFR9+Xh/CBH0nVnGMsJ4ZC8MK6tMgxPKd9/muvn0I4hsicHqdPlLpmB0TlxlhkBAKaVMtOdz6F0w==",
+      "dependencies": {
+        "@chakra-ui/icon": "3.0.16"
       },
       "peerDependencies": {
         "@chakra-ui/system": ">=2.0.0",
@@ -23784,6 +23797,14 @@
       "integrity": "sha512-RpA1X5Ptz8Mt39HSyEIW1wxAz2AXyf9H0JJ5HVx/dBdMZaGMDJ0HyyPBVci0m4RCoJuyG1HHG/DXJaVfUTVAeg==",
       "requires": {
         "@chakra-ui/shared-utils": "2.0.5"
+      }
+    },
+    "@chakra-ui/icons": {
+      "version": "2.0.17",
+      "resolved": "https://registry.npmjs.org/@chakra-ui/icons/-/icons-2.0.17.tgz",
+      "integrity": "sha512-HMJP0WrJgAmFR9+Xh/CBH0nVnGMsJ4ZC8MK6tMgxPKd9/muvn0I4hsicHqdPlLpmB0TlxlhkBAKaVMtOdz6F0w==",
+      "requires": {
+        "@chakra-ui/icon": "3.0.16"
       }
     },
     "@chakra-ui/image": {
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   },
   "dependencies": {
     "@aws-sdk/client-route-53": "^3.261.0",
+    "@chakra-ui/icons": "^2.0.17",
     "@chakra-ui/react": "^2.4.9",
     "@emotion/react": "^11.10.5",
     "@emotion/server": "^11.10.0",


app/components/header.tsx
```diff
@@ -1,11 +1,66 @@
...
 
 export default function Header() {
+  const user = useUser();
   return (
     <header>
-      <Center>
-        <Text fontSize="2xl">Hi! I am the header</Text>
-      </Center>
+      <Box
+        display="flex"
```
humphd, line 12: You can use the `Flex` component vs. `Box` for this.

app/components/header.tsx
```diff
@@ -1,11 +1,66 @@
...
+        backgroundColor="brand.500"
+        height="70px"
+        alignItems="center"
+        shadow="2xl"
+        border="2px"
+        fontSize="18"
+      >
+        <Box
```
humphd, line 20: Use an `HStack` for this, with a `gap`

app/components/header.tsx
```diff
@@ -1,11 +1,66 @@
...
+          <Link to={{ pathname: '/certificate' }}>
+            <Box display="flex" alignItems="center">
+              <LockIcon marginRight="2" />
+              <Text>Certificate</Text>
+            </Box>
+          </Link>
+        </Box>
+        <Link to={{ pathname: '/' }}>
+          <Heading as="h1" size="2xl" color="white">
+            My.Custom.Domains
```
humphd, line 44: Domain (singular)

app/components/header.tsx
```diff
@@ -1,11 +1,66 @@
...
+          justifyContent="flex-end"
+          alignItems="center"
+          color="white"
+          gap="5"
+          marginRight="10"
+        >
+          <Text>{user.username}</Text>
+          <Form action="/logout" method="post">
+            <Button _hover={{ bg: '0' }} size="2xs" type="submit">
+              <Avatar size="md" name={user.firstName}></Avatar>
```
humphd, line 59: Because we'll never have an icon, let's use a generic icon instead of the initials, something like these https://chakra-ui.com/docs/components/avatar#customize-the-fallback-avatar

app/routes/login.tsx
```diff
@@ -15,10 +15,45 @@ export const action = async ({ request }: ActionArgs) => {
 
 export default function Login() {
   return (
-    <Form method="post">
-      <Heading>Login</Heading>
-      <Text>You will be logged in (a session will be created for you)</Text>
-      <Button type="submit">Login</Button>
-    </Form>
+    <Center height="container.md" marginTop="28">
```
humphd, line 18: `h=... mt=...`

app/routes/login.tsx
```diff
@@ -15,10 +15,45 @@ export const action = async ({ request }: ActionArgs) => {
...
+            <Heading size="3xl">My</Heading>
+            <Heading size="3xl" marginLeft="16">
+              Custom
+            </Heading>
+            <Heading size="3xl" marginLeft="48">
+              Domain
+            </Heading>
+          </Box>
+          <Text fontSize="3xl" marginTop="10">
+            Reliable DNS serving from Seneca to students
```
humphd, line 31: What about "Simple, Secure, DNS for Seneca"?

app/routes/login.tsx
```diff
@@ -15,10 +15,45 @@ export const action = async ({ request }: ActionArgs) => {
...
-    <Form method="post">
-      <Heading>Login</Heading>
-      <Text>You will be logged in (a session will be created for you)</Text>
-      <Button type="submit">Login</Button>
-    </Form>
+    <Center height="container.md" marginTop="28">
+      <Grid templateColumns="repeat(2, 1fr)">
+        <GridItem width="lg" height="xs">
+          <Box color="brand.500">
+            <Heading size="3xl">My</Heading>
```
humphd, line 22: Instead of these waterfall boxes, can you do the same `my.custom.domain` concept here?

app/routes/login.tsx
```diff
@@ -15,10 +15,45 @@ export const action = async ({ request }: ActionArgs) => {
...
+              display="flex"
+              alignItems="center"
+              justifyContent="center"
+              width="100px"
+              height="100px"
+              marginTop="16"
+            >
+              Icon
+            </Box>
+            <Box flex={1} display="flex" alignItems="center" justifyContent="center">
```
humphd, line 47: Use `Flex` if you want flex?

humphd (changes requested):
I'm really excited to see this happening, nice work @Ririio.  I've left some initial comments

app/components/header.tsx
Ririio: That's a good call, I forgot they had a flex component

app/routes/login.tsx
Ririio: I think using the actual word as to the key characters is easier for readability

app/routes/login.tsx
Ririio: Like a straight line "My.Custom.Domain"?

app/routes/login.tsx
Ririio: Yeah that sounds better, and more professional

app/routes/login.tsx
Ririio: I kind of like the waterfall aesthetic, makes it looks proportional to the container, not sure though we can ask the others for input

app/routes/login.tsx
Ririio: Yeah I actually forgot that Flex component was a thing for chakra

app/routes/login.tsx
humphd: Yes, straight line vs. waterfall.  I'm not a fan of the waterfall.

Ririio: image

I've made changes with the layout, and made it vertical align instead to make it more straight to the point when a user views the page

app/components/header.tsx
Ririio: image

I made changes accordingly let me know what you think

app/routes/login.tsx
humphd: Nice!  For now, what if we use the `LockIcon` for the logo, so we have something?

Screenshot 2023-02-10 at 9 47 41 AM

app/routes/__index/index.tsx
```diff
@@ -1,9 +1,10 @@
 import { Heading, Text, Button } from '@chakra-ui/react';
 import { json } from '@remix-run/node';
-import { Form, useLoaderData } from '@remix-run/react';
+import { Form } from '@remix-run/react';
```
humphd, line 3: You can remove this if Form isn't being used.

humphd (dismissed):

app/components/header.tsx
```diff
@@ -1,11 +1,57 @@
...
+        <Flex
+          width="50%"
+          justifyContent="flex-end"
+          alignItems="center"
+          color="white"
+          gap="5"
+          marginRight="10"
+        >
+          <Text>{user.username}</Text>
+          <Form action="/logout" method="post">
```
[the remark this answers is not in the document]
Ririio: I recall telescope followed the same format or am I mistaken?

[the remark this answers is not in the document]
Ririio: image

I replaced the previous iteration with a menu. Now the right side of the header should display the `username` and a `menu`. The menu will display the full name of the user capped at 17 characters, and a sign out button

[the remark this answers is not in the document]
humphd: The user will know their name, I'm not sure we gain anything by showing it, esp. when it has to be cut-off like this.  I would just show the `Sign Out` menu item.

What is the purpose of showing the generic avatar icon a second time in the menu?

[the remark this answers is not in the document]
Ririio: The icon makes the menu more filled out, and the name is there to replaced the username, because I think it's better to have it outside the menu. I'll re-write it now to only have the sign out button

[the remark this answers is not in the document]
humphd: I like the idea of the "filled-out" menu, but we have so little data to use, I think it ends up looking more empty by repeating.  Here's what GitHub does:

Screenshot 2023-02-10 at 12 43 08 PM

In contrast to Teams and Slack, which use the image to better effect:

Screenshot 2023-02-10 at 12 42 48 PM
Screenshot 2023-02-10 at 12 42 41 PM

I don't know if we'll ever have more than "Sign Out" or not?  But for sure we'll never have an image for all users, so I wouldn't rely on a design that tries to use it.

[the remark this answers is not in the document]
humphd: Maybe we can put the "ITS Acceptable Use Policy" link in there, or help links, not sure...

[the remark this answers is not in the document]
Ririio: Yeah you do have a point, I replaced it with a single menu item
image

app/components/header.tsx
```diff
@@ -1,11 +1,78 @@
...
+  const user = useUser();
+  const fetcher = useFetcher();
   return (
     <header>
-      <Center>
-        <Text fontSize="2xl">Hi! I am the header</Text>
-      </Center>
+      <Flex
+        backgroundColor="brand.500"
+        height="70px"
```
Myrfion, line 25: Is there any good reason to set static height for the header? Maybe just leave it `auto` and just add horizontal padding if you wanna increase the height?

app/components/header.tsx
```diff
@@ -1,11 +1,78 @@
...
+        fontSize="18"
+      >
+        <HStack width="50%" color="white" justifyContent="flex-start" gap="10" marginLeft="10">
+          <Link to={{ pathname: '/domains' }}>
+            <Flex alignItems="center">
+              <TriangleUpIcon marginRight="2" />
+              <Text>Custom Domain</Text>
+            </Flex>
+          </Link>
+
```
Myrfion, line 37: empty line

app/components/header.tsx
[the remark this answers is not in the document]
humphd: We could consider moving "Domains" and "Certificate" into that menu as well, which is kind of what GitHub does.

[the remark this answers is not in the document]
Ririio: That would leave the header less populated, unless you're referring to adding another menu for only the two which is similar to the original figma design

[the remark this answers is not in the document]
humphd: I mean adding them to the menu you're making:

```
Domains
Certificate
-----------
Sign out
```

We don't have to put it there, just thinking out loud.

Myrfion:
I think the black border for the header is not very on point and doesn't look really good, and it is better to remove it:
image

app/components/header.tsx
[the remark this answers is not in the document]
Eakam1007: Hmm, I think that would make add an extra step to the navigation. I think if there are more pages to navigate to, it would be good. But we only have domains and certificate

app/components/header.tsx
```diff
@@ -1,11 +1,78 @@
...
+        alignItems="center"
+        shadow="2xl"
+        border="2px"
+        fontSize="18"
+      >
+        <HStack width="50%" color="white" justifyContent="flex-start" gap="10" marginLeft="10">
+          <Link to={{ pathname: '/domains' }}>
+            <Flex alignItems="center">
+              <TriangleUpIcon marginRight="2" />
+              <Text>Custom Domain</Text>
```
Eakam1007, line 35: Should this be `Custom Domains` instead of `Custom Domain`?

humphd: Or better: `Domains`.  All domains are custom, so it's redundant and takes more space.

app/routes/__index/index.tsx
```diff
@@ -12,18 +12,14 @@ export const loader = async ({ request }: LoaderArgs) => {
 };
 
 export default function IndexRoute() {
-  const data = useLoaderData<typeof loader>();
-
+  const user = useUser();
   return (
     <div>
```
Myrfion, line 17: Need to wrap it up in Chakra container, something like
```typescript
<Container maxW="container.xl">
```

app/components/header.tsx
Eakam1007: Yea, that is much better

Ririio: image
Should I indent it more or should I keep it as is?

Eakam1007: That looks good to me

app/routes/login.tsx
```diff
@@ -15,10 +16,31 @@ export const action = async ({ request }: ActionArgs) => {
...
   return (
-    <Form method="post">
-      <Heading>Login</Heading>
-      <Text>You will be logged in (a session will be created for you)</Text>
-      <Button type="submit">Login</Button>
-    </Form>
+    <Center height="container.md" marginTop="16">
+      <Grid gap="12">
+        <GridItem width="lg">
+          <VStack marginTop="20" spacing="10">
```
Myrfion, line 22: better use theme spacings, instead of putting exact values in pixels:
https://chakra-ui.com/docs/styled-system/style-props#margin-and-padding

So instead of `marginTop="20"` it would be something like `mt="4"`

app/routes/login.tsx
```diff
@@ -15,10 +16,31 @@ export const action = async ({ request }: ActionArgs) => {
...
+            <Heading size="2xl" color="brand.500">
+              My.Custom.Domain
+            </Heading>
+            <Text fontSize="3xl">Simple, Secure, DNS for Seneca</Text>
+          </VStack>
+        </GridItem>
+        <GridItem border="solid" borderRadius="2xl" borderColor="brand.500">
+          <VStack height="2xs">
+            <Flex width="100px" height="100px" marginTop="10">
```
Myrfion, line 31: same on spacings I said before

app/routes/login.tsx
```diff
@@ -15,10 +16,31 @@ export const action = async ({ request }: ActionArgs) => {
 
 export default function Login() {
   return (
-    <Form method="post">
-      <Heading>Login</Heading>
-      <Text>You will be logged in (a session will be created for you)</Text>
-      <Button type="submit">Login</Button>
-    </Form>
+    <Center height="container.md" marginTop="16">
```
Myrfion, line 19: same about spacings, would be like `mt="4"`

app/components/header.tsx
```diff
@@ -1,11 +1,77 @@
...
+            My.Custom.Domain
+          </Heading>
+        </Link>
+        <Flex
+          width="50%"
+          justifyContent="flex-end"
+          alignItems="center"
+          color="white"
+          gap="5"
+          marginRight="10"
```
Myrfion, line 56: same on spacings here, would be like `mr="2"`

app/components/header.tsx
```diff
@@ -1,11 +1,77 @@
...
+        >
+          <Text>{user.username}</Text>
+          <Menu>
+            <MenuButton as={IconButton} icon={<Avatar size="sm" />} />
+            <MenuList color="black">
+              <MenuItem
+                onClick={() => {
```
Myrfion, line 63: since it is one line function, you could just do 
```typescript
() => fetcher.submit({}, { method: 'post', action: '/logout' })
```
Not need for `{}`

app/components/header.tsx
```diff
@@ -1,11 +1,77 @@
...
+import { TriangleUpIcon, LockIcon } from '@chakra-ui/icons';
+import { Link, useFetcher } from '@remix-run/react';
+
+import { useUser } from '~/utils';
 
 export default function Header() {
+  const user = useUser();
+  const fetcher = useFetcher();
   return (
     <header>
```
Myrfion, line 22: You can remove `header` and instead just add property `as="header"` for Flex which is a child of `header`. It gonna make Flex render as `header`

Myrfion (changes requested):

app/components/header.tsx
Myrfion: Also, I think it would be good to wrap header's content in `Container` too, so it's content would align with content of pages

app/routes/__index/index.tsx
Ririio: I'm not sure about touching index.tsx, because as far as I recall we won't have a home page according to one of our previous meeting

Eakam1007: I think this should be replaced by the domains page (as per https://github.com/Seneca-CDOT/starchart/issues/46)

app/components/header.tsx
Ririio: Can't we just set it to break points if we need to manipulate the value of the height? Something like this?
image

Myrfion: I don't see any good reason why we would need to manipulate the value of the height, in a way that we set it statically, it could just stretch depending on its content height + you could put some vertical padding `px="2"` or something

app/routes/login.tsx
Ririio: Are you perhaps referring to adding onto the theme folder with predetermined values as to the default provided by chakra? I never specified the exact pixel, I only used the values that is set on default. 

image

Myrfion: Chakra already has a pretty decent list of predefined values, but of course, we could extend it if we need to

Ririio: I am using their predefined values, I'm a bit confused as to what I did wrong? `marginTop='20'` is similar to `marginTop='5rem'. It's a default value set out by chakra

Myrfion: Oh, right, I just got confused, because supposed that a fully-written attributes, such as `marginTop` get pixels, while if you need to use theme value you put shortcut like `mt`. My bad, than all good, I would just recommend to use shortcuts to avoid making huge vertical attributes list

Ririio: I only use the full word for readability...

app/components/header.tsx
Myrfion: It is better to avoid settings up static height to avoid cases when your content's height is bigger than the container height, let's say if on a mobile device logo wouldn't fit in one line and would go for another one, but your static height is designed to fit only one line, so it would be overflown.

app/components/header.tsx
Ririio: > You can remove header and instead just add property as="header" for Flex which is a child of header. It gonna make Flex render as header

oh didn't know I can do something like that

> Also, I think it would be good to wrap header's content in Container too, so it's content would align with content of pages

Isn't the header flex box the container? They're already align through the `alignItems` and the percentage of width that they take

app/components/header.tsx
Myrfion: And regarding setting padding (I actually made mistake, it should be something like `py="2"`, since we talk about the height) you could just set vertical padding for container and it would be alright for all screens, while with the height you need to adjust the height for all or some breakpoints

Ririio: Ohh that make sense, I'll remember that

app/components/header.tsx
Myrfion: I mean aligning the header content with page content verticly:
image

Ririio: I'm a bit confused as to what you mean sorry...

Myrfion: No problem, I will try to explain it deeply.

### Issue 
In my opinion, the content of the header (so custom domains button, certificates etc) should be aligned with the content of the pages that we gonna have). 

Right now it is not aligned:
image
You could see that content starts before the yellow line

### What I propose
Wrap both inner contents for header content of all pages in a container with the same `maxW` (probably `container.xl`, so they would be aligned with each other with all possible screen width

Myrfion: https://chakra-ui.com/docs/components/container/usage#container-size

Ririio: Ohh... so you want to remove the spacing on the left to make it more align with the page? Or at least combine the two so they'll be aligned together, am I understanding it right?

Myrfion: I don't say that they should be in the same container, we should just assign their containers with the same maxWidth properties and center them in their containers

Eakam1007: Isn't the header already aligned then? It is the pages that are not aligned with the header?

app/components/header.tsx
```diff
@@ -1,11 +1,73 @@
...
+        </Link>
+
+        <Link to={{ pathname: '/certificate' }}>
+          <Flex alignItems="center">
+            <LockIcon marginRight="2" />
+            <Text>Certificate</Text>
+          </Flex>
+        </Link>
+      </HStack>
+      <Link to={{ pathname: '/' }}>
```
Eakam1007, line 46: I think we should remove this, since the home page will be the domain page anyway. Feels kinda redundant with the domain link. What do you think?

Ririio: Wouldn't we have to replace it later on, because the domain page will became our new index page? Or do we just redirect them to the domain page when they're at `/`?

Eakam1007: Yeah, we will probably need to change the index page to redirect to the domain page. I just mean that the logo linking to anything is kinda redundant. It could just be plain text

Ririio: oh that's what you meant, yeah sure I'll remove it

Eakam1007 (approved):
These pages look great! 
Can you file an issue for writing tests for this; for example, the header links?

dadolhay (approved):
LGTM

SerpentBytes (approved):

dadolhay:
Merging by @Ririio 's request (he is not at a computer so asked me to do so)

PR #174, opened 2023-02-11
Getting ready for IDP + SP work

### LINKED ISSUE #105: Add and configure SAML2 Identity Provider Container for testing
The work to add SAML support is going to require a development Identity Provider (IdP) that can mock Seneca's SSO.  Probably the easiest way to do this is to copy what we already have in Telescope, where we use the [test-saml-idp](https://github.com/Seneca-CDOT/telescope/blob/master/docker/development.yml#L36-L49) container, [configure it](https://github.com/Seneca-CDOT/telescope/blob/master/config/saml20-idp-hosted.php), and specify [users](https://github.com/Seneca-CDOT/telescope/blob/master/config/simplesamlphp-users.php).

The base image is not being maintained, but I don't know if we care?  There are some other forks that have updated it, but they seem mostly abandoned too.

sfrunza13:
Small PR to add Samlify and the Simple SAML IDP exactly as it is in telescope, not yet configured to work together.

Closes #105


config/simplesamlphp-users.php
```diff
@@ -0,0 +1,48 @@
...
+            'eduPersonAffiliation' => array('group1'),
+            /*
+              NOTE: we add a bunch of claims that we expect to get back from Seneca,
+              and need to simulate here.
+            */
+            'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress' => 'user1@example.com',
+            'email' => 'user1@example.com',
+            'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname' => 'Johannes',
+            'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname' => 'Kepler',
+            'http://schemas.microsoft.com/identity/claims/displayname' => 'Johannes Kepler'
```
humphd, line 26: We better add a few more claims, since we will need a `username`:

```
  "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name": "username@seneca-domain.ca",
  "sAMAccountName": "username"
```

Give all 3 users unique usernames that match their email address.

Since you're creating this setup, you also get to name the users.  If you want to immortalize yourself, now is your chance!

humphd (changes requested):
One small thing, and this is ready to go.

We should probably also include a doc update to let people know what the username/passwords are, and give instructions about how to access the login app in development and testing.

config/simplesamlphp-users.php
sfrunza13: Should the email and name attributes have the same values? Our Seneca e-mail account should be used for both right?

sfrunza13: Also perhaps I should ask, cause I am not clear about it. What is sAMAccountName? It seems as though from my searching it is an older username that is of the form domain\username as opposed to the email-like user@domain.com that we use now. Do we all have these? Are these created by default when an account is made using the prefix of the user@domain.com users?

config/simplesamlphp-users.php
```diff
@@ -0,0 +1,55 @@
...
+    'example-userpass' => array(
+        'exampleauth:UserPass',
+        'user1:user1pass' => array(
+            'uid' => array('1'),
+            'eduPersonAffiliation' => array('group1'),
+            /*
+              NOTE: we add a bunch of claims that we expect to get back from Seneca,
+              and need to simulate here.
+            */
+            'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress' => 'user1@example.com',
```
humphd, line 22: Change the email address to match the `'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name'` claim (same email in all claims).  Do this for every account.

config/simplesamlphp-users.php
```diff
@@ -0,0 +1,55 @@
...
+            'email' => 'user1@example.com',
+            'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname' => 'Johannes',
+            'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname' => 'Kepler',
+            'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name'=> 'HeresJohnny@myseneca.ca',
+            'sAMAccountName'=> 'HeresJohnny',
```
humphd, line 27: Seneca requires all lowercase, so let's use lowercase for this and email name portion.  Same for all accounts.

humphd (changes requested):

humphd (approved):
I'm fine with this, but can you file another issue to add these users to our Prisma seed?  We should probably create at least 2 of them in the db.  We might want to leave one out, so we can test the first-run flow.

dadolhay (approved):

PR #181, opened 2023-02-12
Add domain table to domains list page

### LINKED ISSUE #137: Feat: Domains Table
### Overview
This task aims to create a component named `DomainsList.tsx` that will render domains list and trigger certain actions.  

### Design
Refer to Figma for the design

### Implementation
Table should be done with Chakra UI table: https://chakra-ui.com/docs/components/table

Props should look something like:
```typescript
interface DomainsListProps {
 domains: Domain[]
 onAction: (action: DomainsListAction) => void
}

type DomainsListAction = 'EDIT' | 'DELETE' | 'RENEW'  
```

Myrfion:
### Overview
Added `DomainsTable` component, that takes `domains` and `onAction` props and renders Chakra UI table with domains. It also triggers `onAction` callback with different actions. Only `EDIT` action handling is implemented for now, since I need more back-end stuff updates to implement other actions handling. I also added mock file with some domains, just to have something to render for now.

Resolves #137

image

### DIFF
--- a/app/components/domains-table.tsx
+++ b/app/components/domains-table.tsx
@@ -45,34 +45,35 @@ export default function DomainsTable(props: DomainsTableProps) {
   }
 
   function renderDomainStatus(action: RecordStatus) {
-    switch (action) {
-      case 'active':
-        return (
-          <Tooltip label="Domain is live">
-            <CheckCircleIcon color="green.500" boxSize="6" />
-          </Tooltip>
-        );
-      case 'error':
-        return (
-          <Tooltip label="Domain error">
-            <WarningIcon color="brand.500" boxSize="6" />
-          </Tooltip>
-        );
-      case 'pending':
-        return (
-          <Tooltip label="Domain is pending">
-            <Flex
-              width="6"
-              height="6"
-              borderRadius="20"
-              alignItems="center"
-              justifyContent="center"
-              background="yellow.500"
-            >
-              <TimeIcon color="white" boxSize="3" />
-            </Flex>
-          </Tooltip>
-        );
+    if (action === 'active') {
+      return (
+        <Tooltip label="Domain is live">
+          <CheckCircleIcon color="green.500" boxSize="6" />
+        </Tooltip>
+      );
+    }
+    if (action === 'error') {
+      return (
+        <Tooltip label="Domain error">
+          <WarningIcon color="brand.500" boxSize="6" />
+        </Tooltip>
+      );
+    }
+    if (action === 'pending') {
+      return (
+        <Tooltip label="Domain is pending">
+          <Flex
+            width="6"
+            height="6"
+            borderRadius="20"
+            alignItems="center"
+            justifyContent="center"
+            background="yellow.500"
+          >
+            <TimeIcon color="white" boxSize="3" />
+          </Flex>
+        </Tooltip>
+      );
     }
   }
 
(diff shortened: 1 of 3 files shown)


app/components/domains-table.tsx
```diff
@@ -0,0 +1,147 @@
...
+                          aria-label="Refresh domain"
+                          variant="ghost"
+                          ml="2"
+                          onClick={() => onCopyNameToClipboard(domain.name)}
+                        />
+                      </Tooltip>
+                    </Flex>
+                  </Td>
+                  <Td>{domain.type}</Td>
+                  <Td>{domain.value}</Td>
```
Ririio, line 113: Do we have a maxWidth for each column?

app/routes/__index/domains/index.tsx
```diff
@@ -1,11 +1,32 @@
...
+    }
+  }
+
   return (
-    <div>
-      <Heading as="h1" size="3xl">
+    <Container maxW="contianer.xl">
+      <Heading as="h1" size="2xl" mt="8">
         Domains List
       </Heading>
```
Ririio, line 28: We might need to lower the size of this header, I don't know if it's the way the photo was taken, but it looks way too big

Ririio: Should we lower the size of the container? I just think that page looks a bit cramp (maybe that's what we should be going for? Not sure...)

app/components/domains-table.tsx
Myrfion: I think that it is just better to let Chakra decide the size, if the table doesn't cover fit in the screen it just gets horizontally scrollable

app/routes/__index/domains/index.tsx
Myrfion: I guess using `container.xl` size for max width is optimal for page container, don't see a good reason to use a smaller one and have more whitespace

Myrfion: And regarding header, I already lower the size that was before me, can reduce it even more

Myrfion: he is how it looks with xl size instead of 2xl, better to me 

image

app/components/domains-table.tsx
Ririio: Can't we set a max character for the domain attributes, and have a tooltip for users to hover so they can view the entire thing? Should make the table look more clean no?

app/routes/__index/domains/index.tsx
Ririio: `xl` looks nicer, btw are we going to add a description for this like we do on Figma?

Myrfion: Yea, something like this

image

Myrfion: I can add it to this PR too

Ririio: We probably should ask @humphd as to what we should have on the description (if needed) so we can implement it now on your pr

Myrfion: I also feel like we should drop "List" from Domains List, @Ririio what do you think?

Ririio: I agree we can go with just 'Domains' so we follow the header format

app/components/domains-table.tsx
Myrfion: Since we have entire page width for a table I don't see a big need to cut something off + value is the kinda thing that users might need to check frequently and this is pretty much one of the most valuable fields, so again don't see a good reason to hide it

Ririio: I'm not saying to hide it, but to lessen the characters to make the table more aligned? A simple '...' at the end should imply that there's more to the text no?

app/components/domains-table.tsx
```diff
@@ -0,0 +1,147 @@
...
+              <Th>Name</Th>
+              <Th>Type</Th>
+              <Th>Value</Th>
+              <Th>Expiration date</Th>
+              <Th />
+            </Tr>
+          </Thead>
+          <Tbody>
+            {domains.map((domain) => {
+              return (
```
dadolhay, line 95: Nit:
```suggestion
            {domains.map((domain) => (
```

...you need to update the end too to match, this is just a tiny bit cleaner

app/components/domains-table.tsx
```diff
@@ -0,0 +1,147 @@
...
+    navigator.clipboard.writeText(name);
+    toast({
+      title: 'Name was copied to clipboard',
+      position: 'bottom-right',
+      status: 'success',
+    });
+  }
+
+  function renderDomainStatus(action: RecordStatus) {
+    switch (action) {
```
dadolhay, line 48: Nit:

You can do this with an early return style:
```
if (action === 'active'){
  return (...)
}

if (action === 'error'){
  return (...)
}
```

Reason:
It is usually considered a bad practice to use a switch case, because there is too much risk of the execution overflowing to the next case ... just an unclean language element

app/mocks/domains.ts
```diff
@@ -0,0 +1,48 @@
...
+    value: '2001:db8:3333:4444:CCCC:DDDD:EEEE:FFFF',
+    description: 'lorem ipsum',
+    createdAt: new Date(),
+    updatedAt: new Date(),
+    expiresAt: new Date(),
+    status: 'pending',
+    course: null,
+    ports: null,
+  },
+];
```
dadolhay, line 46: Nit:
Just for kicks, could you please add a TXT record as well, so we have one of each type?

dadolhay (approved):
Nice work, only added a few nit-pick stuff

Myrfion:
@dadolhay Did updates on your comments, I appreciate the feedback :)

Eakam1007:
A few questions:

- Should this also include the `Create new domain` button ([figma design](https://www.figma.com/file/XAcnMIC1y4qumWinSalHI9/Starchart-UI?node-id=2105%3A2109&t=wXyoDUqrTy1X5efd-1))
  
  image
  
- Would it make sense to also replace the index route with the domains page, as we don't have a home page? We could file a separate issue for this as well (As mentioned [here](https://github.com/Seneca-CDOT/starchart/pull/166#discussion_r1103077429))

- I remember you mentioned aligning the header with the content of the pages [here](https://github.com/Seneca-CDOT/starchart/pull/166#discussion_r1103160175). Should `Domains` heading on the page be aligned with the header for this page too? I am not sure if this is needed here though

  image

cychu42:
I recall discussing about adding tooltips to each column for clarification of what they are. Is that still the plan? I don't see them (or I might be blind).

app/components/domains-table.tsx
```diff
@@ -0,0 +1,146 @@
...
+                        onClick={() => onCopyNameToClipboard(domain.name)}
+                      />
+                    </Tooltip>
+                  </Flex>
+                </Td>
+                <Td>{domain.type}</Td>
+                <Td>{domain.value}</Td>
+                <Td>
+                  <Flex justifyContent="space-between" alignItems="center">
+                    {domain.expiresAt.toLocaleDateString('en-US')}
```
humphd, line 116: `en-CA`

humphd (approved):
One small nit.  The rest I think we could do in follow-ups, if you file them first.

Great job on this!

dadolhay (approved):
Looks Great!

PR #183, opened 2023-02-12
docs(wiki): ci overview

shawnyu5:
This documentation will eventually land in the wiki, once approved here. It aims to provide an overview of our CI pipeline.

I'll close this PR and move the contents into the wiki once approved here.

### DIFF
--- a/ci.md
+++ b/ci.md
@@ -1,0 +1,53 @@
+# CI pipeline overview
+
+This is an overview of the CI pipeline within starchart
+
+## Concurrency
+
+The `Concurrency` group is used to ensure only one instance of the CI job is running per branch. For example, if 2 commits are pushed in close session to one another, the first commits CI run will be cancelled in favor of the second run.
+
+This is done through defining a concurrency group with a unique identifier: `group: ${{ github.workflow }}-${{ github.ref }}`, in which only 1 CI run can occupy. The newer CI run will always take presence over older runs.
+
+## Jobs
+
+All CI jobs are run concurrently, and utilize caching where possible.
+
+In every job, it will attempt to fetch `node_modules` from a cache, instead of installing it fresh
+
+```yaml
+- name: Restore node_modules cache
+  id: node_modules-cache-restore
+  uses: actions/cache/restore@v3
+  with:
+    # path to node_modules to cache
+    path: ./node_modules
+    # cache module with hash of package-lock.json
+    key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+```
+
+The hash key is the os name followed by the package-lock hash(`Linux-node-63fe6c0ea79bc1812337982f05a442b9a9ee4df56c872a7f4dad8df440079be5`). This is so when app dependencies are updated, and package-lock is regenerated. It will download and cache the newer `node_modules` instead of restoring the older version from cache.
+
+---
+
+The `cache/restore` action exposes a `cache-hit` output, which is set to true when the cache is restored. In the event the cache is not found, `node_modules` must be installed manually in the job:
+
+```yaml
+- name: Install node_modules
+  # install node modules manually only if cache was not restored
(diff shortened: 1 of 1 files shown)


ci.md
```diff
@@ -0,0 +1,55 @@
...
+  with:
+    path:
+      ./node_modules
+      # cache node modules using the same key as restore.
+    key: ${{ steps.node_modules-cache-restore.outputs.cache-primary-key }}
+```
+
+Refer to the Cache action [documentation](https://github.com/marketplace/actions/cache) for more detail regarding caching
+
+> **Note** having the CI run concurrently comes at the cost of some boiler plate code shared across all jobs. Each job must perform caching and restoration of `node_modules` it self.
```
humphd, line 55: "itself"

humphd (changes requested):
One typo, the rest is excellent.  Thanks for documenting this.

This reminds me that we should document how our e2e tests will upload artifacts when there is a failure cc @Eakam1007.

ci.md
shawnyu5: Uhh did we accidentally ping cc?

Eakam1007:
> @Eakam1007 do you want to put the E2E docs in here as well? I think it makes sense as part of the "CI pipeline overview".

Yeah, I think it is fine to put them here as those are part of CI as well

ci.md
```diff
@@ -0,0 +1,53 @@
...
+  id: node_modules-cache-restore
+  uses: actions/cache/restore@v3
+  with:
+    # path to node_modules to cache
+    path: ./node_modules
+    # cache module with hash of package-lock.json
+    key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+```
+
+## The hash key is the os name followed by the package-lock hash(`Linux-node-63fe6c0ea79bc1812337982f05a442b9a9ee4df56c872a7f4dad8df440079be5`). This is so when app dependencies are updated, and package-lock is regenerated. It will download and cache the newer `node_modules` instead of restoring the older version from cache.
```
cychu42, line 28: Nice work!
We might want to use a different way to highlight this text. It's currently the same way as titles.

shawnyu5: oops, idk why that became a subheading. Fixed it

shawnyu5:
If people are happy with this, I'll add this to the wiki for now. We can add the E2E stuff later on.

ci.md
```diff
@@ -0,0 +1,53 @@
...
+  id: node_modules-cache-restore
+  uses: actions/cache/restore@v3
+  with:
+    # path to node_modules to cache
+    path: ./node_modules
+    # cache module with hash of package-lock.json
+    key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+```
+
+The hash key is the os name followed by the package-lock hash(`Linux-node-63fe6c0ea79bc1812337982f05a442b9a9ee4df56c872a7f4dad8df440079be5`). This is so when app dependencies are updated, and package-lock is regenerated. It will download and cache the newer `node_modules` instead of restoring the older version from cache.
```
Eakam1007, line 28:
```suggestion
The hash key is the os name followed by the package-lock hash(`Linux-node-63fe6c0ea79bc1812337982f05a442b9a9ee4df56c872a7f4dad8df440079be5`). This is so when app dependencies are updated and package-lock is regenerated, it will download and cache the newer `node_modules` instead of restoring the older version from cache.
```

This period seems unintentional. Not sure if this is the best fix though

Eakam1007 (changes requested):
One small thing but looks good

Eakam1007 (approved):

SerpentBytes (approved):
Learned something new.

humphd (approved):

SerpentBytes:
@shawnyu5, could you squash and merge this PR?

shawnyu5:
These docs has been added to the wiki

https://github.com/DevelopingSpace/starchart/wiki/CI-pipeline-overview

PR #184, opened 2023-02-12
Add unit tests for Nodemailer

### LINKED ISSUE #147: Implement tests to send test notifications using MailHog
## Description
Issue #126 deals with setting up MailHog to send test notifications to ensure Nodemailer is configured correctly. However, there are no tests implemented to verify whether the Nodemailer module works in the development environment. It would be good idea to implement test our current Nodemailer configuration.

SerpentBytes:
## Description
Resolves #147  by adding initial Unit Tests to test the functionality of Nodemailer. I had to adjust the `sendNotification` function for unit testing. Also, I  have modified the `tests` script to  require `dotenv/config` to read environment variables defined in `.env.example`.

### Steps to Test

1. Setup the project locally (make sure Docker is running)
2. Run `npm test` (all tests should pass)
3. Navigate to the MailHog interface at http://localhost:8025/ to see the test notification:
list view
detailed view

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -79,6 +79,11 @@ jobs:
         image: motoserver/moto:4.1.1
         ports:
           - '5053:5000'
+      mailhog:
+        image: jcalonso/mailhog:latest
+        ports:
+          - 8025:8025
+          - 1025:1025
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "setup": "run-s db:generate db:push db:seed",
     "start": "cross-env NODE_ENV=production node ./build/server.js",
     "start:e2e": "cross-env NODE_ENV=production node --require dotenv/config ./build/server.js",
-    "test": "vitest",
+    "test": "cross-env NODE_OPTIONS=--require=dotenv/config vitest",
     "test:e2e:dev": "cross-env PORT=8080 start-server-and-test dev http://localhost:8080 \"playwright test\"",
     "pretest:e2e:run": "npm run build",
     "test:e2e:run": "cross-env PORT=8811 start-server-and-test start:e2e http://localhost:8811 \"playwright test\"",
--- a/test/unit/notifications.server.test.ts
+++ b/test/unit/notifications.server.test.ts
@@ -1,0 +1,40 @@
+import sendNotification from '../../app/lib/notifications.server';
+
+const { NOTIFICATIONS_EMAIL_USER } = process.env;
+
+describe('Notification server lib function test', () => {
+  const recipient = 'dev@starchart.invalid';
+  const subject = 'Sample subject';
+  const text = 'Sample text';
+
+  test('sending and receiving email notification using sendNotification()', async () => {
+    const sentMailResponse = await sendNotification(recipient, subject, text);
+
+    let messageId = sentMailResponse?.messageId;
+
+    expect(sentMailResponse).toEqual(
+      expect.objectContaining({
+        accepted: expect.arrayContaining([recipient]),
+      })
+    );
+
+    const mailhogResponse = await fetch(
+      `http://localhost:8025/api/v2/search?kind=to&query=${recipient}`
+    );
+    const responseToJSON = await mailhogResponse.json();
+    expect(responseToJSON.items).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          Content: expect.objectContaining({
+            Headers: expect.objectContaining({
+              'Message-ID': expect.arrayContaining([messageId]),
+              'To': expect.arrayContaining([recipient]),
+              'From': expect.arrayContaining([NOTIFICATIONS_EMAIL_USER]),
+            }),
+            Body: expect.stringMatching(text),
+          }),
+        }),
+      ])
+    );
+  });
+});


.github/workflows/ci.yaml
```diff
@@ -61,6 +61,9 @@ jobs:
 
   Unit-Tests:
     runs-on: ubuntu-latest
+    env:
+      NOTIFICATIONS_EMAIL_USER: no-reply@senecacollege.ca
```
Eakam1007, line 65: This might become redundant once https://github.com/Seneca-CDOT/starchart/pull/162 is merged.
These variable are in env.example so they will be loaded because of this change: 
image

We could remove these later after the merge. Or do you want to implement the same change here?

SerpentBytes: Unit Tests are not failing because of adding `env` vars in the CI file for now. I can remove them later.

app/lib/notifications.server.ts
```diff
@@ -34,14 +35,16 @@ const sendNotification = async (emailAddress: string, subject: string, text: str
...
     const transport = initializeTransport();
     logger.debug(`Sending notification to ${emailAddress}`);
-    return await transport.sendMail({
+    await transport.sendMail({
       from: NOTIFICATIONS_EMAIL_USER,
       to: emailAddress,
       subject,
       text,
     });
+    return true;
```
Eakam1007, line 44: Is there any reason a [callback function](https://nodemailer.com/usage/#sending-mail) is not being used for nodemailer to check if there was an error while sending the mail?

SerpentBytes: @Eakam1007, good question. If `sendMail` throws an error, the code in the `catch` block will print the reason for failure.

SerpentBytes: Also, I did not want to add another `logger` statement to log the error message.

Eakam1007: Ah, I did not know that it throws an error. Thanks!

Eakam1007 (dismissed):
Looks good!

app/lib/notifications.server.ts
```diff
@@ -1,6 +1,7 @@
 import { createTransport } from 'nodemailer';
 import { secrets } from 'docker-secret';
 import logger from './logger.server';
+require('dotenv').config();
```
humphd, line 4: The rest of our code is doing this at the level of the script, see https://github.com/Seneca-CDOT/starchart/blob/main/package.json#L23-L24.  Let's do the same for the `test` script.

app/lib/notifications.server.ts
humphd, line 44: I don't think this is right.  If it works, return nothing.  If it fails, throw.

You should return the promise, and in tests you can expect it to resolve or reject.

test/unit/notifications.server.test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  test('sending email notification using sendNotification() ', async () => {
+    const response = await sendNotification(email, subject, text);
+    expect(response).toBe(true);
+  });
+
+  test('verifying MailHog received test email notification via SendNotification() ', async () => {
+    const response = await fetch(`http://localhost:8025/api/v2/search?kind=to&query=${email}`);
+    const responseToJSON = await response.json();
+    const mostRecentMessage = responseToJSON.items[0].Content;
```
humphd, line 16: This is dangerous, since we may add more tests later that mean there are other items in the mailbox.  Don't depend on the state of external systems, order, etc.

A better approach would be to use `expect.arrayContaining()`, which allows you to define a partial array, and make sure that the larger array includes this.

See https://vitest.dev/api/expect.html#expect-arraycontaining and https://vitest.dev/api/expect.html#expect-objectcontaining.

humphd (changes requested):

app/lib/notifications.server.ts
```diff
@@ -31,18 +31,22 @@ const initializeTransport = () => {
...
-    return await transport.sendMail({
-      from: NOTIFICATIONS_EMAIL_USER,
-      to: emailAddress,
-      subject,
-      text,
-    });
-  } catch (error) {
-    logger.warn('Unable to send notification', error);
-  }
+  return new Promise<String>(async (resolve, reject) => {
```
humphd, line 34: No, let's not wrap with additional promises.  An `async` function automatically wraps the `return` in a Promise, and you're adding yet another manual promise here.

We already have a Promise in the form of `transport.sendMail()`, which you can return directly:

```ts
const transport = initializeTransport();
logger.debug(`Sending notification to ${emailAddress}`);
return transport.sendMail(...).catch(err => logger.warn('Unable to send notification', err));
```

Doing this, you don't need `async` on the function either.

app/lib/notifications.server.ts
```diff
@@ -34,12 +34,16 @@ const sendNotification = async (emailAddress: string, subject: string, text: str
...
-      text,
-    });
+    return await transport
+      .sendMail({
+        from: NOTIFICATIONS_EMAIL_USER,
+        to: emailAddress,
+        subject,
+        text,
+      })
+      .catch((error) => {
```
Eakam1007, line 44: Would the catch on line 47 not catch this?

SerpentBytes: If I remove the `try/catch` block and let's say the `NOTIFICATIONS_EMAIL_USER` is not defined, it throws an error instead of a warning. Could you test it out and let me know?

SerpentBytes: I am unsure if we should throw an error.

Eakam1007: I think you can remove the await from line 37 and simply return the promise as @humphd suggested above

app/lib/notifications.server.ts
```diff
@@ -34,7 +34,7 @@ const sendNotification = async (emailAddress: string, subject: string, text: str
   try {
```
Eakam1007, line 34: You can remove the `async` from this function since you are no longer using await

app/lib/notifications.server.ts
```diff
@@ -34,7 +34,7 @@ const sendNotification = async (emailAddress: string, subject: string, text: str
...
     const transport = initializeTransport();
     logger.debug(`Sending notification to ${emailAddress}`);
-    return await transport.sendMail({
+    return transport.sendMail({
```
Eakam1007, line 37: Since the await was removed, I am pretty sure you need a `.catch` now

test/unit/notifications.server.test.ts
```diff
@@ -0,0 +1,33 @@
+import sendNotification from '../../app/lib/notifications.server';
+const { NOTIFICATIONS_EMAIL_USER } = process.env;
+describe('Notification server lib function test', () => {
+  const recipient: string = 'dev@starchart.invalid';
```
humphd, line 4: tip: in TS, you only need to provide a type when it can't be inferred from the context.  I this case, it's clear that you're assigning a `string` so you can omit it.  It might seem odd to leave things out, but it's generally the way it's done.

test/unit/notifications.server.test.ts
```diff
@@ -0,0 +1,33 @@
...
+      expect.objectContaining({
+        accepted: expect.arrayContaining([recipient]),
+      })
+    );
+  });
+
+  test('verifying MailHog received test email notification via SendNotification() ', async () => {
+    const response = await fetch(`http://localhost:8025/api/v2/search?kind=to&query=${recipient}`);
+    const responseToJSON = await response.json();
+    const mostRecentMessage = responseToJSON.items[0];
```
humphd, line 20: This is still problematic.  We should test against `.items` vs. assuming anything about the order or position in the array.  Also, this test now depends on the previous test being run before it, which you can't assume.  There should be no order assumed between test runs.

humphd (changes requested):

SerpentBytes:
I have eliminated test dependency by combining both tests into one. I am working on addressing testing against the `.items` vs `.items[0]`.

test/unit/notifications.server.test.ts
```diff
@@ -0,0 +1,32 @@
...
+        accepted: expect.arrayContaining([recipient]),
+      })
+    );
+    const mailhogResponse = await fetch(
+      `http://localhost:8025/api/v2/search?kind=to&query=${recipient}`
+    );
+    const responseToJSON = await mailhogResponse.json();
+    const mostRecentMessage = responseToJSON.items[0];
+    expect(mostRecentMessage).toEqual(
+      expect.objectContaining({
```
humphd, line 21:
```ts
expect.arrayContaining([
  expect.objectContaining({
    ...
```

humphd (changes requested):

test/unit/notifications.server.test.ts
```diff
@@ -0,0 +1,44 @@
...
+        accepted: expect.arrayContaining([recipient]),
+      })
+    );
+
+    const mailhogResponse = await fetch(
+      `http://localhost:8025/api/v2/search?kind=to&query=${recipient}`
+    );
+    const responseToJSON = await mailhogResponse.json();
+    const mostRecentMessage = responseToJSON.items;
+    expect(mostRecentMessage).toEqual(
```
Eakam1007, line 29: I think you can remove this variable entirely

Eakam1007:
```suggestion
    expect(responseToJSON.items).toEqual(
```

Eakam1007 (changes requested):

test/unit/notifications.server.test.ts
SerpentBytes: Good catch. Thanks.

test/unit/notifications.server.test.ts
```diff
@@ -0,0 +1,43 @@
...
+describe('Notification server lib function test', () => {
+  const recipient = 'dev@starchart.invalid';
+  const subject = 'Sample subject';
+  const text = 'Sample text';
+
+  test('sending and receiving email notification using sendNotification()', async () => {
+    const sentMailResponse = await sendNotification(recipient, subject, text);
+
+    // prevent typescript error: messageId is possibly undefined
+    invariant(sentMailResponse?.messageId, 'Expected sentMailResponse.messageId to be defined');
```
Eakam1007, line 15: I am confused, how does this prevent the typescript error. From what I understand, invariant is something akin to a logger?

SerpentBytes: We were getting type-checking tests failing earlier, saying "messageId is possibly undefined," so @humphd and I discussed the issue on Slack and using the `invariant` function from `tiny-variant` throws if the value is not truthy with an error message, rendering the whole test to fail.

shawnyu5: yeah Im kinda confused about this as well. Why using `invariant` here works, but rather `expect`ing  `sentMailResponse.messageId` to be not null doesnt work. As far is getting rid of the ts error.

Eakam1007: Tested this locally, you could fix this by changing line 16 to `sentMailResponse?.messageId` instead. Then you can remove this line

shawnyu5: wouldn't `expect` do something similar?

package.json
```diff
@@ -21,7 +21,7 @@
     "setup": "run-s db:generate db:push db:seed",
     "start": "cross-env NODE_ENV=production node ./build/server.js",
     "start:e2e": "cross-env NODE_ENV=production node --require dotenv/config ./build/server.js",
-    "test": "vitest",
+    "test": "NODE_OPTIONS=--require=dotenv/config vitest",
```
Eakam1007, line 24:
```suggestion
    "test": "cross-env NODE_OPTIONS=--require=dotenv/config vitest",
```
Without this, `npm test` will not run on windows

test/unit/notifications.server.test.ts
SerpentBytes: > yeah Im kinda confused about this as well. Why using `invariant` here works, but rather `expect`ing `sentMailResponse.messageId` to be not null doesnt work. As far is getting rid of the ts error.

I tried expecting `messageId` to be `typeof` string but that doesn't get rid of the error unless you put the line that initializes messageId in an if statement that checks whether `messageId` exists. I could have done that, but @humphd  recommended using `tiny-invariant` which would throw if `messageId` is undefined, rendering the test to fail.

Eakam1007: Was this the error you were getting? 
image

shawnyu5: > Tested this locally, I think you could fix this by changing line 16 to sentMailResponse?.messageId instead. Then you can remove this line

doing this would fix it I think

shawnyu5:
Please stop force pushing....

test/unit/notifications.server.test.ts
Eakam1007: Yeah, I am pretty sure it does. I ran typecheck after this and didn't see any errors

Eakam1007:
> Please stop force pushing....

I think we don't need to squash and merge locally in general. It makes it hard to re-review changes. 
When we merge PRs, we can make sure to `Squash and Merge` instead and keep the history on the PR

shawnyu5:
Exactly

test/unit/notifications.server.test.ts
Eakam1007: Could you try this @SerpentBytes ?

> Tested this locally, I think you could fix this by changing line 16 to `sentMailResponse?.messageId` instead. Then you can remove this line

Eakam1007 (changes requested):

test/unit/notifications.server.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    );
+    const responseToJSON = await mailhogResponse.json();
+    expect(responseToJSON.items).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          Content: expect.objectContaining({
+            Headers: expect.objectContaining({
+              'Message-ID': expect.arrayContaining([messageId]),
+              'To': expect.arrayContaining([recipient]),
+              'From': expect.arrayContaining([NOTIFICATIONS_EMAIL_USER]),
```
shawnyu5, line 32: Not sure if I like how deeply nested this is. Or understand whats going on here....

Eakam1007 (dismissed):
Looks good!

test/unit/notifications.server.test.ts
Eakam1007: I think this is testing if the response items array contains the object defined here

Eakam1007: Maybe this could be moved into a variable for better readability?

shawnyu5: I think it might be nicer to split up the obj, something like this:

```ts
expect(responseToJSON.items.Content).ToContain(...)
expect(responseToJSON.items.Content.Headers).ToContain(...)
```
idk if this is fully possible

humphd: No, this is pretty standard.  Adding variables adds places where this can fail due to variability.  What's here is looking good to me.

shawnyu5: Testing each of the objects fields one at a time. Rather than all at once

humphd (dismissed):

test/unit/notifications.server.test.ts
shawnyu5: fair enough

Genne23v (changes requested):

test/unit/notifications.server.test.ts
```diff
@@ -0,0 +1,39 @@
+import sendNotification from '../../app/lib/notifications.server';
+const { NOTIFICATIONS_EMAIL_USER } = process.env;
+describe('Notification server lib function test', () => {
+  let messageId: string;
+  const recipient = 'dev@starchart.invalid';
+  const subject = 'Sample subject';
+  const text = 'Sample text';
+
+  test('sending and receiving email notification using sendNotification()', async () => {
+    const sentMailResponse = await sendNotification(recipient, subject, text);
```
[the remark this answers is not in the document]
SerpentBytes: Yes, even if the recipient is `xyz@example.com` the email would still get delivered using MailHog. We don't need real email addresses for testing.

test/unit/notifications.server.test.ts
```diff
@@ -0,0 +1,33 @@
+import sendNotification from '../../app/lib/notifications.server';
+const { NOTIFICATIONS_EMAIL_USER } = process.env;
+describe('Notification server lib function test', () => {
+  const recipient: string = 'dev@starchart.invalid';
+  const subject: string = 'Sample subject';
+  const text: string = 'Sample text';
+
+  test('sending email notification using sendNotification() ', async () => {
```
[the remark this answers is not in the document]
SerpentBytes: This appears to be an old commit.

test/unit/notifications.server.test.ts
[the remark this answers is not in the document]
Genne23v: I see. I was thinking we might need a fail case testing.

Genne23v (approved):

test/unit/notifications.server.test.ts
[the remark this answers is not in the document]
SerpentBytes: In production, we would expect it to fail. Here, we are just testing if are able to send/receive notifications.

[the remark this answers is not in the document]
shawnyu5: Uhhh there is a distinction between production testing and dev testing?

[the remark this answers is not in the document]
SerpentBytes: > I see. I was thinking we might need a fail case testing.

I could add that in a follow-up.

Eakam1007 (approved):

PR #201, opened 2023-02-13
Change schema columns to follow naming convention

### LINKED ISSUE #153: Fix/change the user schema's column names for records and certificate

cychu42:
Closes #153

Changes:
- Change `subdomians` to `record`
- Change `certificates` to be singular
- Change `privkey` to `privateKey`


cychu42:
Huh. Local type check doesn't produce the same error, and the project runs with the seeded database.

humphd:
> Huh. Local type check doesn't produce the same error, and the project runs with the seeded database.

Probably we need to generate the prisma client/types before we run typecheck.  cc @shawnyu5:

https://github.com/DevelopingSpace/starchart/blob/main/.github/workflows/ci.yaml#L258-L293

Genne23v (approved):
It looks good to me

shawnyu5:
> Probably we need to generate the prisma client/types before we run typecheck. cc @shawnyu5:

Tested with a seeded db. Type check run now passes

https://github.com/DevelopingSpace/starchart/actions/runs/4170194203/jobs/7218918845

cychu42:
I reran the newer CI, and now it has no error.

Genne23v (approved):

SerpentBytes (approved):

PR #208, opened 2023-02-14
fix(ci): generate prisma types before type-check job

shawnyu5:
Type-check requires prisma types to be generated to work.

https://github.com/DevelopingSpace/starchart/pull/201#issuecomment-1428876224

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -281,7 +281,7 @@ jobs:
         if: steps.node_modules-cache-restore.outputs.cache-hit != 'true'
         run: npm ci
 
-      - name: Seed database
+      - name: Generate prisma client types
         run: npm run db:generate
 
       - name: Run typecheck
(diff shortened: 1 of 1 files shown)


shawnyu5:
This change works as expected and fixes the issue

https://github.com/DevelopingSpace/starchart/pull/201#issuecomment-1429062279

.github/workflows/ci.yaml
```diff
@@ -281,6 +281,9 @@ jobs:
         if: steps.node_modules-cache-restore.outputs.cache-hit != 'true'
         run: npm ci
 
+      - name: Seed database
```
humphd, line 284: This won't seed the database.  It generates the Prisma client and types.  If this was copied from elsewhere in this file, it should get updated there too for clarity.

humphd (changes requested):

humphd (approved):

cychu42 (approved):

Issue #202, opened 2023-02-14
Header is not responsive


humphd:
The header isn't working on smaller viewport sizes.  We should get it working on any screen size.

<img width="391" alt="Screenshot 2023-02-13 at 8 17 17 PM" src="https://user-images.githubusercontent.com/427398/218613185-893283c8-abbb-467a-a1cf-aaae6d7f5d67.png">

Ririio:
I didn't considered this during the development process. Can I be assigned for this to fix the issue?

humphd:
@Ririio you never need to ask, go ahead and fix any/all bugs we have.

Genne23v:
Closed with #212

PR #212, opened 2023-02-14
Making header responsive

Ririio:
This is a pr in reference to #202 

I added a video demo to demonstrate what it does for different resolution

https://user-images.githubusercontent.com/32577022/218791041-a612f7c3-a54b-48bf-8212-e31a3acaee98.mp4

I added a 'hamburger' state to accommodate for having two links. I was thinking on whether I should just  turn it into an icon, but decided that it's for the best to have the name also be shown to avoid confusion.

I decided to remove the user name at a lower res, and was thinking on whether I should add it as a `MenuItem` or not...

### DIFF
--- a/app/components/header.tsx
+++ b/app/components/header.tsx
@@ -9,8 +9,11 @@ import {
   MenuList,
   MenuItem,
   IconButton,
+  Show,
+  Hide,
+  Button,
 } from '@chakra-ui/react';
-import { TriangleUpIcon, LockIcon } from '@chakra-ui/icons';
+import { TriangleUpIcon, LockIcon, HamburgerIcon } from '@chakra-ui/icons';
 import { Link, useFetcher } from '@remix-run/react';
 
 import { useUser } from '~/utils';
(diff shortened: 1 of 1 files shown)


humphd:
I really like what you've done. There is still some bug happening at the 0:28 mark of your video, though.  One of the breakpoints is wrong.

Screenshot 2023-02-14 at 11 15 44 AM

Ririio:
> I really like what you've done. There is still some bug happening at the 0:28 mark of your video, though. One of the breakpoints is wrong.

oh no I noticed that too, but it's not from the code. Sometimes that can happen when I change resolution on the dev tool. If I try it again with the same format it works as intended

https://user-images.githubusercontent.com/32577022/218796618-1585a584-9bc0-4f53-9faa-47329b08360f.mp4

humphd (approved):

Genne23v (approved):
It looks header is condensed in `iPad` view. I think we can follow up when other UI are get ready. It looks really nice.

Issue #97, opened 2023-02-02
Create models for all tables in MySQL


humphd:
Similar to what's happening in #96, we need to create files in `app/models/*` for all of our data models:

- user
- record
- certificate
- etc.

These should hide the details of working with Prisma, similar to how the user model works https://github.com/Seneca-CDOT/starchart/blob/c86bbe336d8ca9a75a5157bb326b9827c9bd7756/app/models/user.server.ts

The specific API to use needs some thought.  We'll also need some tests.

cychu42:
I thought this issue is just for making functions back then.
I'm not sure I'll get to tests this week, so I'm dissociating this from milestone 0.3 for now.

humphd:
I think we should close this now, and continue work in other issues.  @cychu42 can you do that, unless you disagree for some reason?

cychu42:
Yes, we can close this to mark CRUD as "done". Testing is mentioned in the other issue mentioned.

PR #228, opened 2023-02-15
Add CRUD functions for models via Prisma

cychu42:
Relates to #97

## Changes:
- This PR adds CRUD functions for models via Prisma, for model: User, Record, Certificate, Challenge.
- `deleteUserByEmail()` changes to `deleteUserByUsername()`

## Test:
1. You can try to use the functions to see if they function correctly as CRUD operations. 
- Option 1: You can use them in your code.
- Option 2: You can [make a query file](https://github.com/DevelopingSpace/starchart/wiki/Prisma#3-write-your-first-query) to import and execute these functions.
NOTE: `npx ts-node` needs to have `--require tsconfig-paths/register` or it won't understand `~` for paths( to `db.bserver` and `logger.server`) in our project.

2. `npm run db:studio` can let you view data in database after an operation.

### DIFF
--- a/app/models/certificate.server.ts
+++ b/app/models/certificate.server.ts
@@ -1,0 +1,64 @@
+import type { Certificate } from '@prisma/client';
+
+import { prisma } from '~/db.server';
+
+export type { Certificate } from '@prisma/client';
+
+export async function getCertificateByUsername(username: Certificate['username']) {
+  return prisma.certificate.findUnique({ where: { username } });
+}
+
+export async function createCertificate(
+  username: Certificate['username'],
+  subject: Certificate['subject'],
+  certificate: Certificate['certificate'],
+  orderUrl: Certificate['orderUrl'],
+  privateKey: Certificate['privateKey']
+) {
+  // Set expiration date 90 days from now
+  const validTo = new Date();
+  validTo.setDate(validTo.getDate() + 90);
+
+  return prisma.certificate.create({
+    data: {
+      username,
+      subject,
+      certificate,
+      orderUrl,
+      privateKey,
+      validTo,
+    },
+  });
+}
+
+export async function updateCertificateByUsername(
+  username: Certificate['username'],
+  subject?: Certificate['subject'],
+  certificate?: Certificate['certificate'],
+  orderUrl?: Certificate['orderUrl'],
+  privateKey?: Certificate['privateKey'],
+  validFrom?: Certificate['validFrom']
+) {
+  // If validFrom is changed, set validTo to 90 days from validFrom
+  let validTo;
+  if (validFrom) {
+    validTo = validFrom;
+    validTo.setDate(validTo.getDate() + 90);
+  }
+  return prisma.certificate.update({
+    where: { username },
+    data: {
+      username,
+      subject,
+      certificate,
+      orderUrl,
+      privateKey,
+      validFrom,
+      validTo,
+    },
+  });
+}
+
+export async function deleteCertificateByUsername(username: Certificate['username']) {
+  return prisma.certificate.delete({ where: { username } });
+}
--- a/app/models/challenge.server.ts
+++ b/app/models/challenge.server.ts
@@ -1,0 +1,43 @@
+import type { Challenge } from '@prisma/client';
+
+import { prisma } from '~/db.server';
+
+export type { Challenge } from '@prisma/client';
+
+export async function getChallengeByCertificateId(certificateId: Challenge['certificateId']) {
+  return prisma.challenge.findMany({ where: { certificateId } });
+}
+
+export async function createChallenge(
+  domain: Challenge['domain'],
+  challengeKey: Challenge['challengeKey'],
+  certificateId: Challenge['certificateId']
+) {
+  return prisma.challenge.create({
+    data: {
+      domain,
+      challengeKey,
+      certificateId,
+    },
+  });
+}
+
+export async function updateChallengeById(
+  id: Challenge['id'],
+  certificateId?: Challenge['certificateId'],
+  domain?: Challenge['domain'],
+  challengeKey?: Challenge['challengeKey']
+) {
+  return prisma.challenge.update({
+    where: { id },
+    data: {
+      domain,
+      challengeKey,
+      certificateId,
+    },
+  });
+}
+
+export async function deleteChallengeById(id: Challenge['id']) {
+  return prisma.challenge.delete({ where: { id } });
+}
--- a/app/models/record.server.ts
+++ b/app/models/record.server.ts
@@ -1,0 +1,70 @@
+import type { Record } from '@prisma/client';
+
+import { prisma } from '~/db.server';
+
+export type { Record } from '@prisma/client';
+
+export async function getRecordsByUsername(username: Record['username']) {
+  return prisma.record.findMany({ where: { username } });
+}
+
+export async function createRecord(
+  username: Record['username'],
+  name: Record['name'],
+  type: Record['type'],
+  value: Record['value'],
+  status: Record['status'],
+  description?: Record['description'],
+  course?: Record['course'],
+  ports?: Record['ports']
+) {
+  // Set expiration date 6 months from now
+  const expiresAt = new Date();
+  expiresAt.setMonth(expiresAt.getMonth() + 6);
+
+  return prisma.record.create({
+    data: {
+      username,
+      name,
+      type,
+      value,
+      description,
+      course,
+      ports,
+      expiresAt,
+      status,
+    },
+  });
+}
+
+export async function updateRecordById(
+  id: Record['id'],
+  username?: Record['username'],
+  name?: Record['name'],
+  type?: Record['type'],
+  value?: Record['value'],
+  description?: Record['description'],
+  course?: Record['course'],
+  ports?: Record['ports'],
+  expiresAt?: Record['expiresAt'],
+  status?: Record['status']
+) {
+  return prisma.record.update({
+    where: { id },
+    data: {
+      username,
+      name,
+      type,
+      value,
+      description,
+      course,
+      ports,
+      expiresAt,
+      status,
+    },
+  });
+}
+
+export async function deleteRecordById(id: Record['id']) {
+  return prisma.record.delete({ where: { id } });
+}
--- a/app/models/user.server.ts
+++ b/app/models/user.server.ts
@@ -24,6 +24,23 @@ export async function createUser(
   });
 }
 
-export async function deleteUserByEmail(email: User['email']) {
-  return prisma.user.delete({ where: { email } });
+export async function updateUserByUsername(
+  username: User['username'],
+  firstName?: User['firstName'],
+  lastName?: User['lastName'],
+  email?: User['email']
+) {
+  return prisma.user.update({
+    where: { username },
+    data: {
+      username,
+      firstName,
+      lastName,
+      email,
+    },
+  });
+}
+
+export async function deleteUserByUsername(username: User['username']) {
+  return prisma.user.delete({ where: { username } });
 }


app/models/certificate.server.ts
```diff
@@ -0,0 +1,64 @@
...
+export async function createCertificate(
+  username: Certificate['username'],
+  subject: Certificate['subject'],
+  certificate: Certificate['certificate'],
+  orderUrl: Certificate['orderUrl'],
+  privateKey: Certificate['privateKey']
+) {
+  // Set expiration date 90 days from now
+  let validTo = new Date();
+  validTo.setDate(validTo.getDate() + 90);
```
Ririio, line 20: Is there an issue if you assigned `validTo` with `new Date().getDate() + 90)`?

app/models/certificate.server.ts
```diff
@@ -0,0 +1,64 @@
...
+      username,
+      subject,
+      certificate,
+      orderUrl,
+      privateKey,
+      validFrom,
+      validTo,
+    },
+  });
+}
```
Ririio, line 60: From what I recall with `update`, you always need to pass the original values right?

app/models/certificate.server.ts
```diff
@@ -0,0 +1,64 @@
+import type { Certificate } from '@prisma/client';
+
+import { prisma } from '~/db.server';
+
+export type { Certificate } from '@prisma/client';
+
+export async function getCertificateByUsername(username: Certificate['username']) {
```
Myrfion, line 7: Isn't it better to delete the certificate by `id` rather than a username, since as far as I understand that is the `PK` for it?

app/models/certificate.server.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export async function createCertificate(
+  username: Certificate['username'],
+  subject: Certificate['subject'],
+  certificate: Certificate['certificate'],
+  orderUrl: Certificate['orderUrl'],
+  privateKey: Certificate['privateKey']
+) {
+  // Set expiration date 90 days from now
+  let validTo = new Date();
```
Myrfion, line 19: Minor thing, but I think it is better to use `let` only in case if the variable is mutable, and `const` in all other cases

app/models/certificate.server.ts
Myrfion, line 20: @dadolhay should we consider using some date library instead of native dates? We could use `date-fns` or `moment` or whatever you like (I am pro `date-fns`)

app/models/certificate.server.ts
cychu42: My thinking is that `username` is more accessible than finding out the id/index of a certificate. We only record one (current) certificate per user, so it should be fine.

app/models/certificate.server.ts
cychu42: If the value is `undefined`, Prisma will know to leave that alone. Not `null`, by the way.

app/models/certificate.server.ts
Ririio: `username` acts kind of an id, because it was set that students won't have the same `email address`, so it should work, but I do understand that ID would be a standard and good practice to do

app/models/certificate.server.ts
Ririio: should we refrain from updating the `unchanged` values?

app/models/certificate.server.ts
cychu42: When we do that and leave `validTo` undefined, the issue is that typescript will complain about `validTo` possibly being undefined. Therefore, we would still end up with 2 lines (one to declare, and another to set date), and I still need to add code to solve the complain on top of that.

app/models/certificate.server.ts
cychu42: Good question!

app/models/certificate.server.ts
Ririio: it wouldn't be undefined no? You're assigning it a values from the start `let validTo = new Date().getDate() + 90`

app/models/certificate.server.ts
cychu42: Oh, good catch. I forgot that it's an object so you can do that.

app/models/certificate.server.ts
cychu42: @Ririio Well, it will just effectively do nothing, so it shouldn't cause problems either way.

Ririio: of course lmfao, I was just wondering

app/models/certificate.server.ts
cychu42: A database table always needs something as an unique identifier (primary key) for the row, which can often be an index number like `id`.
You can argue that we should keep `username` as the primary keep instead of using id, which we can change in a different issue, because that will require change in schema.

app/models/certificate.server.ts
cychu42: Oh, I misunderstood you. In that case, it would complain about `Type 'number' is not assignable to type 'string | Date'`.

Ririio: oh gotcha

cychu42:
I tried to rebase my topic branch because there was some issue with history.

Eakam1007:
> I tried to rebase my topic branch because there was some issue with history not matching, when I attempted to make changes from feedback.

I think a bunch of other (not your) changes are showing up after the rebase. You could try rebasing with main

cychu42:
Yeah, I accidently synced the wrong branch to upstream's main, which caused the history mismatch.
Then I tried to rebase without realizing that, and it caused the messy rebase.
I reverted the history and redid it properly with force push, so it's fine now.

app/models/certificate.server.ts
Myrfion: > A database table always needs something as an unique identifier (primary key) for the row, which can often be an index number like `id`. You can argue that we should keep `username` as the primary key instead of using `id`, which we can change in a different issue, because that will require change in schema.

My point is that it is best to use the primary key as an id for certain entities in a table, and doing otherwise might lead to confusion among developers and feels like an antipattern to me. 

Again, I am not a big DB guy so my knowledge is limited, I am curious what @humphd @dadolhay might think on it

Ririio: I think it's best to use index as to username. Not just for helping developers out in the future, but also for having a database that expands overtime

Eakam1007: I don't think it is confusing to say that you want to get a certificate for a user

SerpentBytes: It's okay to use a username if it's always going to stay unique. Since we are building for Seneca, I guess it will always stay unique.

Eakam1007: Yea, I think it will. Also, we can always change this if someday we want to have multiple certificates for a user

Ririio: There is a difference in speed if you use an integer (id) as to a varchar (username); it has something to do with the space they take. Not sure if our database is going to be big enough that the difference will be apparent, but I do recommend it just so we don't have to worry about it later down the line

cychu42: If you are to use the function, which way is more useful to you?

Eakam1007: Username makes more sense to me

Ririio: Actually now that I see it... I don't see an `id` for user, only a `username` assigned as a primary key. I'll just go with username, because I don't want to burden you with changing the schema to include an `id` just for the extra speed

app/models/certificate.server.ts
dadolhay: I have not used `date-fns` before, but there is a library called `luxon`. It's from the same folks who created `moment`, and learning from the mistakes they created `luxon`. Just putting it on your radar, not saying that we have to use that or anything.

`moment` is a huge, heavy weight library, I'd not use it if I didn't have to.
To be honest, I'd not pull in a library until we have to, or it is clear that in the future we will need to do so. (so we can be consistent later)

app/models/certificate.server.ts
cychu42: I think the point about speed is applicable to any PK(primary key). For example, ASP.NET has a function that finds an item by PK, which is faster than other methods. There is a choice to make` username `the PK in `Certificate` model.

In case there's a confusion, each table has their own PK, so what goes on in `User` model/table has no real bearing on what's the PK in other models/tables.

Ririio: I don't know if it changes the speed when reading through a string vs an int. Anyways, there shouldn't be an issue in terms of functionality

cychu42: Yeah, I meant making `username` a PK so this function potentially runs faster as it is.

Eakam1007: I think that should be tackled in a different issue

Ririio (approved):
I only have a nitpick which is the use of `id`, but otherwise the pr looks great!

cychu42:
I hear you. I still think it's more useful to use `username`, and we can talk about potentially making it the PK maybe in a follow up issue.

app/models/certificate.server.ts
```diff
@@ -0,0 +1,64 @@
...
+  orderUrl: Certificate['orderUrl'],
+  privateKey: Certificate['privateKey']
+) {
+  // Set expiration date 90 days from now
+  const validTo = new Date();
+  validTo.setDate(validTo.getDate() + 90);
+
+  return prisma.certificate.create({
+    data: {
+      username,
```
Eakam1007, line 24: Instead of specifying the username, I think you should connect the existing user using `connect`: https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#connect-an-existing-record

I tested your create function, and got this error in the db studio:
```
Inconsistent query result: Field user is required to return data, got `null` instead.
  
Query:
{
  "modelName": "Certificate",
  "operation": "findMany",
  "args": {
    "take": 100,
    "skip": 0,
    "select": {
      "id": true,
      "username": true,
      "subject": true,
      "certificate": true,
      "orderUrl": true,
      "privateKey": true,
      "validFrom": true,
      "validTo": true,
      "user": true,
      "challenge": {
        "select": {
          "id": true
        }
      }
    }
  }
}
```

Eakam1007 (changes requested):
The comment for `createCertificate` might also apply on `createChallenge` and `createRecord` though I have only tested it for `createCertificate`

app/models/certificate.server.ts
Eakam1007:
```suggestion
      user: { connect: { username } }
```

I tested with this, and that resolves the error

cychu42: Thanks for showing me that. That approach can work, but I would need to make another API call to update `user`, so I think the current method is simpler.

As for the error, I didn't run into any error message.
Can you give me more details on this? What did you do to have the query executing? Where is the error displayed? 
I didn't run into any error messages.

cychu42: From the error, it looks like you were trying to do something with `findMany()` on `Certificate` model. It's not a function I made or used in this PR for 'Certificate'. It's just a default function from Prisma itself.

Eakam1007: The findMany() is used by the db studio to get all the certificates. Try creating a certificate, and then open db studio in the browser to see this error. I will update with a screenshot shortly

Eakam1007: Hmmm can't replicate the error anymore. Not sure why, but I guess there is no need to worry about this. Previously, it seemed like the certificate was not connecting to the user. I got the error in the browser

Eakam1007: @cychu42  Could you please confirm that this does not happen for you:
- Use a query file to create a new user, and then a new certificate for the user
- Run `npm run db:studio` and check if the error appears in your browser

Eakam1007 (approved):
I tested all these methods locally; looks great!

Can you file an issue to add test for this code?

app/models/certificate.server.ts
cychu42: I received no error.
Maybe there was mismatch in your previous input for username, or an user got incorrectly removed? Just guesses.

cychu42:
Sure. 
Thank you all for the review.
@humphd Can you help with the ESLint issue?

app/models/certificate.server.ts
Eakam1007: Yeah, I tried removing a user but you can't remove a user without removing the certificate. I am thinking maybe my local environment was messed up

humphd:
Done

PR #231, opened 2023-02-15
Build Docker image for linux/amd64, linux/arm64, and linux/arm64/v8

humphd:
This updates CI so that it builds the Docker image for multiple architectures (we only do linux/amd64 right now).  As usual, this is hard to test without actually doing it, so I'll fix bugs after landing.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -321,6 +321,9 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
 
+      - name: Set up QEMU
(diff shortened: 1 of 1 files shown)


Genne23v (approved):
I don't see any spelling error. It looks good to me.

humphd:
I've added linux/arm64/v8 to the list, since I need that locally for my M1 mac to test the staging deploy.

dadolhay (approved):
Looks good!

Genne23v (approved):
Looks good to me

Issue #213, opened 2023-02-14
Implementation of UI for Certificate


Ririio:
Creating the certificate page based off the figma template. 

There's no way to access the database at the moment, thus I decided to create a temporary `Certificate` data to be used for the route.

PR #232, opened 2023-02-16
Implement UI for certificate

Ririio:
This pr is based on issue #213 

Here's a live demo of the UI:

[Live Demo](https://user-images.githubusercontent.com/32577022/219228645-28b96fd2-3baa-4e5a-a2fa-f25df7fcca36.mp4)

- I decided to use an [Accordion](https://chakra-ui.com/docs/components/accordion) instead of a `textArea`, I find it a lot more cleaner
- I used `useState` to display different components once the user has pressed the `Request a Certificate` button
- Instead of having a large download button for each key, I decided to switch it to a button that has hover effect, tooltip and gap, to ensure that user don't accidentally press the button
- I'm unable to access the `seed` certificate, so I'm using a temporary object to produce the necessary output


dadolhay (changes requested):
I think the loading progress is not the correct UX here.

A certificate request will probably take at the very minimum 1-2 minutes, users will not wait that long for that animation (it looks cool for sure, but still)

Also, it feels like it tries to keep the user glued to the screen, but the process is unfortunately not that fast for this to be feasible.

This is why we talked about showing some kind of pending state in the same structure we are showing the certificate details in the second screen in your video

dadolhay:
On the other hand, the cert details screen looks very nice :)

Ririio:
> I think the loading progress is not the correct UX here.
> 
> A certificate request will probably take at the very minimum 1-2 minutes, users will not wait that long for that animation (it looks cool for sure, but still)
> 
> Also, it feels like it tries to keep the user glued to the screen, but the process is unfortunately not that fast for this to be feasible.
> 
> This is why we talked about showing some kind of pending state in the same structure we are showing the certificate details in the second screen in your video

Yeah... I'm struggling a bit with understanding what a good UX is for the loading page. From what I can understand, you want to head directly to the 2nd page, but provide some sort of a "loading" buffer before displaying the details? Like a spinning wheel of some kind?

dadolhay:
Yes, with some kind of description, that we are creating the certificate, but this is an asyncrhronous process, requiring multiple 3rd party providers (route53 and let's encrypt), so for this reason it will take a few minutes

humphd:
I really love this @Ririio, great job.  A few comments:

- For the "pending" screen, I think it should be static (i.e., no animation), which will help with it taking a while and not seeming broken.  We could include an image from https://undraw.co/illustrations coloured red for example.  Not this one, but something like this with text saying, "We've ordered your certificate and will notify you when it's ready":

Screenshot 2023-02-16 at 1 19 22 PM
 
- We said that the main part of the UI should be centred vs. left-aligned.
-

Ririio:
> For the "pending" screen, I think it should be static (i.e., no animation), which will help with it taking a while and not seeming broken. We could include an image from https://undraw.co/illustrations coloured red for example. Not this one, but something like this with text saying, "We've ordered your certificate and will notify you when it's ready":

We're allowed to use this website for illustration? I

> We said that the main part of the UI should be centred vs. left-aligned

I recall there was a discussion for the ui to be left-aligned-- I had to change the layout on the figma template

humphd:
> We're allowed to use this website for illustration?

Yes, it is [open licensed](https://undraw.co/license)

> I recall there was a discussion for the ui to be left-aligned-- I had to change the layout on the figma template

What I heard in that meeting was "left-aligned in a container that is centred," but if I'm wrong, do what you have.

cc @Myrfion, @Eakam1007

Eakam1007:
> What I heard in that meeting was "left-aligned in a container that is centered," but if I'm wrong, do what you have.
Yeah, I believe it was discussed that the form should not touch the left edge but should be a bit more towards the center (so basically a margin on the left)

I think this suggestion should make it work

Ririio:
> Yeah, I believe it was discussed that the form should not touch the left edge but should be a bit more towards the center (so basically a margin on the left)

Just to be clear, I just have to move the container more to the center, and have its content left align?

Eakam1007:
> Just to be clear, I just have to move the container more to the center, and have its content left align?

I would say it is up to you how you decide to do it. But yeah, I think that should work

humphd:
What I'm more looking for is that this be responsive.  When we're on a really wide screen, it's nicer to put it in the middle.  On a narrower screen, it's going to be more on the left, etc.

We can also do that stuff in a follow up.  But putting the "main" part of this in a container that is in the center, and the contents are left-aligned within it, seems good to me.

If it doesn't work, don't do it.  Just a thought.

Ririio:
@humphd @dadolhay I changed the loading screen, and aligned the containers in the middle as per your suggestion. If you don't mind taking a look here's the video...

https://user-images.githubusercontent.com/32577022/219515489-b32ebaff-23ed-49e1-aaca-a2efb32fc8cb.mp4

humphd (commented):
I really like where this is going.

Reading the code, I wonder if we should be splitting this up into a few larger components vs. having such deep nesting on state.  That could happen in a follow-up, so we don't block this for 0.3, but I think this file is getting pretty big, and we could separate the three states into smaller pieces of code that would be easier to maintain.

I'm tagging some other reviewers who can help with thinking about the front-end stuff.

Ririio:
> Reading the code, I wonder if we should be splitting this up into a few larger components vs. having such deep nesting on state. That could happen in a follow-up, so we don't block this for 0.3, but I think this file is getting pretty big, and we could separate the three states into smaller pieces of code that would be easier to maintain.

If we're splitting them up, can I create a directory called `components` underneath the `certificate` to split the code? I'm not familiar with the standard practice of remix in terms of creating components, so I refrained from doing it. Am I allowed to do something like this?

image

Eakam1007:
I believe there already is a components directory under app. I think you could use that?

Ririio:
> I believe there already is a components directory under app. I think you could use that?

sure

Ririio:
@Eakam1007 is this what you meant?

humphd (changes requested):
`certificate-display.tsx` does the exact same UI twice.  You could break that up into re-usable pieces too, which will halve this code, and pass the things that are different (e.g., the title, the value in the accordion view) as props.

Because you have a bunch of "certificate" components (e.g., `loading` is pretty vague`, I'd put them all under `app/components/certificate/*` so they are logically grouped together.  Then you can break out smaller bits without it getting overwhelming.

In your `certificate/index.tsx`, don't be afraid to use an `if`:

```ts
if(loading) {
  return <Loading />
}

if(certificate) {
  return <CertificateDetails certificate={certificate} />
}

return <RequestCertificate />
```

app/components/certificate/certificate-display.tsx
```diff
@@ -0,0 +1,111 @@
...
+}
+
+export default function CertificateDisplay(props: Props) {
+  const toast = useToast();
+
+  function onCopy() {
+    let key = props.cert.certificate;
+    let val = 'Public Key';
+
+    if (props.priv) {
```
humphd, line 31: Instead of doing this, pass down the `title` to show (e.g., `'Private Key'`) and also the `value` to use (i.e., the specific PEM string to show).  This component shouldn't have to know anything about what it's showing, do that above it.

app/components/certificate/certificate-display.tsx
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  return (
+    <Flex flexDirection="column" gap="5">
+      <Flex flexDirection="column" gap="4">
+        <HStack gap="4">
+          <Heading as="h4" size="sm">
+            {props.priv ? 'Private Key' : 'Public Key'}
+          </Heading>
+          <Tooltip label="Copy Public Key">
```
humphd, line 51: This `public` vs. `private` logic should get figured out via props vs hard-coding like this and below.

app/components/certificate/certificate-display.tsx
```diff
@@ -0,0 +1,111 @@
...
+              _hover={{
+                background: 'whitesmoke',
+                color: 'teal.500',
+              }}
+              aria-label="Copy to Clipboard"
+              icon={<CopyIcon fontSize="md" />}
+              onClick={() => onCopy()}
+            />
+          </Tooltip>
+          <Tooltip label="Download Public Key">
```
humphd, line 65: and here

app/components/certificate/certificate-display.tsx
```diff
@@ -0,0 +1,111 @@
...
+                      status: 'success',
+                    })
+                  }
+                />
+              }
+            />
+          </Tooltip>
+        </HStack>
+        <Text>
+          {props.priv
```
humphd, line 91: Pass this down as `displayText` or `description` or something vs. doing this.

app/components/certificate/description.tsx
```diff
@@ -0,0 +1,36 @@
+import { Flex, Heading, Text, HStack, VStack } from '@chakra-ui/react';
+import type { Certificate } from '@prisma/client';
+
+interface Props {
+  request: Boolean;
+  cert: Certificate;
```
humphd, line 6: Don't require knowledge of a certificate.  Pass down the dates instead.

app/components/certificate/description.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <Text>
+        {props.request
+          ? "Certificate: Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s"
+          : "Initial: Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s"}
+      </Text>
+      {props.request ? (
+        <HStack gap="6" marginTop="2">
+          <VStack>
+            <Text fontWeight="bold">Created On</Text>
+            <Text>{props.cert.validFrom.toDateString()}</Text>
```
humphd, line 24: Use `chakra.time` for a `<time>` element, and use the `datatime` attribute to set a machine readable value.

app/components/certificate/loading.tsx
```diff
@@ -0,0 +1,18 @@
+import { Center, Flex, Box, Image, Heading } from '@chakra-ui/react';
+
+import image from 'img/undraw_Processing_re_tbdu.png';
```
humphd, line 3: I'm unclear how Remix handles assets.  Are these supposed to go in `public/` or in the tree like this?

app/components/certificate/request.tsx
```diff
@@ -0,0 +1,21 @@
+import { Flex, Heading, Text, HStack } from '@chakra-ui/react';
+import type { Certificate } from '@prisma/client';
+
+interface Props {
+  cert: Certificate;
```
humphd, line 5: Pass the specific thing down you want to show: `domainName`

app/routes/__index/certificate/index.tsx
```diff
@@ -1,11 +1,65 @@
...
+    validFrom: curDate,
+    validTo: new Date(curDate.getFullYear(), curDate.getMonth() + 6),
+  };
+
+  if (loading) {
+    return <Loading />;
+  } else {
```
humphd, line 39: No `else` after `return`

app/routes/__index/certificate/index.tsx
```diff
@@ -1,11 +1,65 @@
...
 
 export default function CertificateIndexRoute() {
-  return (
-    <div>
-      <Heading as="h1" size="3xl">
-        Certificate Info or option to request a certificate
-      </Heading>
-    </div>
-  );
+  const [request, setRequest] = useState(false);
```
humphd, line 11: What does `request` represent?  Would this not be `certificate` (i.e., you'd show the certificate display only if you have a certificate)?

app/components/certificate/loading.tsx
humphd: We'll also need to include the license for this image, or we are breaking the license.

humphd (changes requested):

app/components/certificate/certificate-display.tsx
```diff
@@ -0,0 +1,111 @@
...
+  AccordionItem,
+  AccordionButton,
+  AccordionIcon,
+  AccordionPanel,
+  useToast,
+} from '@chakra-ui/react';
+import { DownloadIcon, CopyIcon } from '@chakra-ui/icons';
+import type { Certificate } from '@prisma/client';
+
+interface Props {
```
Myrfion, line 19: I started using naming for props as name of component + Props, so I think we should stick to start naming, so it would be like `CertificateDisplayProps`

app/components/certificate/certificate-display.tsx
```diff
@@ -0,0 +1,111 @@
...
+  cert: Certificate;
+  priv: Boolean;
```
Myrfion, line 21: The naming is not very clear to me, I would suggest to add prefix is and put the whole word, so like `isPrivate`

app/components/certificate/certificate-display.tsx
```diff
@@ -0,0 +1,111 @@
...
+}
+
+export default function CertificateDisplay(props: Props) {
```
Myrfion, line 24: I would suggest to destruct props, so to do something like `const {isPrivate, certificate} = props` looks cleaner like that

app/routes/__index/certificate/index.tsx
```diff
@@ -1,11 +1,65 @@
...
+    return (
+      <Center>
+        <Flex flexDirection="column" gap="5" width="4xl">
+          {request && (
+            <>
+              <Description request={true} cert={certificate} />
+              <Divider />
+              <CertificateDisplay cert={certificate} priv={false} />
+              <Divider />
+              <CertificateDisplay cert={certificate} priv={true} />
```
Myrfion, line 49: if you wanna make boolean attribute true you could just write it as:
```typescript
<CertificateDisplay 
cert={certificate} 
priv  />
```
so without `={true}`

app/routes/__index/certificate/index.tsx
```diff
@@ -1,11 +1,65 @@
...
+
+  function handleSubmit() {
+    setLoading(true);
+    setTimeout(() => {
+      setRequest(true);
+      setLoading(false);
+    }, 5000);
+  }
+
+  let curDate: Date = new Date(2023, 2);
```
Myrfion, line 22: value is immutable, so better make it `const` + I think it is better to avoid to main shortcuts is variable naming, just to make it more clear, so `currentDate`

app/components/certificate/request.tsx
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export default function Request(props: Props) {
```
Myrfion, line 8: Sam thing on props for component naming

app/components/certificate/description.tsx
```diff
@@ -0,0 +1,36 @@
...
+          </VStack>
+          <VStack>
+            <Text fontWeight="bold">Expires On</Text>
+            <Text>{props.cert.validTo.toDateString()}</Text>
+          </VStack>
+        </HStack>
+      ) : (
+        ''
```
Myrfion, line 32: Since you have nothing to return in the `else` case I think it is better to use `and` operator instead of the ternary one. So it would be just 
```typescript
{props.request && <SomethingYouWant />
```

Myrfion (changes requested):
Hi, PR looks good! I added some minor change requests that would improve code quality in my opinion

app/components/certificate/request.tsx
Myrfion: Also name itself seems to be very general, especially the work `Request` Maybe do something like `CertificateRequestView`

app/routes/__index/certificate/index.tsx
Eakam1007: This might be an opinion thing, but I like this explicit style of specifying true more

app/components/certificate/certificate-display.tsx
Ririio: you got a point

app/components/certificate/certificate-display.tsx
Ririio: Ohhh I never knew that was possible

app/components/certificate/certificate-display.tsx
Ririio: I'm a bit confused as to what you mean, do you mind explaining it one more time?

app/components/certificate/certificate-display.tsx
Ririio: Maybe I can pass a string on the prop, and use that as a state of either 'Private' and 'Public' to be used on the text? Would that work?

app/components/certificate/description.tsx
Ririio: I'm sorry I can't seem to find the document relating to `chakra.time` do you happen to have the link for it?

app/components/certificate/loading.tsx
Ririio: I'm also not sure... Alex might know a bit more about it, because she has also have files on the `img` directory

app/routes/__index/certificate/index.tsx
Ririio: the 'request' is a state when the certificate has already been requested (true) or not (false)

app/routes/__index/certificate/index.tsx
Ririio: I think it's good to follow the format established by the first props, that's just my opinion though...

app/routes/__index/certificate/index.tsx
Ririio: I miss that lul, and I agree with the naming it's much better your way

app/components/certificate/description.tsx
Ririio: That's true

app/components/certificate/certificate-display.tsx
Ririio: I decided to create a `switch` and 3 new variables that will be used for the `key`, `currentState` and `Description`. This should remove any need for hard coding, because the output is depended on the current state of the `isPrivate` props

app/components/certificate/request.tsx
Ririio: I am not quite sure what you mean by this?

Eakam1007 (commented):
Could you please add e2e tests for this or create a issue for them?

app/components/certificate/certificate-display.tsx
```diff
@@ -0,0 +1,118 @@
...
+}
+
+export default function CertificateDisplay(props: CertificateDisplayProps) {
+  const { isPrivate, certificate } = props;
+  const toast = useToast();
+  let description: string = '',
+    currentState: string = '',
+    key: string = '';
+
+  switch (isPrivate) {
```
humphd, line 31: I don't think this makes sense.  The point of props is to pass down the values to be used.  You don't need to put any logic into this component, and can simply pass `currentState`, `key`, `description` down as props from the parent component.

Also, I would change the names:

* `currentState` is really `name` or `title`
* `key` is really `value`, which is either a `certificate` or a `private key` (i.e., `key` doesn't make sense as a common name)

app/components/certificate/certificate-display.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  return (
+    <Flex flexDirection="column" gap="5">
+      <Flex flexDirection="column" gap="4">
+        <HStack gap="4">
+          <Heading as="h4" size="sm">
+            {`${currentState} key`}
+          </Heading>
+          <Tooltip label={`Copy ${currentState} Key`}>
```
humphd, line 62: Throughout this file, you append ` Key` to strings.  As I mention above, that doesn't work for a certificate.  I would use the `name`/`title` prop and have it include everything you need: `Copy Certificate` and `Copy Private Key` where `title='Certificate'` and `title='Private Key'`.

app/components/certificate/description.tsx
```diff
@@ -0,0 +1,44 @@
+import { Flex, Heading, Text, HStack, VStack } from '@chakra-ui/react';
+
+interface DescriptionProps {
+  request: Boolean;
```
humphd, line 4: What is the meaning of `request`?  The term is overloaded, and we probably need something more specific.

app/components/certificate/description.tsx
```diff
@@ -0,0 +1,44 @@
...
+  validFrom: Date;
+  validTo: Date;
+}
+
+export default function DescriptionSection(props: DescriptionProps) {
+  const { request, validFrom, validTo } = props;
+  let description: String = '';
+
+  switch (request) {
```
humphd, line 13: Don't use switch for a boolean comparison.

app/components/certificate/description.tsx
```diff
@@ -0,0 +1,44 @@
...
+    case true:
+      description =
```
humphd, line 15: Don't embed logic into components when you can pass props (i.e., pass `description` down if its value can change).

app/components/certificate/description.tsx
```diff
@@ -0,0 +1,44 @@
...
+      break;
+  }
+
+  return (
+    <Flex flexDirection="column" gap="3" marginTop="14">
+      <Heading as="h1" size="xl">
+        Certificate
+      </Heading>
+      <Text>{description}</Text>
+      {request && (
```
humphd, line 30: So does `request` here mean "should show this UI"?  If so, name it something like that.

app/components/certificate/loading.tsx
humphd: I was looking for someone to go read the docs, which are here https://remix.run/docs/en/v1/guides/migrating-react-router-app#asset-imports.

app/components/certificate/request.tsx
```diff
@@ -0,0 +1,22 @@
+import { Flex, Heading, Text, HStack } from '@chakra-ui/react';
+
+interface CertificateRequestProps {
+  subject: String;
```
humphd, line 4: `String` with a capital `S` is the JavaScript constructor function.  You want `string` lower-case 's' for TypeScript.

app/routes/__index/certificate/index.tsx
humphd: If this is whether or not the certificate has been requested, use that: `certificateRequested`

app/components/certificate/loading.tsx
humphd: Another question: does undraw not supply these as SVGs?  They will scale much better if so.

humphd (changes requested):

app/components/certificate/certificate-display.tsx
```diff
@@ -0,0 +1,100 @@
...
+  AccordionIcon,
+  AccordionPanel,
+  useToast,
+} from '@chakra-ui/react';
+import { DownloadIcon, CopyIcon } from '@chakra-ui/icons';
+
+interface CertificateDisplayProps {
+  title: string;
+  value: string;
+  description: string;
```
humphd, line 21: Nice, so much better

app/components/certificate/certificate-display.tsx
```diff
@@ -0,0 +1,100 @@
...
+}
+
+export default function CertificateDisplay(props: CertificateDisplayProps) {
+  const { title, description, value } = props;
```
humphd, line 25: nit: 

```ts
export default function CertificateDisplay({ title, description, value }: CertificateDisplayProps) {
```

app/components/certificate/certificate-display.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const toast = useToast();
+
+  function onCopy() {
+    navigator.clipboard.writeText(value);
+    toast({
+      title: `${title} was copied to clipboard`,
```
humphd, line 31: "the clipboard"

app/components/certificate/loading.tsx
```diff
@@ -0,0 +1,18 @@
+import { Center, Flex, Box, Image, Heading } from '@chakra-ui/react';
+
+import undrawSvg from 'img/undraw_processing_re_tbdu.svg';
```
humphd, line 3: We're still missing the license on this image, it needs to be included in our repo somehow (maybe in `img/*`)

app/routes/__index/certificate/index.tsx
```diff
@@ -1,11 +1,83 @@
...
+            />
+            <Divider />
+            <CertificateDisplay
+              title="Private Key"
+              value={certificate.privateKey}
+              description="Private: Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s"
+            />
+          </>
+        )}
+        {!certificateRequested && (
```
humphd, line 66: This pattern feels off to me.  We have a state that will result in one of two component sub-trees being rendered in this view.  I think you should make two components and do:

```ts
{ certificateRequested ? <View1 /> : <View2 /> }
```

Maybe name them:

1. CertificateRequested
2. CertificateAvailable

There might be something better, not sure.

humphd (changes requested):

app/components/certificate/certificate-display.tsx
Ririio: oh that makes it much better than calling it again inside

app/routes/__index/certificate/index.tsx
Ririio: I actually thought about doing this on my previous commits, but I can't seem to figure out how I'll handle the button for `request certificate`

Ririio: oh it's actually fairly simple to do...

app/components/certificate/certificate-available.tsx
```diff
@@ -0,0 +1,41 @@
...
+  publicKey: string;
+  privateKey: string;
+}
+
+export default function CertificateAvailable({
+  validFrom,
+  validTo,
+  publicKey,
+  privateKey,
+}: CertificateRequestProps) {
```
humphd, line 18: `CertificateAvailableProps` to match the component name

app/components/certificate/certificate-request.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+interface CertificateRequestProps {
+  subject: string;
+  validFrom: Date;
+  validTo: Date;
+  setLoading: React.Dispatch<React.SetStateAction<boolean>>;
+  setCertificateRequested: React.Dispatch<React.SetStateAction<boolean>>;
+}
+
+export default function CertificateRequestView({
```
humphd, line 13: `CertificateRequestViewProps` to match component name

app/components/certificate/certificate-request.tsx
```diff
@@ -0,0 +1,51 @@
+import { Flex, Heading, Text, HStack, Button } from '@chakra-ui/react';
+
+import Description from './description';
+
+interface CertificateRequestProps {
+  subject: string;
+  validFrom: Date;
+  validTo: Date;
+  setLoading: React.Dispatch<React.SetStateAction<boolean>>;
```
humphd, line 9: These seem wrong.  Take an `onClick` in the props, and move the `handleSubmit` function up to the parent component.

app/components/certificate/certificate-request.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <Heading as="h3" size="md" marginBottom="3">
+          Domain Name
+        </Heading>
+        <Text>Lorem Ipsum is simply dummy text of the printing and typesetting industry</Text>
+
+        <HStack>
+          <Text fontWeight="bold">Domain Name:</Text> <Text>{subject}</Text>
+        </HStack>
+      </Flex>
+      <Button width="3xs" shadow="xl" onClick={handleSubmit}>
```
humphd, line 46: This `onClick` should call into the function passed down on props, and not bother doing anything in this component (put logic up the component tree).

app/components/certificate/description.tsx
```diff
@@ -0,0 +1,36 @@
+import { Flex, Heading, Text, HStack, VStack } from '@chakra-ui/react';
+
+interface DescriptionProps {
```
humphd, line 3: `DescriptionSectionProps` to match component name

humphd (changes requested):
This is looking better and better.  A few more things

app/components/certificate/certificate-request.tsx
Ririio: I was a bit reluctant, because I'm not quite familiar with how we handle the data types for functions. Should be working on the next commit

app/components/certificate/certificate-request.tsx
```diff
@@ -0,0 +1,41 @@
+import { Flex, Heading, Text, HStack, Button } from '@chakra-ui/react';
+
+import Description from './description';
+
+interface CertificateRequestViewProps {
+  subject: string;
+  validFrom: Date;
+  validTo: Date;
+  handleSubmit: () => void;
```
humphd, line 9: To be consistent with how React does this, you'd name this `onRequest`.  Event handlers use the `on*` prefix.

humphd (changes requested):
One small change, and this is r+ from me

app/components/certificate/certificate-request.tsx
Ririio: Ohh I'll remember that

humphd (approved):

Eakam1007 (approved):
The page is not really responsive: 

image

Though this can be addressed in another issue, if you would like to create one. And same goes for e2e tests for this

Otherwise, looks good

PR #233, opened 2023-02-16
Improve domain name validation

### LINKED ISSUE #185: Improve regex for domain name validation
Current validation is a simple regex which is not reliable to filter unwanted domain name. Introduce `validator` library to the project and improve validation logic. 

Domain names...
- should be `[name].[studentId].starchart.com`
- can't start or end with -
- can use A-Za-z0-9- (upper case is fine, but we'll convert to lower)
- can't contain multiple `-` consecutively

Genne23v:
Resolve #185 

Adding domain name validation as following rules

- Domain name pattern should be `[name].[studentId].rootDomain.com(.)`
- Domain name can contain only alphanumerical characters, `-`, and `_`
- Domain name should not start or end with `-`
- Domain name cannot contain multiple consecutive `-` or `_`
- Domain name can contain uppercase in UI but it is converted to lowercase before validation
- `studentId` only contains alphanumerical characters

Also domain name is lowercased before adding it to Route 53, but original name is stored in DB

### DIFF
--- a/app/lib/dns.server.ts
+++ b/app/lib/dns.server.ts
@@ -6,6 +6,7 @@ import {
   GetChangeCommand,
 } from '@aws-sdk/client-route-53';
 import { isIPv4, isIPv6 } from 'is-ip';
+import isFQDN from 'validator/lib/isFQDN';
 
 import type {
   CreateHostedZoneResponse,
@@ -48,18 +49,23 @@ export const createHostedZone = async (domain: string) => {
   }
 };
 
-export const createRecord = (type: RecordType, name: string, value: string) => {
+export const createRecord = (username: string, type: RecordType, name: string, value: string) => {
   try {
-    return upsertRecord(type, name, value);
+    return upsertRecord(username, type, name, value);
   } catch (error) {
     logger.warn(error);
     throw new Error(`Error occurred while creating resource record`);
   }
 };
 
-export const upsertRecord = async (type: RecordType, name: string, value: string) => {
+export const upsertRecord = async (
+  username: string,
+  type: RecordType,
+  name: string,
+  value: string
+) => {
   try {
-    if (!isNameValid(name)) {
+    if (!isNameValid(name, username)) {
       throw new Error('Invalid name provided');
     }
 
@@ -98,9 +104,14 @@ export const upsertRecord = async (type: RecordType, name: string, value: string
   }
 };
 
-export const deleteRecord = async (type: RecordType, name: string, value: string) => {
+export const deleteRecord = async (
+  username: string,
+  type: RecordType,
+  name: string,
+  value: string
+) => {
   try {
-    if (!isNameValid(name)) {
+    if (!isNameValid(name, username)) {
       throw new Error('Invalid name provided');
     }
 
@@ -157,8 +168,31 @@ export const getChangeStatus = async (changeId: string) => {
   }
 };
 
-const isNameValid = (name: string) => {
-  return /^[a-z0-9-]+.[a-z0-9-]+.[a-z]+.[a-z]+\.?$/.test(name);
+/* Domain name rules
+1. Domain name pattern should be [name].[username].rootDomain.com
+2. Domain name can contain only alphanumerical characters, '-', and '_'
+3. Domain name should not start or end with -
+4. Domain name cannot contain multiple consecutive '-' or '_'
+5. Domain name can contain uppercase in UI but it is converted to lowercase before validation */
+export const isNameValid = (name: string, username: string) => {
+  const rootDomain = process.env.ROOT_DOMAIN!;
+
+  /* Domain name must end with username and root domain. 
+  Here it removes username and root domain, 
+  to validate only subdomain that user has input */
+  const toRemove = `.${username}.${rootDomain}`;
+  if (!name.endsWith(toRemove)) {
+    return false;
+  }
+  const subdomain = name.substring(0, name.length - toRemove.length);
+
+  //It only validates subdomain name, not username and root domain
+  return (
+    /^(?!.*[-_]{2,})(?!^[-])[a-z0-9_-]+[a-z0-9]$/.test(subdomain) &&
+    isFQDN(name, {
+      allow_underscores: true,
+    })
+  );
 };
 
 const isValueValid = (type: RecordType, value: string) => {
--- a/app/lib/domains.server.ts
+++ b/app/lib/domains.server.ts
@@ -10,17 +10,17 @@ interface DnsRecord {
   value: string;
 }
 
-interface DomainRecord {
+interface DBRecord {
   id?: Record['id'];
-  username?: Record['username'];
+  username: Record['username'];
   description?: Record['description'];
   course?: Record['course'];
   ports?: Record['ports'];
 }
 
-type Domain = DnsRecord & DomainRecord;
+export type DomainRecord = DnsRecord & DBRecord;
 
-export const createUserDomain = async (data: Domain) => {
+export const createUserRecord = async (data: DomainRecord) => {
   if (!data.username) {
     throw new Error('Username is not provided');
   }
@@ -38,7 +38,7 @@ export const createUserDomain = async (data: Domain) => {
   }
 
   const [, result] = await Promise.all([
-    dns.createRecord(data.type, data.name, data.value),
+    dns.createRecord(data.username, data.type, data.name.toLowerCase(), data.value),
     prisma.record.create({
       data: {
         username: data.username,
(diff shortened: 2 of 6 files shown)


app/lib/dns.server.ts
```diff
@@ -157,8 +158,30 @@ export const getChangeStatus = async (changeId: string) => {
...
+  if (name.endsWith('.')) {
+    subdomain = name.replace(`.${rootDomain}.`, '');
+  } else {
+    subdomain = name.replace(`.${rootDomain}`, '');
+  }
+  return (
+    /^(?!.*[-_]{2,})(?!^[-])[a-z0-9_-]+[^-_]\.[a-z0-9]+$/.test(subdomain) &&
+    validator.isFQDN(name, {
+      allow_underscores: true,
+      allow_trailing_dot: true,
```
dadolhay, line 182: I think this should be false, we should not allow both variants IMO

app/lib/dns.server.ts
```diff
@@ -157,8 +158,30 @@ export const getChangeStatus = async (changeId: string) => {
...
+  let subdomain: string;
+
+  //It only validates domain name and studentId, not root domain
+  if (name.endsWith('.')) {
+    subdomain = name.replace(`.${rootDomain}.`, '');
+  } else {
+    subdomain = name.replace(`.${rootDomain}`, '');
+  }
+  return (
+    /^(?!.*[-_]{2,})(?!^[-])[a-z0-9_-]+[^-_]\.[a-z0-9]+$/.test(subdomain) &&
```
dadolhay, line 179: I would remove [studentId] from the subdomain too like:
```js
const postfix = `.${studentId}.${rootDomain}`;
if (!name.endsWith(postfix)) {return false;};

const subdomain = name.substring(0, name.length - postfix);
```

now you can validate only the [name] part, without having to worry about any dots
Very impressive regex work but 'aaa..bbb' still passes :( possible that it would be caught by isFQDN

app/lib/domains.server.ts
```diff
@@ -18,7 +18,7 @@ interface DomainRecord {
   ports?: Record['ports'];
 }
 
-type Domain = DnsRecord & DomainRecord;
+export type Domain = DnsRecord & DomainRecord;
 
 export const createUserDomain = async (data: Domain) => {
```
dadolhay, line 23: Could we rename these to records? I strongly feel about the term `domain` is not correct here

dadolhay (changes requested):

app/lib/dns.server.ts
Genne23v: Thanks for checkout out! Yes, it would be much easier without `studentId`. I will have to speak with @sfrunza13 if I can easily get `studentId` info. I will add more test cases and improve it.

app/lib/domains.server.ts
Genne23v: I was thinking an easy name for outsiders, but I think we should have a consistency as I look at the table and everything. I will update them all.

app/lib/dns.server.ts
Genne23v: There was a suggestion to add trailing dot. I think it's because Route 53 automatically adds a dot somewhere. But it would very unlikely get `root domain` from Route 53. 

I thought you would need underscore for `_acme...`

dadolhay: yes, I need the underscore, but I'd not use the trailing dot. If Route53 needs it, I'd add it at the invocation of route53 and abstract away this complexity from the rest of the system

app/lib/dns.server.ts
humphd: We can't get student IDs.  We're only going to get their username (bit before `@myseneca.ca` or `@senecacollege.ca` for employees).  It may contain a `.`, letters, numbers, a `-`.

dadolhay: I think we are talking about the same thing, it's just not named correctly ... What would be a good name for that? In an email that would be called `localpart` but I'm sure we are not going to call that

humphd: As far as the SAML claim we get, it's the `SAML account name`, so maybe `account name` or `username` is good enough?

package.json
```diff
@@ -59,6 +59,7 @@
     "react-dom": "^18.2.0",
     "samlify": "^2.8.8",
     "tiny-invariant": "^1.3.1",
+    "validator": "^13.9.0",
```
humphd, line 62: If we're going to pull in this dependency, we could use it to replace the IPv4 and IPv6 dependency we brought in before.  That is, this module can do a lot of different checks for us, so we can reduce the number of dependencies we have.

Can you file an issue to update that later?

humphd (approved):

package.json
Genne23v: Yes, I will

app/lib/dns.server.ts
```diff
@@ -157,8 +168,23 @@ export const getChangeStatus = async (changeId: string) => {
...
-  return /^[a-z0-9-]+.[a-z0-9-]+.[a-z]+.[a-z]+\.?$/.test(name);
+/* Domain name rules
+1. Domain name pattern should be [name].[username].rootDomain.com
+2. Domain name can contain only alphanumerical characters, '-', and '_'
+3. Domain name should not start or end with -
+4. Domain name cannot contain multiple consecutive '-' or '_'
+5. Domain name can contain uppercase in UI but it is converted to lowercase before validation */
+export const isNameValid = (name: string, username: string) => {
+  const rootDomain = process.env.ROOT_DOMAIN!;
+  let subdomain = name.replace(`.${username}.${rootDomain}`, '');
```
dadolhay, line 179: The problem with this solution is, that if replace does not find `.${username}.${rootDomain}`, it will do nothing (return the unchanged string). This is why I suggested that you do the .endsWith() first to confirm the correct structure.

Like, as is now, I think If I sent in 'localhost' as `name` (only that, nothing else), this would pass

dadolhay: (Locahost is usually treated as a fully qualified domain name to my best knowledge)

dadolhay (changes requested):

app/lib/dns.server.ts
Genne23v: Absolutely valid suggestion. I will update.

app/lib/dns.server.ts
```diff
@@ -157,8 +168,28 @@ export const getChangeStatus = async (changeId: string) => {
...
+  let subdomain: string = '';
+
+  if (!name.endsWith(`.${username}.${rootDomain}`)) {
+    return false;
+  }
+  subdomain = name.replace(`.${username}.${rootDomain}`, '');
+
+  //It only validates subdomain name, not username and root domain
+  return (
+    /^(?!.*[-_]{2,})(?!^[-])[a-z0-9_-]+[^-_]$/.test(subdomain) &&
```
dadolhay, line 188:
```suggestion
    /^(?!.*[-_]{2,})(?!^[-])[a-z0-9_-]+[a-z0-9]$/.test(subdomain) &&
```
the difference is this does not accept . at the end

app/lib/dns.server.ts
```diff
@@ -157,8 +168,28 @@ export const getChangeStatus = async (changeId: string) => {
...
+4. Domain name cannot contain multiple consecutive '-' or '_'
+5. Domain name can contain uppercase in UI but it is converted to lowercase before validation */
+export const isNameValid = (name: string, username: string) => {
+  const rootDomain = process.env.ROOT_DOMAIN!;
+  let subdomain: string = '';
+
+  if (!name.endsWith(`.${username}.${rootDomain}`)) {
+    return false;
+  }
+  subdomain = name.replace(`.${username}.${rootDomain}`, '');
```
dadolhay, line 184:
```suggestion
  const toRemove = `.${username}.${rootDomain}`;
  if (!name.endsWith(toRemove)) {
    return false;
  }
  subdomain = name.substring(0, name.length - toRemove.length);
```
Sorry for picking on this, but I'd also make sure that this is removed at the end of the string

dadolhay (changes requested):

app/lib/dns.server.ts
```diff
@@ -157,8 +168,29 @@ export const getChangeStatus = async (changeId: string) => {
...
-  return /^[a-z0-9-]+.[a-z0-9-]+.[a-z]+.[a-z]+\.?$/.test(name);
+/* Domain name rules
+1. Domain name pattern should be [name].[username].rootDomain.com
+2. Domain name can contain only alphanumerical characters, '-', and '_'
+3. Domain name should not start or end with -
+4. Domain name cannot contain multiple consecutive '-' or '_'
+5. Domain name can contain uppercase in UI but it is converted to lowercase before validation */
+export const isNameValid = (name: string, username: string) => {
+  const rootDomain = process.env.ROOT_DOMAIN!;
+  let subdomain: string = '';
```
humphd, line 179: There's no value in doing this.  Define your variable where you'll use it, and don't set it equal to a string you'll never use (empty string).

app/lib/dns.server.ts
```diff
@@ -157,8 +168,29 @@ export const getChangeStatus = async (changeId: string) => {
...
+
+  const toRemove = `.${username}.${rootDomain}`;
```
humphd, line 181: 181-185 needs a comment to explain why you're doing this.

humphd (changes requested):

dadolhay (approved):

humphd (approved):

Issue #43, opened 2023-01-27
Deploy the app to staging


humphd:
I had a good meeting with ITS today, and we have an initial plan for deploying.  I'll try to capture what we discussed here.

We're going to try using Docker Swarm mode to host staging and production.  I've written a [walkthrough](https://github.com/Seneca-CDOT/starchart/wiki/Docker#deployment) showing how it will work.

The basic idea is that we'll run 2 Linux VMs.  We'll have Docker running on both, and one will be a `manager` node and the other a `worker` node in a swarm cluster.  Because our app is so small, we won't need more than this, but we can easily add more nodes if we have to.

We'll run multiple copies (2?) of the Starchart container in production, which will let us have some fault-tolerance, load-balancing, and zero-downtime updates.  For staging, we probably only need a single instance of the container.

We can run both staging and production on the same swarm cluster.  ITS uses an [F5 hardware-based load balancer](https://www.f5.com/glossary/load-balancer), and they are going to configure it so that our apps get SSL certificates and handles routing to the appropriate ports.

Both staging and production will work like this:

|Route|Port|App|
|------|----|-----|
|`/hooks/deploy` | `9000`| [webhook](https://github.com/adnanh/webhook)|
|`/`     | `3000`| Starchart app|

When we push to GitHub, we'll run a GitHub Action to build our Docker image and push it to Docker Hub.  We'll then trigger an HTTP `POST` to a [webhook](https://github.com/adnanh/webhook) that we have running on the manager node.  It will send the new Docker image tag that we want to use and will trigger a script to update the service on the swarm cluster.

Our URLs are likely going to be:

|Environment| URL |
|------------|------|
|production| https://mycustomdomain.senecacollege.ca|
|staging | https://mycustomdomain-dev.senecacollege.ca|

So the F5 load balancer will route https://mycustomdomain-dev.senecacollege.ca/ to the staging Remix server running on port `3000` on the swarm cluster, and https://mycustomdomain-dev.senecacollege.ca/hooks/deploy to the webhook server running on port `9000`.  Production will do do something similar, but with 2 different ports.

We'll use [Docker Swarm Secrets](https://docs.docker.com/engine/swarm/secrets/) to manage things like our AWS credentials, secret keys, certificates, passwords, etc.  Our Remix app will be able to access those secrets by reading files at `/run/secrets/<secret_name>`. We can us something like https://github.com/hwkd/docker-secret for the secrets loading in development.

[...]

PR #235, opened 2023-02-16
Use shared redis connection in FlowProducer

humphd:
While working on #43, I ran into a bug where the container can't connect to Redis.  It turns out to be that if you don't specify redis connection options, the `FlowProducer` will try to use `localhost`.

This updates our `FlowProducer` to use the shared `redis` connection we have.

### DIFF
--- a/app/queues/certificate/certificate-flow.server.ts
+++ b/app/queues/certificate/certificate-flow.server.ts
@@ -8,6 +8,8 @@ import {
 import { orderCompleterQueueName, orderCompleterWorker } from './order-completer-worker.server';
 import { dnsCleanerQueueName, dnsCleanerWorker } from './dns-cleaner-worker.server';
 
+import { redis } from '~/lib/redis.server';
+
 import type { FlowJob } from 'bullmq';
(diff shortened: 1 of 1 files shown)


app/queues/certificate/certificate-flow.server.ts
```diff
@@ -8,6 +8,8 @@ import {
 import { orderCompleterQueueName, orderCompleterWorker } from './order-completer-worker.server';
 import { dnsCleanerQueueName, dnsCleanerWorker } from './dns-cleaner-worker.server';
 
+import { redis } from '~/lib/redis.server';
+
```
Eakam1007, line 12: Just out of curiosity, are the empty lines needed here?

Eakam1007 (approved):

dadolhay (approved):

app/queues/certificate/certificate-flow.server.ts
Eakam1007: Ah, didn't know there were three separations. Thanks

PR #236, opened 2023-02-17
#195 SSO and SLO additions

### LINKED ISSUE #195: Add SAML Server
Configure SP in code, probably hiding values in the .env file and the IDP from the metadata on the IDP container url.

Following https://github.com/remix-run/examples/pull/130/files/ec66b3060fac83eec2389eb0c96aad6d8ea4aed1#diff-02d2b71e481b2495b8a72af14f09fc28238298c7f1d19a540e37c9228985b0da SAMLServer file.

sfrunza13:
This Pull Request aims to close #195 

I used the previously added IDP and configured Starchart as a service provider that requests login and logout to it using redirect and post bindings. The SP metadata is generated by Samlify methods and then exposed on the /sp route.

The SAML response will have its assertions signed by the IDP.

The "POST-BINDING" for the logout actually is a get request and the response is found in the url.

## TO TEST THIS
You can take the code and pull it into a branch of your own or with gh cli or however you choose to do it and run the regular steps to run starchart. The login container should be up and running in docker and the login button will redirect to the IDP and the signout will also do the same. With the container running you can go here http://localhost:8081/simplesaml/saml2/idp/metadata.php?output=xhtml and see the IDP Metadata.

testing account for IDP:
username: user1
password: user1pass

You can also go into the network tab and check the flow of the SAML requests and responses through every HTTP request. The payload will have SAML response or SAML request and you can plug them into a [decoder](https://www.samltool.com/decode.php) and get the XML of the request or response. In this way you can see that the response for example has signed assertions.


sirinoks (commented):
Some questions here and there, overall it makes sense.
I'm lacking instruction on how to test this, and confirm if it runs/works

config/saml20-idp-hosted.php
```diff
@@ -26,7 +26,7 @@
     'authproc' => array(
         3 => array(
             'class' => 'saml:AttributeNameID',
-            'attribute' => 'email',
+            'attribute' => 'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name',
```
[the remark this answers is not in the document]
sfrunza13: This is one of the claims that comes in a SAML response, it's a way of identifying the attributes and what they mean, this one for example is name which identifies a unique subject, in our case the username. https://learn.microsoft.com/en-us/azure/active-directory/develop/reference-saml-tokens

app/routes/login/callback.tsx
```diff
@@ -0,0 +1,84 @@
...
+        remember: false,
+        redirectTo: '/',
+      });
+    }
+
+    // return to access denied if redirect in createUserSession did not take
+    return redirect('/access_denied');
+  } else {
+    //Request method is not post, why are you here?
+    return redirect('/');
```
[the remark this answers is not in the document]
sfrunza13: If someone I suppose in theory you could make a put or a delete request to this route and the action would be triggered

app/routes/logout/callback.tsx
```diff
@@ -0,0 +1,16 @@
+import type { LoaderArgs } from '@remix-run/node';
+import { redirect } from '@remix-run/node';
+import { logout } from '~/session.server';
```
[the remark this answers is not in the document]
sfrunza13: LoaderArgs and ActionArgs are types defined by Remix to add typesafety to Loading and Action server functions. We import it basically explicitly to use it as a type for the parameters to the loader or the action.

app/routes/login/callback.tsx
```diff
@@ -0,0 +1,84 @@
...
+    email: 'user1@myseneca.ca',
+    'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname': 'Johannes',
+    'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname': 'Kepler',
+    'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name': 'user1@myseneca.ca',
+    sAMAccountName: 'user1',
+    'http://schemas.microsoft.com/identity/claims/displayname': 'Johannes Kepler'
+  }
+}
+*/
+export const action = async ({ request }: ActionArgs) => {
```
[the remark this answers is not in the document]
sfrunza13: from what I understand this is just the way that the remix functions are written, within the curly braces we place all of the parameters that should come into the action and these are declared as being ActionArgs.

app/routes/login.tsx
```diff
@@ -11,17 +11,26 @@ import {
...
+  const user = await getUsername(request);
+
+  //If not then create a login request to the IDP's redirect binding
+  if (!user) {
+    const idp = await getIdp();
+    const { context } = sp.createLoginRequest(idp, 'redirect');
+    return redirect(context);
+  }
+  // fallback if someone accidentally landed here and was already logged in.
+  if (user) {
```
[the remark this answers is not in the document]
sfrunza13: I could probably do that 🤔

app/routes/login/callback.tsx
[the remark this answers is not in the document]
humphd: This is using two different ideas in JS/TS.

First, this says that `action` is a function which takes a single argument: an `Object` (`{...}` means this), and that we only care about using one of the properties on that `Object`: the `request` property.  When you explode an `Object` (or `Array`) like this, it is called [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment).  Consider various ways of writing it, arriving at the final version used here:

```js
function action(options) {
  const request = options.request;
}

function action(options) {
  const { request } = options;
}

function action({ request}) {
}
```

In TS we go further, and define the shape (i.e., names of the properties on the `Object`) and their types using `: <Type>`. In this case, `ActionArgs` is a type that defines an `Object` that has something called `request`, which itself has various properties and methods we can use.

humphd (changes requested):
This is amazing.  I'm thrilled to see SAML2 based SSO logins working for Starchart as part of 0.3!  Really great work on this.

I'd like to see us improve this code a bunch, but I don't want to block it landing in 0.3 either.  Can you please take a look at my feedback and decide how much you want to do before 0.3, and what you want to do in 0.4.  Anything that will happen later should get filed as follow-up issues.

Also, this code is easily testable via Playwright, and I think that should be your primary task in 0.4, so we can always know that it works going forward.  See my existing login/logout Playwright test code in Telescope for some ideas:

https://github.com/Seneca-CDOT/telescope/blob/master/src/api/sso/test/e2e/browser-util.js#L14-L57

I'm sure @Eakam1007 can provide guidance too.

.env.example
```diff
@@ -18,3 +18,7 @@ LOG_LEVEL=debug
 # Nodemailer config
 NOTIFICATIONS_EMAIL_USER="no-reply@senecacollege.ca"
 MAILHOG_SMTP_PORT=1025
+SAML_IDP_METADATA=http://localhost:8081/simplesaml/saml2/idp/metadata.php
+HOSTNAME = http://localhost`
```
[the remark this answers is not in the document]
humphd: Also, trailing backtick

.env.example
```diff
@@ -18,3 +18,7 @@ LOG_LEVEL=debug
 # Nodemailer config
 NOTIFICATIONS_EMAIL_USER="no-reply@senecacollege.ca"
 MAILHOG_SMTP_PORT=1025
+SAML_IDP_METADATA=http://localhost:8081/simplesaml/saml2/idp/metadata.php
```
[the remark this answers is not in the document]
humphd: Thinking about this a bit more.  It might be better to have all of the exports from the `saml.server.ts` module be `async`.  This would allow us to pull the IdP metadata on startup and use that (i.e., first request would get the metadata and config the idp, subsequent runs would be faster since we already have it).

.env.example
```diff
@@ -18,3 +18,7 @@ LOG_LEVEL=debug
...
+HOSTNAME = http://localhost`
+# Our apps's Entity ID, which is also the URL to our metadata.
+SAML_ENTITY_ID=http://host.docker.internal:8080/sp
```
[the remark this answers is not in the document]
sfrunza13: I am on windows, although I am using wsl for docker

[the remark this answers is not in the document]
humphd: That's a good sign.  If it works there, it likely "works" in general.  I'll resolve this until someone files an issue on it.

sfrunza13:
@humphd I have been looking at this a lot over the past day and I don't know if it works exactly as intended. Validation of signature along with what is meant to be signed (right now I am trying to have the assertion signed but upon further inspection don't know if that's working and whether I'm supposed to be signing the assertions or the message), and the logout flow along with parsing the logout response might need to be added or changed. I don't think I can do all of that for this MS, I am going to do some cleaning up of the current code and try to get this in and change it moving forward if that's ok with you.

humphd:
> @humphd I have been looking at this a lot over the past day and I don't know if it works exactly as intended. Validation of signature along with what is meant to be signed (right now I am trying to have the assertion signed but upon further inspection don't know if that's working and whether I'm supposed to be signing the assertions or the message), and the logout flow along with parsing the logout response might need to be added or changed. I don't think I can do all of that for this MS, I am going to do some cleaning up of the current code and try to get this in and change it moving forward if that's ok with you.
> 
> It seems to work but I think I have missed the mark on a lot of important things.

I agree that nailing down all the various pieces of this is beyond the scope of 0.3.

Since what you have works in development, I'm inclined to say that we should fix the code-level issues I've highlighted, and file issues on the rest to do in 0.4+.

Let's not block this on being perfect.  You've moved us ahead a long way, and that's what a weekly milestone is all about.

app/routes/login.tsx
```diff
@@ -11,17 +11,24 @@ import {
 } from '@chakra-ui/react';
 import { LockIcon } from '@chakra-ui/icons';
 import type { ActionArgs } from '@remix-run/node';
+import { redirect } from '@remix-run/node';
 import { Form } from '@remix-run/react';
 
-import { createUserSession } from '~/session.server';
+import { getUsername } from '~/session.server';
+import { getIdp, sp } from '~/saml.server';
```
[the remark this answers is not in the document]
sfrunza13: I added some functions in the saml server so we dont have to export sp and idp but export those instead. Is that sufficient or should there be a seperate file that is the only one to ever import saml.server?

.env.example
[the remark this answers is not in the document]
sfrunza13: I am trying the following but it is claiming that neither `http://localhost:8081/simplesaml/saml2/idp/metadata.php` or `http://host.docker.internal:8081/simplesaml/saml2/idp/metadata.php` exist
```
//Take the metadata stood up by the IDP and use it as the metadata for our IDP object
const idp = samlify.IdentityProvider({
  metadata: readFileSync(`${process.env.SAML_IDP_METADATA}`),
});
```

What am I doing wrong?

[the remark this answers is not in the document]
sfrunza13: Wait, I think I misunderstood, you would like to have the idp metadata stored as a file local to starchart, is that correct?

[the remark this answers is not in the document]
sfrunza13: In the last commit I had it read from the idp config locally, when I try to load idp once in saml server at the top level it gives me the error that await can not be used at the top level.

app/routes/login/callback.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  if (request.method != 'POST') {
+    //Request method is not post, why are you here?
+    return redirect('/');
+  }
+
+  const body = Object.fromEntries(formData);
+  const extract = await parseLoginResponse(body);
+
+  //Try and extract the username and see if there is an existing user by that name
```
humphd, line 54: nit: sometimes you do `//nospace` and other times `// space`.  Can you always do the latter (i.e., include a space after `//`)?

app/routes/login/callback.tsx
```diff
@@ -0,0 +1,81 @@
...
+  if (extract.attributes.sAMAccountName) {
```
humphd, line 55: Same advice as before, flip your logic and put the error/shorter case first:

```ts
if(!extract.attribues.sAMAccountName) {
   // TODO: Create some form of access denied notification to redirect to here
  return redirect('/access_denied');
}

// bulk of code here, not indented...
```

app/routes/login/callback.tsx
```diff
@@ -0,0 +1,81 @@
...
+    'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name': 'user1@myseneca.ca',
+    sAMAccountName: 'user1',
+    'http://schemas.microsoft.com/identity/claims/displayname': 'Johannes Kepler'
+  }
+}
+*/
+export const action = async ({ request }: ActionArgs) => {
+  const formData = await request.formData();
+
+  if (request.method != 'POST') {
```
humphd, line 46: Do this first, since a only a `POST` will have form data to parse.

humphd: Also, `!==` vs. `!=` in JS please.  The former does implicit type casting, which you never want.

app/routes/login/callback.tsx
```diff
@@ -0,0 +1,81 @@
...
+    //Request method is not post, why are you here?
+    return redirect('/');
+  }
+
+  const body = Object.fromEntries(formData);
+  const extract = await parseLoginResponse(body);
```
humphd, line 52: `extract` is an odd name.  `samlResponse`?

app/routes/logout/callback.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+//Logout, destroying the session with Starchart
+export const loader = async ({ request }: LoaderArgs) => {
+  const url = new URL(request.url);
+
+  const SAMLResponse = url.searchParams.get('SAMLResponse');
+
+  if (SAMLResponse) {
+    return await logout(request);
+  } else {
```
humphd, line 13: No else-after-return.

app/saml.server.ts
```diff
@@ -0,0 +1,53 @@
...
+    {
+      Binding: 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST',
+      Location: process.env.HOSTNAME + '/logout/callback',
+    },
+  ],
+});
+
+//Take the metadata stood up by the IDP and use it as the metadata for our IDP object
+const idp = samlify.IdentityProvider({
+  metadata: readFileSync(`${process.env.SAML_IDP_METADATA}`),
```
humphd, line 31: If it's a filename vs. the contents of a file, you should name it as such: `SAML_IDP_METADATA_FILE` or `_FILENAME`.

humphd (changes requested):
I don't see how to get this landed in 0.3, since this is pretty big and still needs work.  We need PRs to come in sooner so there is time to review in future.

.env.example
```diff
@@ -18,3 +18,7 @@ LOG_LEVEL=debug
 # Nodemailer config
 NOTIFICATIONS_EMAIL_USER="no-reply@senecacollege.ca"
 MAILHOG_SMTP_PORT=1025
+SAML_IDP_METADATA_FILE='./config/idp-metadata.xml'
```
humphd, line 21: Put the contents of this file in `dev-secrets/IDP_METADATA` vs. here.

app/routes/login.tsx
```diff
@@ -11,17 +11,24 @@ import {
...
-  });
+  // Check if a session with a username exists
+  const user = await getUsername(request);
+
+  // If not then create a login request to the IDP's redirect binding
+  if (!user) {
+    const context = await createLoginRequest();
+    return redirect(context);
+  }
+  if (user) {
```
humphd, line 29: This is unnecessary, based on line 25.  You can simply `return ...` here.

.env.example
humphd, line 21: Also, please separate this section with a blank line and a comment:

```ini

# SSO Config
...
```

app/routes/login/callback.tsx
```diff
@@ -0,0 +1,79 @@
...
+*/
+export const action = async ({ request }: ActionArgs) => {
+  if (request.method !== 'POST') {
+    // Request method is not post, why are you here?
+    return redirect('/');
+  }
+
+  const formData = await request.formData();
+  const body = Object.fromEntries(formData);
+  const samlResponse = await parseLoginResponse(body);
```
humphd, line 51: Because you only use `attributes`, you might as well pull that out here:

```ts
const { attributes } = await parseLoginResponse(body);
```

app/routes/login/callback.tsx
```diff
@@ -0,0 +1,79 @@
...
+      samlResponse.attributes['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress']
+    );
+  }
+
+  // Either way create a session
+  return createUserSession({
+    request: request,
+    username: username,
+    remember: false,
+    redirectTo: '/',
```
humphd, line 77: Do we have access to the `redirectTo` param here?  When you are on a page, and get redirected to the login, it usually keeps this param so you can get sent back to the place where you were.  It should be in `params` I think, but check my math.

app/saml.server.ts
```diff
@@ -0,0 +1,53 @@
...
+import { readFileSync } from 'fs';
+
+samlify.setSchemaValidator(validator);
+
+// Here we configure the service provider: https://samlify.js.org/#/sp-configuration
+
+const sp = samlify.ServiceProvider({
+  entityID: process.env.SAML_ENTITY_ID,
+  nameIDFormat: ['urn:oasis:names:tc:SAML:2.0:nameid-format:persistent'],
+  wantAssertionsSigned: true,
```
humphd, line 14: Is this true?

app/saml.server.ts
```diff
@@ -0,0 +1,53 @@
...
+    {
+      Binding: 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST',
+      Location: process.env.HOSTNAME + '/logout/callback',
+    },
+  ],
+});
+
+// Take the metadata stood up by the IDP and use it as the metadata for our IDP object
+const idp = samlify.IdentityProvider({
+  metadata: readFileSync(`${process.env.SAML_IDP_METADATA_FILE}`),
```
humphd, line 31: Pull this out of `secrets`

humphd (changes requested):
This needs a few fixes, but I was able to make it work.

You should update `CONTRIBUTING.md` to include a section on how to login (i.e., a table with usernames/passwords) so we can point our devs at it.  None of them are going to know how to login after you land this!

app/saml.server.ts
sfrunza13: http://localhost:8080/sp

```
<EntityDescriptor xmlns="urn:oasis:names:tc:SAML:2.0:metadata" xmlns:assertion="urn:oasis:names:tc:SAML:2.0:assertion" xmlns:ds="http://www.w3.org/2000/09/xmldsig#" entityID="http://host.docker.internal:8080/sp">
<SPSSODescriptor AuthnRequestsSigned="false" WantAssertionsSigned="true" protocolSupportEnumeration="urn:oasis:names:tc:SAML:2.0:protocol">
<NameIDFormat>urn:oasis:names:tc:SAML:2.0:nameid-format:persistent</NameIDFormat>
<SingleLogoutService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="http://localhost:8080/logout/callback"/>
<AssertionConsumerService index="0" Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="http://localhost:8080/login/callback"/>
</SPSSODescriptor>
</EntityDescriptor>
```

I think so

humphd: But that is being generated based on your config.  I'm saying, "Are we sure we want this assertion to be signed"?  Signed with what?  We have no keys in here?  I'm confused.

sfrunza13: Although I can't properly tell if the assertions are being signed or not, I would think it should say so in the extract but it doesn't seem to. This is something that I am going to spend some more time on.

app/routes/login/callback.tsx
sfrunza13: I remembered that Chris's example had something for this using RelayState so that the redirectUrl can persist to the callback. I added that.

app/saml.server.ts
sfrunza13: I thought the IDP signs the assertions, the SP is just saying that it wants the signature, and the IDP has its own set of keys and cert

app/saml.server.ts
```diff
@@ -0,0 +1,54 @@
...
+    {
+      Binding: 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST',
+      Location: process.env.HOSTNAME + '/logout/callback',
+    },
+  ],
+});
+
+// Take the metadata stood up by the IDP and use it as the metadata for our IDP object
+const idp = samlify.IdentityProvider({
+  metadata: readFileSync(secrets.SAML_IDP_METADATA_FILE),
```
humphd, line 32: Three things:

1. You don't need to `readFileSync()` on the `secrets`, they are already re-hydrated from the filesystem
2. Rename it to `SAML_IDP_METADATA` (drop the `_FILE` since they are all files)
3. You should add some code at the top of this to check that this secret exists, and throw if not.  We can't run the app without it.

```ts
const { SAML_IDP_METADATA } = secrets;
if(!SAML_IDP_METADATA) {
  throw new Error('Missing SAML_IDP_METADATA secret');
}
```

humphd (changes requested):
Looking good, a few small things.

sfrunza13:
Not entirely sure but after most recent changes I get the following whenever I try to log in:
```
TypeError: The "url" argument must be of type string. Received an instance of Object
    at new NodeError (node:internal/errors:399:5)
    at validateString (node:internal/validators:163:11)
    at Url.parse (node:url:176:3)
    at Object.urlParse (node:url:147:13)
    at buildRedirectURL (C:\Users\stefe\starchart\node_modules\samlify\src\binding-redirect.ts:55:25)
    at Object.loginRequestRedirectURL (C:\Users\stefe\starchart\node_modules\samlify\src\binding-redirect.ts:116:16)
    at ServiceProvider.createLoginRequest (C:\Users\stefe\starchart\node_modules\samlify\src\entity-sp.ts:71:32)
    at createLoginRequest (C:\Users\stefe\starchart\app\saml.server.ts:44:26)
    at action2 (C:\Users\stefe\starchart\app\routes\login.tsx:28:27)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

I'll take a look again tomorrow

humphd:
> Not entirely sure but after most recent changes I get the following whenever I try to log in:
> 
> ```
> TypeError: The "url" argument must be of type string. Received an instance of Object
>     at new NodeError (node:internal/errors:399:5)
>     at validateString (node:internal/validators:163:11)
>     at Url.parse (node:url:176:3)
>     at Object.urlParse (node:url:147:13)
>     at buildRedirectURL (C:\Users\stefe\starchart\node_modules\samlify\src\binding-redirect.ts:55:25)
>     at Object.loginRequestRedirectURL (C:\Users\stefe\starchart\node_modules\samlify\src\binding-redirect.ts:116:16)
>     at ServiceProvider.createLoginRequest (C:\Users\stefe\starchart\node_modules\samlify\src\entity-sp.ts:71:32)
>     at createLoginRequest (C:\Users\stefe\starchart\app\saml.server.ts:44:26)
>     at action2 (C:\Users\stefe\starchart\app\routes\login.tsx:28:27)
>     at processTicksAndRejections (node:internal/process/task_queues:95:5)
> ```
> 
> I'll take a look again tomorrow

That stack implies that for some reason the IDP's single sign on config is wrong, see https://github.com/tngan/samlify/blob/18ef1392eb92f9fcaa79d0fafe25e9d8d2fcb1cf/src/binding-redirect.ts#L93

```
const base = metadata.idp.getSingleSignOnService(binding.redirect);
```

sfrunza13:
I put the path to the xml file in config in the secret, when I should have probably just been putting the xml in the secret directly, that seems to do it.

sfrunza13:
Should I delete the metadata file in the config folder?

app/routes/login.tsx
```diff
@@ -11,17 +11,23 @@ import {
...
-    username: 'starchartdev',
-    remember: false,
-    redirectTo: '/',
-  });
+  // Check if a session with a username exists
+  const user = await getUsername(request);
+
+  // If not then create a login request to the IDP's redirect binding
+  if (!user) {
+    const samlRedirectURL = await createLoginRequest(new URL(request.url));
```
humphd, line 26: Question: what is `request.url` going to be here?

What about the case of `http://localhost:8080/login?redirectTo=%2Fdev`, where I want to go to `/dev` but get redirected to the login page?

app/routes/login/callback.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const formData = await request.formData();
+  const body = Object.fromEntries(formData);
+  const { attributes } = await parseLoginResponse(body);
+
+  // Try and extract the username and see if there is an existing user by that name
+  if (!attributes.sAMAccountName) {
+    // TODO: Make this redirect to access denied page
+    return redirect('/');
+  }
+  const returnTo: string = body.RelayState ? body.RelayState.toString() : '/';
```
humphd, line 58: Don't specify a type if you don't have to (seems odd I know).  TS can infer most types.  Only decorate with types when it can't.

app/saml.server.ts
```diff
@@ -0,0 +1,59 @@
...
+  metadata: SAML_IDP_METADATA,
+});
+
+export function metadata() {
+  return sp.getMetadata();
+}
+
+export async function createLoginRequest(url?: URL) {
+  const { context } = sp.createLoginRequest(idp, 'redirect');
+  const returnTo = url ? url.searchParams.get('redirectTo') : '/';
```
humphd, line 45: Nice

humphd (changes requested):

sfrunza13:
should I be casting relayState to string?

```
Type 'FormDataEntryValue' is not assignable to type 'string'.
  Type 'File' is not assignable to type 'string'.ts(2322)
```

the above occurs on 78 in login/callback.tsx

humphd:
> should I be casting relayState to string?
> 
> ```
> Type 'FormDataEntryValue' is not assignable to type 'string'.
>   Type 'File' is not assignable to type 'string'.ts(2322)
> ```
> 
> the above occurs on 78 in login/callback.tsx

What does `parseLoginResponse()` actually return?

All of the Samlify tests use this pattern:

```ts
  const { samlContent, extract } = await sp.parseLoginResponse(idpcustomNoEncrypt, 'redirect', parseRedirectUrlContextCallBack(context));
  t.is(typeof id, 'string');
  t.is(samlContent.startsWith('<samlp:Response'), true);
  t.is(samlContent.endsWith('/samlp:Response>'), true);
  t.is(extract.nameID, 'user@esaml2.com');
  t.is(extract.attributes.name, 'mynameinsp');
  t.is(extract.attributes.mail, 'myemailassociatedwithsp@sp.com');
  t.is(extract.response.inResponseTo, '_4606cc1f427fa981e6ffd653ee8d6972fc5ce398c4');
```

So is it on `extract`?  That is: `extract.RelayState` or `extract.relayState` (not sure) and `extract.attributes`?

sfrunza13:
> > should I be casting relayState to string?
> > ```
> > Type 'FormDataEntryValue' is not assignable to type 'string'.
> >   Type 'File' is not assignable to type 'string'.ts(2322)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > the above occurs on 78 in login/callback.tsx
> 
> What does `parseLoginResponse()` actually return?
> 
> All of the Samlify tests use this pattern:
> 
> ```ts
>   const { samlContent, extract } = await sp.parseLoginResponse(idpcustomNoEncrypt, 'redirect', parseRedirectUrlContextCallBack(context));
>   t.is(typeof id, 'string');
>   t.is(samlContent.startsWith('<samlp:Response'), true);
>   t.is(samlContent.endsWith('/samlp:Response>'), true);
>   t.is(extract.nameID, 'user@esaml2.com');
>   t.is(extract.attributes.name, 'mynameinsp');
>   t.is(extract.attributes.mail, 'myemailassociatedwithsp@sp.com');
>   t.is(extract.response.inResponseTo, '_4606cc1f427fa981e6ffd653ee8d6972fc5ce398c4');
> ```
> 
> So is it on `extract`? That is: `extract.RelayState` or `extract.relayState` (not sure) and `extract.attributes`?

I created a method that is named the same as the sp one, I take relayState out of the body. I don't think the relayState is returned by the sp method.

humphd:
OK. I'm not as deep into the changes you've made on this as you are, so you tell me where this is at and let's get this finished.

sfrunza13:
```export async function parseLoginResponse(body: { [k: string]: FormDataEntryValue }) {
  const { extract } = await sp.parseLoginResponse(idp, 'post', {
    body,
  });
  const relayState = body.RelayState;
  console.log(relayState);
  return { attributes: extract, relayState: relayState };
}
```
on line 54 of saml.server.ts

app/routes/login/callback.tsx
```diff
@@ -0,0 +1,80 @@
...
+    );
+  }
+
+  // Either way create a session
+  return createUserSession({
+    request: request,
+    username: username,
+    remember: false,
+    redirectTo: returnTo,
+  });
```
sfrunza13, line 79: @humphd line 78 here is needing it to be specifically string

humphd:
```ts
redirectTo: returnTo as string,
```

app/routes/login.tsx
```diff
@@ -11,17 +11,25 @@ import {
...
-    remember: false,
-    redirectTo: '/',
-  });
+  // Check if a session with a username exists
+  const user = await getUsername(request);
+
+  // If not then create a login request to the IDP's redirect binding
+  if (!user) {
+    const url = new URL(request.url);
+    const redirectTo = url.searchParams.get('redirectTo');
```
humphd, line 27:
```ts
const redirectTo = url.searchParams.get('redirectTo') ?? undefined;
```

This will give us `undefined` if we get back `null`, which is better for the `createLoginRequest` params.  I'll comment below on how to do it.

app/routes/login/callback.tsx
humphd: I assume this isn't an issue any more?

app/saml.server.ts
```diff
@@ -0,0 +1,60 @@
...
+  metadata: SAML_IDP_METADATA,
+});
+
+export function metadata() {
+  return sp.getMetadata();
+}
+
+export async function createLoginRequest(redirectUrl?: string | null) {
+  const { context } = sp.createLoginRequest(idp, 'redirect');
+  const returnTo = redirectUrl ? redirectUrl : '/';
```
humphd, line 45: `redirectUrl || '/'`

app/saml.server.ts
```diff
@@ -0,0 +1,60 @@
...
+  return context + '&RelayState=' + returnTo;
```
humphd, line 46: A better way to do this is to use the built-in `URL` constructor that the browser/node provide us (avoid hand-crafting complex strings like URLs, paths, etc as much as you can):

```ts
const url = new URL(context);
url.searchParams.append('RelayState', returnTo);
return url.href;
```

app/saml.server.ts
```diff
@@ -0,0 +1,60 @@
...
+  const { context } = sp.createLogoutRequest(idp, 'redirect', { nameID: user });
+  return context;
+}
+
+export async function parseLoginResponse(body: { [k: string]: FormDataEntryValue }) {
+  const { extract } = await sp.parseLoginResponse(idp, 'post', {
+    body,
+  });
+  const relayState = body.RelayState as string;
+  return { samlResponse: extract, relayState: relayState };
```
humphd, line 59: Don't repeat when key/value are the same:

```ts
return { samlResponse: extract, relayState };
```

app/saml.server.ts
```diff
@@ -0,0 +1,60 @@
...
+export async function createLogoutRequest(user: string) {
+  const { context } = sp.createLogoutRequest(idp, 'redirect', { nameID: user });
+  return context;
+}
+
+export async function parseLoginResponse(body: { [k: string]: FormDataEntryValue }) {
+  const { extract } = await sp.parseLoginResponse(idp, 'post', {
+    body,
+  });
+  const relayState = body.RelayState as string;
```
humphd, line 58: Could the response ever not have this?  Because we aren't creating it, and don't control it, it's harder to do what you're doing here and know it will work.

app/saml.server.ts
```diff
@@ -0,0 +1,60 @@
...
+// Take the metadata stood up by the IDP and use it as the metadata for our IDP object
+const idp = samlify.IdentityProvider({
+  metadata: SAML_IDP_METADATA,
+});
+
+export function metadata() {
+  return sp.getMetadata();
+}
+
+export async function createLoginRequest(redirectUrl?: string | null) {
```
humphd, line 43: You can avoid the `?` and `| null` dance here if you always pass `string` or `undefined` (which my change above will do).  Here you can set a default value:

```ts
export async function createLoginRequest(redirectUrl: string = '/') {
```

humphd (changes requested):

humphd:
You also have a conflict with `.env.example` on `main`.

app/saml.server.ts
sfrunza13: I found the following online 
>the relay state is opaque data that the sender (IdP or SP) sends to the partner site (SP or IdP) along with the SAML request.
The receiver must return this relay state along with the SAML response.

and I have seen it referenced this way before. It gives me the impression that whatever we send with the request as RelayState should always come back as well.

humphd: OK, let's go with this then.

humphd (approved):
Excellent. Thanks for all your work writing and evolving this.

sfrunza13:
> Excellent. Thanks for all your work writing and evolving this.

Thank you for the help, lets get one more review

SerpentBytes (approved):
I was able to test it successfully on my machine. Run into a small issue, and @sfrunza13 was able to help me figure out what I was doing wrong. 

Great work!

PR #242, opened 2023-02-18
Domains listing integration from server + delete UI

### LINKED ISSUE #197: Feat: Integrate Domains Table with Back end
This task could be done only when back end for domains is ready. 
It should include the following: 
- integrate domain extension
- integrate domain deletion (also need to add dialog that gonna confirm user's intention)

Myrfion:
## Overview
What this PR does: 
- it adds loading of the domain from the server side instead of showing mock domains
- it adds the UI for deleting domains (Alert Dialog) 

### Fetching domains
integrating fetching domains using standard `json` and `useLoaderData` I faced an issue:
**Remix serialize and deserializes data sent from the server, but it doesn't deserialize dates, so return them like strings.**

This is a known problem in Remix and I found the discussion about it: https://github.com/remix-run/remix/discussions/1910

I found the solution of using `remix-typedjson` the most simple and beautiful. Using that lib you need just:
- `json` -> `typedjson`
- `useLoaderData` -> `useTypedLoaderData`

Using that now we can freely reuse types from Prisma without writing any custom converters, serializers etc. 

 ### Alert dialog
 I also added a simple Alert dialogue which let user know that changes of deleting the domain are irreversible. Took the component from Chakra 
 https://chakra-ui.com/docs/components/alert-dialog
 
 ### Some screenshots 
image
image

closes #197

### DIFF
--- a/app/routes/__index/domains/index.tsx
+++ b/app/routes/__index/domains/index.tsx
@@ -1,22 +1,38 @@
 import { AddIcon } from '@chakra-ui/icons';
 import { Button, Container, Flex, Heading, Text } from '@chakra-ui/react';
+import type { LoaderArgs } from '@remix-run/node';
+import { Response } from '@remix-run/node';
 import { Link, useNavigate } from '@remix-run/react';
 import type { DomainsTableAction } from '~/components/domains-table';
 import DomainsTable from '~/components/domains-table';
-import DOMAINS_MOCK from '~/mocks/domains';
+import { getRecordsByUsername } from '~/models/record.server';
+import { getUser } from '~/session.server';
+import type { Record } from '@prisma/client';
+import { typedjson, useTypedLoaderData } from 'remix-typedjson';
+
+export const loader = async ({ request }: LoaderArgs) => {
+  const user = await getUser(request);
+  if (!user) {
+    throw new Response('No user', { status: 500 });
+  }
+
+  return typedjson(await getRecordsByUsername(user.username));
+};
 
 export default function DomainsIndexRoute() {
+  const domains = useTypedLoaderData<typeof loader>();
   const navigate = useNavigate();
 
-  function onDomainAction(domainID: number, action: DomainsTableAction) {
+  function onDomainAction(domain: Record, action: DomainsTableAction) {
     switch (action) {
       case 'EDIT':
-        navigate(domainID.toString());
+        navigate(domain.id.toString());
         break;
       case 'DELETE':
-        // TODO: implement delete functionaty
+        console.log('Delete: ', domain);
         break;
       case 'RENEW':
+        console.log('Renew: ', domain);
         // TODO: implement renew functionaty
         break;
     }
@@ -37,7 +53,7 @@ export default function DomainsIndexRoute() {
           <Button rightIcon={<AddIcon boxSize={3} />}>Create new domain</Button>
         </Link>
       </Flex>
-      <DomainsTable domains={DOMAINS_MOCK} onAction={onDomainAction} />
+      <DomainsTable domains={domains} onAction={onDomainAction} />
     </Container>
   );
 }
--- a/app/components/domains-delete-alert-dialog.tsx
+++ b/app/components/domains-delete-alert-dialog.tsx
@@ -1,0 +1,42 @@
+import {
+  AlertDialog,
+  Button,
+  AlertDialogOverlay,
+  AlertDialogContent,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogBody,
+} from '@chakra-ui/react';
+import React from 'react';
+
+interface DomainsDeleteAlertDialogProps {
+  isOpen: boolean;
+  onCancel: () => void;
+  onConfirm: () => void;
+}
+
+export default function DomainsDeleteAlertDialog(props: DomainsDeleteAlertDialogProps) {
+  const { isOpen, onCancel, onConfirm } = props;
+  const cancelRef = React.useRef(null);
+
+  return (
+    <AlertDialog isCentered isOpen={isOpen} onClose={onCancel} leastDestructiveRef={cancelRef}>
+      <AlertDialogOverlay>
+        <AlertDialogContent>
+          <AlertDialogHeader fontSize="lg" fontWeight="bold">
+            Delete Domain
+          </AlertDialogHeader>
+          <AlertDialogBody>Are you sure? You can't undo this action afterwards.</AlertDialogBody>
+          <AlertDialogFooter>
+            <Button colorScheme="gray" onClick={onCancel} ref={cancelRef}>
+              Cancel
+            </Button>
+            <Button colorScheme="brand" onClick={onConfirm} ml="3">
+              Delete
+            </Button>
+          </AlertDialogFooter>
+        </AlertDialogContent>
+      </AlertDialogOverlay>
+    </AlertDialog>
+  );
+}
--- a/app/components/domains-table.tsx
+++ b/app/components/domains-table.tsx
@@ -11,6 +11,7 @@ import {
   Tooltip,
   Card,
   useToast,
+  useDisclosure,
 } from '@chakra-ui/react';
 import type { Record, RecordStatus } from '@prisma/client';
 import {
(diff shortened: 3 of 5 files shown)


Myrfion:
Also I am not sure how I suppose to handle the case when `getUser` returns me nothing, for now I put some message and 500 status code, but I have a feeling that it probably should be different

app/routes/__index/domains/index.tsx
```diff
@@ -1,22 +1,38 @@
...
-import DOMAINS_MOCK from '~/mocks/domains';
+import { getRecordsByUsername } from '~/models/record.server';
+import { getUser } from '~/session.server';
+import type { Record } from '@prisma/client';
+import { typedjson, useTypedLoaderData } from 'remix-typedjson';
+
+export const loader = async ({ request }: LoaderArgs) => {
+  const user = await getUser(request);
+  if (!user) {
+    throw new Response('No user', { status: 500 });
```
Myrfion, line 16: @humphd not sure how I suppose to handle this case, just put it like that for now, I am curious what you think

humphd: Don't use `getUser`, use https://github.com/DevelopingSpace/starchart/blob/main/app/session.server.ts#L48-L58 which will handle the redirect case if it's missing.

app/routes/__index/domains/index.tsx
```diff
@@ -1,22 +1,38 @@
...
+  }
+
+  return typedjson(await getRecordsByUsername(user.username));
```
humphd, line 19: Doesn't Remix [already handle the typing](https://remix.run/docs/en/v1/guides/data-loading#basics) on loader data via `typeof loader`?

app/routes/__index/domains/index.tsx
```diff
@@ -1,22 +1,38 @@
...
+import type { LoaderArgs } from '@remix-run/node';
+import { Response } from '@remix-run/node';
 import { Link, useNavigate } from '@remix-run/react';
 import type { DomainsTableAction } from '~/components/domains-table';
 import DomainsTable from '~/components/domains-table';
-import DOMAINS_MOCK from '~/mocks/domains';
+import { getRecordsByUsername } from '~/models/record.server';
+import { getUser } from '~/session.server';
+import type { Record } from '@prisma/client';
+import { typedjson, useTypedLoaderData } from 'remix-typedjson';
```
humphd, line 11: nit:  Separate your imports such that you do third-party, internal, types:

```ts
import foo from 'foo';
...

import { getUser } from '~/session.server';
...

...
```

This makes it a lot easier to understand what's happening.

humphd (changes requested):
This PR is pretty big to submit the day before a release.  You need to get work in with time to review and iterate.

dadolhay (changes requested):
The terminology is incorrect, Imo these should be called records

app/routes/__index/domains/index.tsx
Myrfion: I described this problem in the PR description, the problem that remix has is it doesn't deserialize dates, so it just returns them like strings in run-time

Myrfion:
> The terminology is incorrect, Imo these should be called records

@dadolhay I agree that we should have consistent terminology across the whole project but the problem is that it is not a thing now even in the server code. I tried to use the exclusive term `Domain` because this is how it is called on the user side. Maybe you know some good practices on how to shift to consistent terminology for the whole project once and forever

humphd:
I filed #245 to deal with this throughout the code.  We need to fix it everywhere.

dadolhay:
> > The terminology is incorrect, Imo these should be called records
> 
> @dadolhay I agree that we should have consistent terminology across the whole project but the problem is that it is not a thing now even in the server code. I tried to use the exclusive term `Domain` because this is how it is called on the user side. Maybe you know some good practices on how to shift to consistent terminology for the whole project once and forever

Whereever I worked in the past, we always went with "rip the bandaid off" in these kinds of situations. The longer we are letting the wrong term be in the system, the more confusion and incorrect use it will generate as we are adding additional code. I'd suggest a search and replace, go through the whole app, and correct all incorrect filenames, and uses at once. Some devs will have a bad time merging the conflicts, but it's still better than letting it spread. At least this is my opinion.

app/routes/__index/domains/index.tsx
Myrfion: @humphd what do you think about this? Do you think my justification for using this lib is fair compared to regular remix stuff?

humphd:
> > > The terminology is incorrect, Imo these should be called records
> > 
> > 
> > @dadolhay I agree that we should have consistent terminology across the whole project but the problem is that it is not a thing now even in the server code. I tried to use the exclusive term `Domain` because this is how it is called on the user side. Maybe you know some good practices on how to shift to consistent terminology for the whole project once and forever
> 
> Whereever I worked in the past, we always went with "rip the bandaid off" in these kinds of situations. The longer we are letting the wrong term be in the system, the more confusion and incorrect use it will generate as we are adding additional code. I'd suggest a search and replace, go through the whole app, and correct all incorrect filenames, and uses at once. Some devs will have a bad time merging the conflicts, but it's still better than letting it spread. At least this is my opinion.

I agree, let's fix it all at once, and do it in #245 across the whole code base.  Don't worry about it in this PR.

app/components/record-delete-alert-dialog.tsx
```diff
@@ -0,0 +1,42 @@
+import {
+  AlertDialog,
+  Button,
+  AlertDialogOverlay,
+  AlertDialogContent,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogBody,
+} from '@chakra-ui/react';
+import React from 'react';
```
humphd, line 10: `import { useRef } from 'react';`

app/routes/__index/domains/index.tsx
```diff
@@ -1,22 +1,34 @@
...
-  function onDomainAction(domainID: number, action: DomainsTableAction) {
+  function onDomainAction(domain: Record, action: DomainsTableAction) {
     switch (action) {
       case 'EDIT':
-        navigate(domainID.toString());
+        navigate(domain.id.toString());
         break;
       case 'DELETE':
-        // TODO: implement delete functionaty
+        console.log('Delete: ', domain);
```
humphd, line 28: Let's file issues now to implement these please, and not use `console` in here, or we'll have eslint warnings.

app/routes/__index/domains/index.tsx
```diff
@@ -1,22 +1,34 @@
...
         break;
       case 'RENEW':
+        console.log('Renew: ', domain);
```
humphd, line 31: Same with this, please file an issue to implement this.

app/routes/__index/domains/index.tsx
```diff
@@ -1,22 +1,34 @@
 import { AddIcon } from '@chakra-ui/icons';
 import { Button, Container, Flex, Heading, Text } from '@chakra-ui/react';
+import type { LoaderArgs } from '@remix-run/node';
 import { Link, useNavigate } from '@remix-run/react';
+import { typedjson, useTypedLoaderData } from 'remix-typedjson';
 import type { DomainsTableAction } from '~/components/domains-table';
 import DomainsTable from '~/components/domains-table';
-import DOMAINS_MOCK from '~/mocks/domains';
```
humphd, line 6: Can we delete this file now?

humphd (changes requested):

app/routes/__index/domains/index.tsx
Myrfion: #271

app/routes/__index/domains/index.tsx
Myrfion: #272

humphd (approved):

app/routes/__index/domains/index.tsx
```diff
@@ -1,17 +1,28 @@
...
+import { getRecordsByUsername } from '~/models/record.server';
+import { requireUsername } from '~/session.server';
+import type { Record } from '@prisma/client';
+
+export const loader = async ({ request }: LoaderArgs) => {
+  const username = await requireUsername(request);
+
+  return typedjson(await getRecordsByUsername(username));
```
SerpentBytes, line 15: So `typedjson` is sort of like `JSON.parse()` but for TypeScript?

SerpentBytes (approved):

app/routes/__index/domains/index.tsx
Myrfion: I would say more like `json` function from remix but with better support of types

Myrfion:
Hi @dadolhay, since we decided to address your changes request in another PR and in this one I changed the terminology for current changes, could you please resolve your changes request so I could merge the PR in

PR #243, opened 2023-02-18
Implement domain creation page UI

### LINKED ISSUE #48: Add a domain creation page
- This page should include a form that allows users to create a new domain. The user domain management page would link to this page (see #46).

- The page should be build using [Chakra UI components](https://chakra-ui.com/docs/components) as much as possible.

- The page should be build with responsive design in mind, so it can scale to different screen sizes.

- The page should use minimal css of its own, and use the theme (see #34) as much as possible. This would reduce the amount of manual work for dark mode.

This should be done after a design has been created as per #36.

Eakam1007:
Resolves #48 

- Implemented `domain/dns record` creation form UI based on [figma](https://www.figma.com/file/XAcnMIC1y4qumWinSalHI9/Starchart-UI?node-id=2175%3A3583&t=oM3dYngtGD0QSrZf-0) design
- Added a simple action to create a `domain/dns record` and redirect to domain edit page. This is simply for testing purposes and would be replaced in the future
- Added error checking for `Type` field. If any value other than the ones allowed is sent, it will show an error. (Note that to do this, you would need to manually edit the html) Any other validations can be added following this pattern

> Domain should be replaced with DNS Record and will be fully handled in #245 

image

### Steps to test
- Run app locally
- Run `npm run db:studio` at the same time to see the `DNS Records/Domains` being created
- Navigate to `/domains/new`
- Check form style
- Check tooltips text
- Ensure `Domain name`, `Type`, and `Value` fields are required
- Try entering different values and check that `DNS Records/Domains` are created for valid data


app/components/domain-form.tsx
```diff
@@ -0,0 +1,51 @@
+import { AddIcon, InfoIcon } from '@chakra-ui/icons';
+import { Button, HStack, Input, Select, Textarea, Tooltip, VStack } from '@chakra-ui/react';
+import { Form } from '@remix-run/react';
+
+export default function DomainForm() {
```
Ririio, line 5: If I recall from one of the previous discussion we'll have the container centralized, but have its content left aligned?

app/components/domain-form.tsx
```diff
@@ -0,0 +1,51 @@
...
+        </HStack>
+        <HStack w="100%">
+          <Input placeholder="Course" name="course" />
+          <Tooltip>
+            <InfoIcon></InfoIcon>
+          </Tooltip>
+        </HStack>
+        <Textarea placeholder="Description" rows={10} name="description" />
+      </VStack>
+      <Button type="submit" mt="6" rightIcon={<AddIcon boxSize={3.5} mt="0.15rem" />}>
```
Ririio, line 46: We probably should avoid hardcoding our spacing, and use the default that chakra provides

app/components/domain-form.tsx
```diff
@@ -0,0 +1,51 @@
+import { AddIcon, InfoIcon } from '@chakra-ui/icons';
+import { Button, HStack, Input, Select, Textarea, Tooltip, VStack } from '@chakra-ui/react';
+import { Form } from '@remix-run/react';
+
+export default function DomainForm() {
+  return (
+    <Form className="domain-form" method="post">
+      <VStack maxW="xl" spacing="5">
+        <HStack w="100%">
+          <Input placeholder="Domain Name" name="name" />
```
humphd, line 10: Every form element needs a label.  You can't use the `placeholder` as a label because as soon as the user starts typing, they lose the info about what the field represents.  You can use the `placeholder` to give an example of valid input.

app/components/domain-form.tsx
```diff
@@ -0,0 +1,51 @@
...
+          <Tooltip>
+            <InfoIcon></InfoIcon>
```
humphd, line 12: These need text to be useful.

app/routes/__index/domains/new.tsx
```diff
@@ -1,11 +1,52 @@
...
+import DomainForm from '~/components/domain-form';
+import { createRecord } from '~/models/record.server';
+import { getUsername } from '~/session.server';
+
+import type { ActionArgs } from '@remix-run/node';
+import { RecordType } from '@prisma/client';
+
+export const action = async ({ request }: ActionArgs) => {
+  const form = await request.formData();
+  const name = form.get('name')?.toString();
```
humphd, line 13: I would use [zod](https://www.npmjs.com/package/zod) for this.

humphd (changes requested):

app/components/domain-form.tsx
Eakam1007: I did not see any chakra spacing that was small enough for this, so I went with a hardcoded value

app/components/domain-form.tsx
Eakam1007: I think the idea was to leave a margin on the left but no margin on mobile. I will try to make it work with your suggestion

app/components/domain-form.tsx
Eakam1007: Hmmm, the design does not have labels but only uses placeholders. Should I add labels at the top of the fields? Any thoughts @Myrfion @Ririio ?

app/components/domain-form.tsx
Eakam1007: I am not sure what text to display here. Do we have something or should I add text based on what I understand? @Myrfion @Ririio

app/components/domain-form.tsx
humphd: The design is wrong.  Decades of work on usability of forms says that you need to have a label for each form element.  There are a few reasons:

1. on mobile, the label is `for="id-of-input"` which increases the touch area (i.e., touch the label or input and get focus)
2. placeholder-labels disappear as soon as there is text entered.  If a user has multiple fields like this, and an error "incorrect input" they have no idea which field is the wrong one.  You need to make the form navigable after it has data included

app/components/domain-form.tsx
Ririio: You can add the text based on your understanding, and we can take a look at it and see if it's correct

dadolhay (changes requested):
The terminology is incorrect, IMO these hould be called records:

A record *contain* a domain name, a type, value etc.

Eakam1007:
> The terminology is incorrect, IMO these hould be called records:
> 
> A record _contain_ a domain name, a type, value etc.

Yes, it should be record. But should we change the text that is shown to the user to `Records` or only in code? Relevant issue #245

app/components/domain-form.tsx
Eakam1007: Yeah, I will add the labels at the top of the fields for now

Myrfion: Agree with @humphd  on that, we need labels to be labels, not placeholders

app/components/domain-form.tsx
Myrfion, line 12: If some element doesn't have any child (so nothing inside of it) we should just leave the as selfclosable tag, so
```typescript
<InfoIcon />
```

app/components/domain-form.tsx
Myrfion: https://chakra-ui.com/docs/styled-system/theme#spacing

Check it out it has `0.5` spacing

Myrfion (dismissed):

dadolhay:
> > The terminology is incorrect, IMO these hould be called records:
> > A record _contain_ a domain name, a type, value etc.
> 
> Yes, it should be record. But should we change the text that is shown to the users to `Records` or only change in code and keep showing `Domains` to the users? Relevant issue #245

I think we should keep proper terminology. If the user knows about the record type (a, aaaa, cname, etc.) they should know about it belonging to a record too.

app/theme/index.ts
```diff
@@ -1,6 +1,22 @@
...
-const theme = extendTheme({ colors }, withDefaultColorScheme({ colorScheme: 'brand' }));
+const theme = extendTheme(
+  {
+    colors,
+    styles: {
+      global: {
+        '.domain-form': {
+          // InfoIcon
+          '.chakra-icon': {
+            color: '#d9d9d9',
```
SerpentBytes, line 12: I would use the colour value defined value in `./colors` here.

Eakam1007: I decided to not add another value to that file because from what I understand, I would need to define a new property just for one color value

SerpentBytes: I just pointed that out because, let's say, we have to change colours to a lighter/darker shade later on; we could update values defined in the `./colors` file, which should be enough. In such cases, we could easily miss updating the value of this icon if we are not careful enough. However, if we are confident we would never be required to change the colour scheme to a darker/lighter shade, then hardcoding here is fine.

app/components/domain-form.tsx
Eakam1007: With `0.5` spacing, that 0.025 rem difference makes the plus sign slightly off center

humphd:
@Eakam1007 let me know when you want this reviewed again.  I see you have a bunch of new commits, but tests not passing. I'll wait to hear from you.

app/components/domain-form.tsx
Eakam1007: Hmmm, I could add `Center` but it doesn't affect the page at all. I have added a margin to the left for certain screen sizes. For smaller screens, the margin disappears so I don't think it is worth adding a `Center` here anymore.

app/theme/index.ts
Eakam1007: So, the issue with using one of the existing colors is that, none of the colors in the `./colors` file match this color. Should I add another object for gray colors?
```ts
const colors = {
  brand: {
    50: '#ffe4e3',
    100: '#feb7b6',
    ...
  },
  gray: {
    ...
  }
};
```

SerpentBytes: @Myrfion @Ririio What would you guys suggest?

Ririio: I would add the default values and add a new instance for the gray color. Maybe like gray.350, should follow the format though from light to darkest

Eakam1007: Should I add an entire color palate in this case based around the light grayish color of the icon?

Ririio: I think that's how it works

Eakam1007: I could not get it to work with `gray` as it overrides existing colors for a lot of components. So, I added a different name: `brand_gray`. If this looks weird, we can revert to manually overriding the color value for this one icon only

Eakam1007:
For tests, this shows up after clicking the sign in button. This started happening after rebasing with main, so removing the tests for now. I will create another issue for this

image

app/models/record.server.ts
```diff
@@ -9,34 +9,44 @@ export async function getRecordsByUsername(username: Record['username']) {
...
-      course,
-      ports,
-      expiresAt,
-      status,
-    },
-  });
+  return prisma.record.create({ data: { ...data, expiresAt } });
 }
 
+// export async function createRecord(
```
dadolhay, line 21: Please remove dead code. Git never forgets, so there is no need to keep it here
I suggest using the "Git Lens" plugin in VsCode if you have not encountered it yet

dadolhay (changes requested):

app/models/record.server.ts
Eakam1007: Oh, sorry about that. Seems like I missed this

humphd:
> In the testing environment, this shows up after clicking the sign in button. This started happening after rebasing with main, so removing the tests for now. I will create another issue for this
> 

I'm pretty sure this means that you need to bring down your Docker containers and restart them, so that the metadata config gets included:

```ts
docker compose down
docker compose up -d
```

Eakam1007:
> I'm pretty sure this means that you need to bring down your Docker containers and restart them, so that the metadata config gets included:
> 
> ```ts
> docker compose down
> docker compose up -d
> ```

Unfortunately not, I even tried deleting build and running `npm run build` again alongside deleting all the containers and creating them again. Also, using `127.0.0.1` instead of `localhost` in the URL does not seem to work either

humphd:
> > I'm pretty sure this means that you need to bring down your Docker containers and restart them, so that the metadata config gets included:
> > ```ts
> > docker compose down
> > docker compose up -d
> > ```
> 
> Unfortunately not, I even tried deleting build and running `npm run build` again alongside deleting all the containers and creating them again. Also, using `127.0.0.1` instead of `localhost` in the URL does not seem to work either

cc @sfrunza13

app/components/dns-record/form.tsx
```diff
@@ -0,0 +1,63 @@
+import { AddIcon, InfoIcon } from '@chakra-ui/icons';
+import { Button, Input, Select, Textarea, Tooltip, VStack } from '@chakra-ui/react';
+import { Form } from '@remix-run/react';
+import FormField from './form-field';
+
+interface FormParams {
```
humphd, line 6: `DomainFormProps` would be the more usual name for this

app/components/dns-record/form.tsx
```diff
@@ -0,0 +1,63 @@
...
+  typeError?: string; // Error for 'Type' field
+}
+
+export default function DomainForm({ typeError }: FormParams) {
+  return (
+    <Form className="domain-form" method="post">
+      <VStack maxW="xl" spacing="2">
+        <FormField label="Domain Name" isRequired={true}>
+          <Input name="name" />
```
humphd, line 15: In a follow-up, let's use an [`<InputRightAddon>`](https://chakra-ui.com/docs/components/input#left-and-right-addons) to show the right-hand parts of the domain (i.e., we only want `foo` in `foo.username.domain.com`, so show `{username}.domain.com` on the right.)

app/routes/__index/domains/new.tsx
```diff
@@ -1,11 +1,66 @@
...
+  return error.issues.find((issue) => issue.path[0] === field)?.message;
+}
+
+export const action = async ({ request }: ActionArgs) => {
+  // Create a Zod schema for validation
+  // Optional is not needed as we get '' if nothing is entered
+  const DnsRecord = z.object({
+    name: z.string().min(1), // We do not want to consider '' a valid string
+    type: z.nativeEnum(RecordType),
+    value: z.string().min(1),
```
humphd, line 25: In a follow-up, I bet we could do this with a function and some regex to check for different things based on the `type`.

app/routes/__index/domains/new.tsx
```diff
@@ -1,11 +1,66 @@
...
+    ports: z.string(),
```
humphd, line 26: These three all need `.optional()` I think, see https://github.com/colinhacks/zod#optionals

app/routes/__index/domains/new.tsx
```diff
@@ -1,11 +1,66 @@
...
+    course: z.string(),
+    description: z.string(),
+  });
+
+  const newDnsRecordParams = await parseFormSafe(request, DnsRecord);
+  const username = await requireUsername(request);
```
humphd, line 32: I would do this first, since we shouldn't be here if the user isn't logged in.

app/routes/__index/domains/new.tsx
```diff
@@ -1,11 +1,66 @@
...
+
+  // If validations failed, we return the errors to show on the form
+  // Currently only returns 'type' field errors as no other validations exist
+  // Also, form cannot be submitted without required values
+  if (newDnsRecordParams.success === false) {
+    return json({
```
humphd, line 38: Do we need to use `typedjson` here?  cc @Myrfion

app/routes/__index/domains/new.tsx
```diff
@@ -1,11 +1,66 @@
...
+      errors: {
+        typeError: errorForField(newDnsRecordParams.error, 'type'),
+      },
+    });
+  }
+
+  // Create a pending record... for now
+  // This will most likely be replaced by some other logic
```
humphd, line 46: fyi @Genne23v

humphd (changes requested):
Looks good, I ran it locally.  I've left some comments and a few things to fix.  There are some things we can improve in follow-ups.  Nice work!

app/components/dns-record/form.tsx
Eakam1007: I didn't go for `DomainFormProps` cause it should actually be `DnsRecordFormProps` and that felt a bit long. I could change it if the length doesn't matter

humphd: It should go with your component, so name it `*Props`

app/routes/__index/domains/new.tsx
Myrfion: As far as I understand no dates or other types that might require `typedjson` are returned in response, so I don't see a necessity for it there, just only if you think we should use it everyevere

humphd: I'd suggest using the same approach throughout the code.

app/components/dns-record/form.tsx
Eakam1007: To be clear, do you want it to be `DnsRecordFormProps` or just `*Props`, or `FormProps` ?

humphd: Whatever the name of the component actually is, so if it's `Form`, then `FormProps`.  But your component is `DnsRecordForm` right?  So `DnsRecordFormProps`.

Eakam1007: Alright, thanks for the clarification!

app/routes/__index/domains/new.tsx
Eakam1007: When I was testing this, the props would resolve to `''` ,aka an empty string, if the field was blank on the form. This is why I removed the `optional()` cause they would never not be present

humphd: OK, sounds good.

app/routes/__index/domains/new.tsx
Eakam1007: Created #277 for this and the [other comment](https://github.com/DevelopingSpace/starchart/pull/243#discussion_r1117072649) about using `<InputRightAddon>`

humphd (approved):
Great stuff

dadolhay (approved):

dadolhay:
@Eakam1007 Please don't forget to merge this if all is done

PR #254, opened 2023-02-21
Use Docker Secrets in dev and production

humphd:
This is part of #43.  I need to be able to set all the Docker Secrets for the container.  In development, we can create any secrets we want in `dev-secrets/*`.  The filename is the name of the secret, and the contents of the file is its value.

I've rewritten the code to use a new `~lib/secrets.server` module, which does the right thing in dev and production.

We probably need to move more things out of `env` and into `secrets`, so help me find them.

NOTE: the database URL has to be in both places, due to how Prisma works with the command line tooling.


dadolhay (approved):

shawnyu5:
Does our E2E tests need to be modified as well?

humphd:
> Does our E2E tests need to be modified as well?

yeah, just debugging that now.  There are 2 issues: 1) we can't use `NODE_ENV=production` here; 2) we're running from within `./build` in e2e tests, so I need to get the path levels right for loading the secrets, or copy them over into the build dir.

Still trying to get it right.

SerpentBytes (dismissed):

app/lib/dns.server.ts
```diff
@@ -21,14 +23,34 @@ if (process.env.NODE_ENV === 'production') {
...
+    return {
+      accessKeyId: AWS_ACCESS_KEY_ID,
+      secretAccessKey: AWS_SECRET_ACCESS_KEY,
+    };
+  }
+
+  return {
+    accessKeyId: process.env.AWS_ACCESS_KEY_ID || AWS_ACCESS_KEY_ID,
+    secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY || AWS_SECRET_ACCESS_KEY,
+    sessionToken: process.env.AWS_SESSION_TOKEN,
```
Genne23v, line 46: For my better understanding, these credentials are used in `staging` as we don't have any credentials for DNS in `dev` environment.

app/lib/lets-encrypt.server.ts
```diff
@@ -173,21 +175,20 @@ class LetsEncrypt {
...
+
+    if (process.env.NODE_ENV === 'production') {
       if (!process.env.LETS_ENCRYPT_ACCOUNT_EMAIL) {
         throw new Error('The env LETS_ENCRYPT_ACCOUNT_EMAIL is missing');
       }
 
-      this.#accountKey = LETS_ENCRYPT_ACCOUNT_PRIVATE_KEY_PEM;
       this.#directoryUrl = acme.directory.letsencrypt.production;
     } else {
       // For testing and local development, let's use an ad-hoc generated key
```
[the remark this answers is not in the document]
humphd: Done.

app/lib/dns.server.ts
humphd: This gives the ability to provide them if you want (e.g., if you want to connect to real AWS locally in dev).

humphd:
I've updated this so it passes tests.  It wasn't easy to figure out the right way to do it, but I think I've got a solid solution now that works in all the various ways we run the code (bundled vs. unbundled, vitest vs. e2e vs. dev vs. stand-alone).  Let me know if any of this doesn't make sense.

Genne23v (approved):
It looks good to me

humphd:
I'm going to land this, but fyi @dadolhay and @sfrunza13, this will affect you both in PRs you have open now.

PR #256, opened 2023-02-21
Making Certificate UI responsive

### LINKED ISSUE #252: Certificate UI responsiveness
We have to ensure that the UI for certificate is responsive to accommodate for a myriad of screen resolutions

Ririio:
I added an `xs` breakpoint, because there are far too much space in between `base` and `sm` around `30em`, and some devices are between those which did cause an issue when trying to scale down some of the UI.

Here's a demo of the UI **(Updated)**:

https://user-images.githubusercontent.com/32577022/220522777-9882826a-65f5-4e45-8e05-f852c7eadfc3.mp4

If you noticed having some white spaces in the header, unfortunately unless I hard coded specifically to accommodate for that specific resolution, that white spot can occur. On the other hand, the UI should work perfectly with most devices

Close #252

### DIFF
--- a/app/components/certificate/certificate-available.tsx
+++ b/app/components/certificate/certificate-available.tsx
@@ -4,15 +4,15 @@ import Description from './description';
 import CertificateDisplay from './certificate-display';
 
 interface CertificateAvailableProps {
-  validFrom: Date;
-  validTo: Date;
+  validFromFormatted: string;
+  validToFormatted: string;
   publicKey: string;
   privateKey: string;
 }
 
 export default function CertificateAvailable({
-  validFrom,
-  validTo,
+  validFromFormatted,
+  validToFormatted,
   publicKey,
   privateKey,
 }: CertificateAvailableProps) {
@@ -21,8 +21,8 @@ export default function CertificateAvailable({
       <Description
         description="Certificate: Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s"
         certRequested={true}
-        validFrom={validFrom}
-        validTo={validTo}
+        validFromFormatted={validFromFormatted}
+        validToFormatted={validToFormatted}
       />
       <Divider />
       <CertificateDisplay
--- a/app/components/certificate/certificate-display.tsx
+++ b/app/components/certificate/certificate-display.tsx
@@ -34,66 +34,71 @@ export default function CertificateDisplay({ title, description, value }: Certif
   }
 
   return (
-    <Flex flexDirection="column" gap="5">
-      <Flex flexDirection="column" gap="4">
-        <HStack gap="4">
-          <Heading as="h4" size="sm">
-            {title}
-          </Heading>
-          <Tooltip label={`Copy ${title}`}>
-            <IconButton
-              backgroundColor="transparent"
-              color="black"
-              size="xs"
-              _hover={{
-                background: 'whitesmoke',
-                color: 'teal.500',
-              }}
-              aria-label="Copy to Clipboard"
-              icon={<CopyIcon fontSize="md" />}
-              onClick={() => onCopy()}
-            />
-          </Tooltip>
-          <Tooltip label={`Download ${title}`}>
-            <IconButton
-              backgroundColor="transparent"
-              color="black"
-              size="xs"
-              _hover={{
-                background: 'brand.500',
-                color: 'white',
-              }}
-              aria-label="Download"
-              icon={
-                <DownloadIcon
-                  fontSize="md"
-                  onClick={() =>
-                    toast({
-                      title: `Downloading ${title}`,
-                      position: 'bottom-right',
-                      status: 'success',
-                    })
-                  }
-                />
-              }
-            />
-          </Tooltip>
-        </HStack>
-        <Text>{description}</Text>
-        <Accordion allowMultiple>
-          <AccordionItem>
-            <AccordionButton>
-              <Box as="span" flex="1" textAlign="left">
-                Show/Hide
-              </Box>
-              <AccordionIcon />
-            </AccordionButton>
-            <AccordionPanel>
-              <Text>{value}</Text>
-            </AccordionPanel>
-          </AccordionItem>
-        </Accordion>
-      </Flex>
+    <Flex
+      flexDirection="column"
+      gap="4"
+      textAlign={{ base: 'center', md: 'left' }}
+      fontSize={{ base: 'sm', md: 'md' }}
+    >
+      <HStack gap="4" alignSelf={{ base: 'center', md: 'flex-start' }}>
+        <Heading as="h4" size="sm">
+          {title}
+        </Heading>
+        <Tooltip label={`Copy ${title}`}>
+          <IconButton
+            backgroundColor="transparent"
+            color="black"
+            size="xs"
+            _hover={{
+              background: 'whitesmoke',
+              color: 'teal.500',
+            }}
+            aria-label="Copy to Clipboard"
+            icon={<CopyIcon fontSize="md" />}
+            onClick={() => onCopy()}
+          />
+        </Tooltip>
+        <Tooltip label={`${title} is Downloaded`}>
+          <IconButton
+            backgroundColor="transparent"
+            color="black"
+            size="xs"
+            _hover={{
+              background: 'brand.500',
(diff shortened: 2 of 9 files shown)


Genne23v:
@Ririio Probably separate issue needs to be open. I think the header content is very condensed in iPad size viewports. I can write an issue, if it is a valid improvement. 

Screenshot 2023-02-21 at 9 48 43 PM

Genne23v:
I think date can be `MMM DD YYYY` format without `day`.

Genne23v:
This is just my opinion. To me, it's a bit weird that `Created On` is not left-aligned while other contents are all left-aligned. 

Screenshot 2023-02-21 at 9 55 31 PM

app/components/certificate/certificate-display.tsx
```diff
@@ -34,66 +34,71 @@ export default function CertificateDisplay({ title, description, value }: Certif
...
+              background: 'brand.500',
+              color: 'white',
+            }}
+            aria-label="Download"
+            icon={
+              <DownloadIcon
+                fontSize="md"
+                onClick={() =>
+                  toast({
+                    title: `Downloading ${title}`,
```
Genne23v, line 76: Download will be less than a second. Perhaps `Download` can avoid potential misunderstanding.

app/components/certificate/certificate-request.tsx
```diff
@@ -16,24 +16,38 @@ export default function CertificateRequestView({
...
         <Text>Lorem Ipsum is simply dummy text of the printing and typesetting industry</Text>
-
-        <HStack>
-          <Text fontWeight="bold">Domain Name:</Text> <Text>{subject}</Text>
-        </HStack>
+        <Flex
+          flexDirection={{ base: 'column', md: 'row' }}
+          alignItems={{ base: 'center', md: 'flex-start' }}
+        >
+          <Text fontWeight="bold">Domain Name:&nbsp;</Text>
```
Genne23v, line 46: To me, it's a bit strange that many `Domain Name` are in different sizes. Perhaps it could be better to use a table or bullets without title? 

Screenshot 2023-02-21 at 10 08 44 PM

Genne23v (changes requested):

app/components/certificate/certificate-request.tsx
Ririio: I'm actually a bit confused about this. If we're going to display the `*` is there really a need to show all domains?

app/components/certificate/certificate-display.tsx
Ririio: sure that make sense

app/components/certificate/certificate-available.tsx
```diff
@@ -1,4 +1,4 @@
-import { Divider } from '@chakra-ui/react';
+import { Divider, Spacer } from '@chakra-ui/react';
```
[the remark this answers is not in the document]
Ririio: > This is just my opinion. To me, it's a bit weird that `Created On` is not left-aligned while other contents are all left-aligned.

It's more for the way the dates are aligned, because we are not displaying each date as a row, it's a bit odd for me to not make them centered, but if you have a different approach I'll be glad to take a look and change it

Ririio:
> This is just my opinion. To me, it's a bit weird that `Created On` is not left-aligned while other contents are all left-aligned.

It's more for the way the dates are aligned, because we are not displaying each date as a row, it's a bit odd for me to not make them centered, but if you have a different approach I'll be glad to take a look and change it

app/components/certificate/certificate-request.tsx
```diff
@@ -1,4 +1,4 @@
-import { Flex, Heading, Text, HStack, Button } from '@chakra-ui/react';
+import { Flex, Heading, Text, Button, VStack } from '@chakra-ui/react';
```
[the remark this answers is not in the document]
Ririio: > @Ririio Probably separate issue needs to be open. I think the header content is very condensed in iPad size viewports. I can write an issue, if it is a valid improvement.
> 

No it's all good, I can change them on my next commit

app/components/certificate/certificate-request.tsx
Genne23v: I'm actually struggling to understand too. Probably worth discussing in a meeting.

Ririio:
> @Ririio Probably separate issue needs to be open. I think the header content is very condensed in iPad size viewports. I can write an issue, if it is a valid improvement.
> 

No it's all good, I can change them on my next commit

humphd:
@Myrfion and @Eakam1007, can you two pls review this?

app/components/certificate/certificate-request.tsx
dadolhay: For the certificate there should only be one domain shown, the root domain like
`*.username.senecastudent.ca`

app/components/certificate/certificate-available.tsx
```diff
@@ -4,8 +4,8 @@ import Description from './description';
 import CertificateDisplay from './certificate-display';
 
 interface CertificateAvailableProps {
-  validFrom: Date;
-  validTo: Date;
+  validFrom: string;
```
Myrfion, line 7: Since in other places we use dates as dates why you decided to change them to strings here?

app/routes/__index/certificate/index.tsx
```diff
@@ -35,19 +35,29 @@ export default function CertificateIndexRoute() {
     }, 5000);
   }
 
+  function dateFormat(val: Date): string {
+    let date = val.toLocaleDateString('en-US', {
```
Myrfion, line 39: value of the `date` is not immutable so no need to declare it as `let` + you could just write it simpler:
```typescript
function dateFormat(val: Date): string {
return val.toLocaleDateString('en-US', {
month: 'short',
      day: '2-digit',
      year: 'numeric',
    });
}
```

I also think that the function name should be something like `formatDate` or `getFormatedDate`, `dateFormat` sounds like a variable name to me

app/theme/breakpoints.ts
```diff
@@ -0,0 +1,10 @@
+const breakpoints = {
```
Myrfion, line 1: Maybe I ask you why you decided to add these specific breakpoints and what is wrong with Chakra's default breakpoints?

Myrfion (changes requested):

app/routes/__index/certificate/index.tsx
Ririio: yeah you're right, I like `formatDate` better

app/theme/breakpoints.ts
Ririio: The `base` and `small` has a `30em` difference. I noticed that the difference was too big, and I need a breakpoint that is in between them. I added the `xs` just for that

app/components/certificate/certificate-request.tsx
Genne23v: I was actually talking about repeated same titles. I don't know what content will be under big title, but it seems strange to have the same title twice. I guess I'm the only one who's sensitive on this. I think it's opinionated and it shouldn't be an issue.

Genne23v (dismissed):

app/components/certificate/certificate-available.tsx
dadolhay: @Ririio If you format this in the parent component,  you sould probably use a separate name, it did send me for a spin too at the first time
Suggest validFromFormatted, validToFormatted

Ririio: You're right... It can cause problems if I use the same name format for two completely different data type. I'll be more careful for next time

dadolhay (dismissed):
LGTM

humphd (dismissed):

sfrunza13:
Is there something blocking this from going in? It seems as though it has been reviewed and the reviews were addressed.

sfrunza13:
> Is there something blocking this from going in? It seems as though it has been reviewed and the reviews were addressed.

I just looked at timestamps, my bad.

Myrfion (dismissed):

Ririio:
I had to reset the commit to a previous one due to problems with accidental merging instead of rebase. The following commits after were merely a change in the props name, and resolving an issue with theme directory so there shouldn't be any difference except that it's all rebased instead of multiple commit lines

Genne23v (approved):

dadolhay (approved):

PR #265, opened 2023-02-22
Don't force insecure requests to upgrade to HTTPS for dev/testing

### LINKED ISSUE #244: Sign in does not work on Safari
### Local Environment
After clicking the `Sign In` button to localhost on safari, the following is shown. Safari uses `https` but even after manually changing it to `http`, the login page is shown again:
(Video provided by @Genne23v)

https://user-images.githubusercontent.com/67077705/219962357-4baa7cdf-bfab-4b07-86de-0ace58612c95.mp4

### Test Environment

This also resulted in these failed tests: https://github.com/DevelopingSpace/starchart/actions/runs/4213508191. The tests fail for Desktop (webkit) and Mobile Safari
> Please ignore the failed test for chrome since that was due to missing wait time for sign in.

A video of the failed test run, taken from the above link (download the `playwright report` artifact, look under failed tests, and retry#1)

https://user-images.githubusercontent.com/67077705/219923031-5418df78-3aed-438c-8659-cfeaded30696.mp4

Some investigation is required to find out what is happening here. It seems like the Sign in button is clicked but no redirect happens.

humphd:
Fixes #244.  This alters our CSP settings to not force HTTP origins to be upgraded to HTTPS unless we are in production.  For local dev and testing (CI), we'll use HTTP.  I've also changed the way the app runs in e2e tests so it doesn't use `production` either. 

To test this, run the server and load using Safari.  If you can login, it works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "db:studio": "prisma studio",
     "setup": "run-s db:generate db:push db:seed",
     "start": "cross-env NODE_ENV=production node ./build/server.js",
-    "start:e2e": "cross-env NODE_ENV=production node --require dotenv/config ./build/server.js",
+    "start:e2e": "cross-env NODE_ENV=testing node --require dotenv/config ./build/server.js",
     "test": "cross-env SECRETS_OVERRIDE=1 NODE_OPTIONS=--require=dotenv/config vitest",
     "test:e2e:dev": "cross-env PORT=8080 start-server-and-test dev http://localhost:8080 \"playwright test\"",
     "pretest:e2e:run": "cross-env SECRETS_OVERRIDE=1 run-s build",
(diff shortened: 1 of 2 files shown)


SerpentBytes (commented):
It works locally on Safari. We could do something similar, so we don't have to switch to Chrome or Firefox to access MailHog UI.

SerpentBytes (approved):

Eakam1007 (approved):
This still fails in the testing environment but since it seems to work on safari outside of the test environment, I think it might be a different issue for the tests

Eakam1007:
We do not have any tests for logging in on main, but I was trying to run a test for `dns record` creation on another branch, and for that I needed to login first. That is where the login failed on safari ([failed run here - mobile chrome failure was due to a different error which was fixed](https://github.com/DevelopingSpace/starchart/actions/runs/4213508191))

humphd:
Let's fix that when we add a test for login.

PR #267, opened 2023-02-22
Fix dev run script for windows

### LINKED ISSUE #266: Run dev does not work on windows
Re: https://github.com/DevelopingSpace/starchart/blob/bb5b013101ea24ec409a383000a3a161c5bc802b/package.json#L9

This does not work on windows; likely because `cross-env` is missing.

Eakam1007:
Fixes #266 

Added `cross-env` to script so that `npm run dev` works on windows

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "build": "run-s build:*",
     "build:remix": "remix build",
     "build:server": "esbuild --platform=node --format=cjs ./server.ts --outdir=build --bundle",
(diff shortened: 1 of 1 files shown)


sfrunza13 (approved):
Is this what leads to the following?
```
 'SECRETS_OVERRIDE' is not recognized as an internal or external command,
operable program or batch file.
```

SerpentBytes (approved):

humphd:
Thanks for catching this and the fix, @Eakam1007!

Issue #115, opened 2023-02-03
Send email notification when SSL certificate is going to expire


humphd:
Similar to #113, we need to automate sending email notifications before SSL certificates expire.  Let's Encrypt certificates are good for 90 days.  We need to alert the user before they will expire so that they can renew them.  I'm not sure when we should do this (1 month before? 1 week before?).

humphd:
Could do, yes.

PR #268, opened 2023-02-23
Add Worker/Queues to process notification based on status

### LINKED ISSUE #108: Send email notification when domain becomes active or errors
Our Route53 code has a way to query the state of a DNS record.  Route53 needs to sync the record to multiple DNS servers before it's available.  We'll know if it's `PENDING` or `INSYNC`, and in the database schema , this translates to a record's `status` being one of `active`, `error`, or `pending`.

When a user creates or updates a DNS record, we should monitor this process and when it completes, send an email message to the user.  The process will go from `status=pending` to 1. `status=active` or 2. `status=error`.  In the case of 1. the DNS is now ready for use.  In the case of 2. something went wrong, and the user should probably try again.

### LINKED ISSUE #113: Send warning notification before expiring a DNS record
In #112 we're going to add support for expiring/removing DNS records once they expire.  We should send a notification to the user **before** we do this, so that they can renew them if we like.

We should have a configurable environment variable that indicates how close to the expiry date we should send the notification.  Similar to how #112 works (and perhaps as part of it), we can query to find all records that are going to expire within the defined period, and send email notifications to the users.

SerpentBytes:
## Description
This PR resolves #113, #115. 

### What's working?
- [x] Worker/Queue pair to run checks every 5 mins for DNS record and certs expiring in less than a month
- [x] Send Notification once.

### Steps to Test / Re-Test
- Delete  `redis` data and restart the Redis container (clear any previous jobs)
- `npm run setup` (to update to DB)
- Run the project (make sure Redis is up and running)
- `npm run db:studio` (to see the `lastNotified` field getting updated)
- Navigate to the MailHog interface at http://localhost:8025 to see notifications after 5 mins _(see Note)_

> **Note**
> To test if the jobs are repeating without waiting for 5 mins, modify the value of `repeat` [here](https://github.com/DevelopingSpace/starchart/pull/268/files#diff-db77765e8ad712b1a2c77389faf7bd213fc792da9cb30622f09719a57821e5a4R113) to every 5 seconds.


humphd:
Filed https://github.com/DevelopingSpace/starchart/issues/270 for the DNS flow I'm describing above.

dadolhay:
I agree with what @humphd proposed, with the exception, that this should not only run once a day, it should run every hour or even every 5 minutes.

If you allow tasks to pile up throughout the day, it will have to send a lot of mails all at once, and it is a lot easier to hit a rate limit + 3rd party providers do not really like large spike-like loads. I'd suggest to run it often, and only handle the tasks that are due, that represents a lot more even load on the system.

humphd:
> I agree with what @humphd proposed, with the exception, that this should not only run once a day, it should run every hour or even every 5 minutes.
> 
> If you allow tasks to pile up throughout the day, it will have to send a lot of mails all at once, and it is a lot easier to hit a rate limit + 3rd party providers do not really like large spike-like loads. I'd suggest to run it often, and only handle the tasks that are due, that represents a lot more even load on the system.

Sounds good, and something we can do load testing on later, then measure and tune.

SerpentBytes:
Just so the requirements are clear, for this PR, I  am being asked to:
- evolve the current code so it closes  #115
- Repeat job processing every 5 mins to prevent getting rate-limited

humphd:
One thing to call out with repeat jobs: we're going to need to not spam users. For example, if you run this every 5 mins (or every day), we need some way to keep track of the fact that we've already notified a user and don't need to do it again.  There are various ways to approach this, and we'll need one of them.

SerpentBytes:
I have modified the code to check for expiring records and certificates every 5 mins and persist if the user is notified. At the moment the type is boolean, but we can replace it with a number value so we can send two reminders.

humphd:
Let's move this to 0.5.  I'm not going to be able to review it fully in time for shipping today.

app/queues/notifications/certificate-expiration-monitor.server.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+);
+
+certificateExpirationCheckQueue.on('error', (err) => {
+  logger.warn('Error running check for domains about to expire', err);
+});
+
+export const addCertificateExpirationCheckJobs = async () => {
+  let jobName = `${certExpirationQueueName}:${new Date().toLocaleString('en-CA')}`;
+  return await certificateExpirationCheckQueue.add(jobName, await getCertificatesByExpiration(), {
```
Genne23v, line 54: I think `await` is redundant before `return`

app/queues/notifications/certificate-expiration-monitor.server.ts
```diff
@@ -0,0 +1,59 @@
...
+  let monthFromToday = new Date();
+  monthFromToday.setMonth(monthFromToday.getMonth() + 1);
+  return await prisma.certificate.findMany({
+    where: {
+      AND: [
+        {
+          validTo: {
+            lte: monthFromToday,
+          },
+          isNotified: false,
```
Genne23v, line 23: I think we should be able to send multiple notifications before expiration. Sorry if I got it wrong, `isNotified` prop sounds we will send only one notification. I'm not sure what is the right interval to send notifications. I think we can consider `lastNotificationSent` prop to compare if the notification is sent today, it will not send again.

app/queues/notifications/record-expiration-monitor.worker.ts
```diff
@@ -0,0 +1,53 @@
...
+import { redis } from '~/lib/redis.server';
+import logger from '~/lib/logger.server';
+import type { DomainStatusData } from './record-cert-expiration-data.interface';
+import { addNotification } from './notifications.server';
+import { prisma } from '~/db.server';
+
+export const recordExpirationQueueName = 'notifications-recordExpiration';
+
+const updateRecordNotifiedStatus = async (recordId: number) => {
+  return await prisma.record.update({
```
Genne23v, line 11: Redundant `await` before `return`

Genne23v (changes requested):

app/queues/notifications/certificate-expiration-monitor.server.ts
SerpentBytes: That was [my initial thought](https://github.com/DevelopingSpace/starchart/pull/268#issuecomment-1445208861). For the scope of this PR, I set it to a boolean, as we are sending notifications 1 month prior to expiration. However, to make the code extendible, I would reflect this change.

app/queues/notifications/certificate-expiration-monitor-worker.server.ts
```diff
@@ -0,0 +1,53 @@
...
+import { redis } from '~/lib/redis.server';
+import logger from '~/lib/logger.server';
+import type { DomainStatusData } from './record-cert-expiration-data.interface';
+import { addNotification } from './notifications.server';
+import { prisma } from '~/db.server';
+
+// name for the queue
+export const certExpirationQueueName = 'notifications-certExpiration';
+const updateCertificateNotifiedStatus = async (certId: number) => {
+  return await prisma.certificate.update({
```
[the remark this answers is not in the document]
dadolhay: in this case async too :)
- Await will happen automatically at return
- since you are returning a promise from prisma.certificate.update, no need to do async, you'll get a promise either way

app/queues/notifications/certificate-expiration-monitor.server.ts
```diff
@@ -0,0 +1,59 @@
...
+);
+
+certificateExpirationCheckQueue.on('error', (err) => {
+  logger.warn('Error running check for domains about to expire', err);
+});
+
+export const addCertificateExpirationCheckJobs = async () => {
+  let jobName = `${certExpirationQueueName}:${new Date().toLocaleString('en-CA')}`;
+  return await certificateExpirationCheckQueue.add(jobName, await getCertificatesByExpiration(), {
+    repeat: { pattern: secrets.NOTIFICATIONS_REPEAT_PATTERN },
```
dadolhay, line 55: I don't think that this needs to be a secret, I think it;s quite enough to either hardcode it, or use an env

app/queues/notifications/record-expiration-monitor.worker.ts
```diff
@@ -0,0 +1,53 @@
...
+    where: {
+      id: recordId,
+    },
+    data: {
+      isNotified: true,
```
dadolhay, line 16: Question: Where do you set this back to false? When the renew button is pressed?

app/queues/notifications/certificate-expiration-monitor.server.ts
```diff
@@ -0,0 +1,59 @@
...
+  });
+};
+
+export const certificateExpirationCheckQueue = new Queue<DomainStatusData[]>(
+  certExpirationQueueName,
+  {
+    connection: redis,
+    defaultJobOptions: {
+      attempts: 3,
+      backoff: {
```
dadolhay, line 40: Since we run this periodically, I don't think that retry is needed at all

dadolhay (changes requested):

app/queues/notifications/record-expiration-monitor.worker.ts
SerpentBytes: Good catch. I actually forgot we could renew certificates.

SerpentBytes: The scope of this PR is to just just send one month notice. I am not sure if I am supposed to set the value of isNotified to true in this PR.

app/queues/notifications/certificate-expiration-monitor.server.ts
SerpentBytes: I think it would be better to set it as env so it's easier to update the value if needed.

prisma/seed.ts
```diff
@@ -117,7 +117,7 @@ async function seed() {
       privateKey:
         '-----BEGIN CERTIFICATE-----ApfFCv0O65TYkp5jEWSlU8PhKYD43nXA=-----END CERTIFICATE-----',
       validFrom: new Date(),
-      validTo: certExpDate,
+      validTo: new Date(),
```
dadolhay, line 120: Could you please remind me what is the reason for this change?
It seems inclrrect to set the validFrom and validTo both to the same date

.env.example
```diff
@@ -25,6 +25,8 @@ LOG_LEVEL=debug
 NOTIFICATIONS_EMAIL_USER="no-reply@senecacollege.ca"
 SMTP_PORT=1025
 
+NOTIFICATIONS_REPEAT_PATTERN="* */5 * * * *"
```
dadolhay, line 28: I'm not sure if this is correct. I can see that this is what the BullMQ docs say, but I still don't think it is correct:

If I read this correctly, you are saying, repeat this **every second** if `(minute % 5) === 0` (minute can be devided by 5 and there is no remainder)

I'd be much happier if we used `repeat: {every: 5 * 60 * 1000}`. A lot of people don't understand the cronjob format

source: https://docs.bullmq.io/guide/jobs/repeatable
and I think their docs is incorrect
> pattern: '* 15 3 * * *',
> // Repeat job once every day at 3:15 (am)

in this case it should have been `0 15 3 * * *` ... theirs should repeat in every second during that 1 minute period ... at least according to how crond works

prisma/seed.ts
SerpentBytes: Typo. It should should just validTo that should be set to new Date.

dadolhay: What I mean is
```
validFrom: new Date(),
validTo: new Date(),
```
will create a (mock) certificate with a 0 valid time duration. We should either move validFrom to the past, or validTo to the future (as it was originally)

I don't see the reason we are doing this change... am I missing something?

SerpentBytes: I am fetching certificates based on expiration. I wanted an entry that will always expire in less than a month so we can send a notification. That's why I set validTo to a new date so when I run the query, it doesn't return empty.

SerpentBytes: I could make validTo to be seven days from the current date.

dadolhay: That sounds good!

app/queues/notifications/record-expiration-monitor.worker.ts
SerpentBytes: > It depends, If record renew is already complete, then by introducing this field, you should also integrate it to the appropriate logic to set it correctly (on record renew set it to false)
> 
> If record renew is not yet implemented, it will fall on that task to also add this. Not sure what's the state of that functionality

I don't think the record renew feature is fully implemented yet. Please let me know if I am wrong.

.env.example
```diff
@@ -25,6 +25,8 @@ LOG_LEVEL=debug
 NOTIFICATIONS_EMAIL_USER="no-reply@senecacollege.ca"
 SMTP_PORT=1025
 
+NOTIFICATIONS_REPEAT_PATTERN=5 * 60 * 1000
```
humphd, line 28: I don't think this needs to be configurable.  I would hard-code it into the app.

prisma/schema.prisma
```diff
@@ -32,6 +32,7 @@ model Record {
   updatedAt   DateTime     @updatedAt
   expiresAt   DateTime
   status      RecordStatus
+  timesNotified Int    @default(0)
```
humphd, line 35: Adding @cychu42 to this discussion for thoughts.

I'm not convinced about this approach (i.e., having a single integer count).  We have different types of notifications we want to send, and a count isn't going to cover them all.  I wonder if we should add a new table?

I don't have a solid proposal, but imagine something like this:

```prisma
model Notification {
  id            Int          @id @default(autoincrement())
  username      String
  sentAt        DateTime     @default(now())

  // This isn't perfect: we need to keep track of the record/cert this is associated with...
  certificateId Int?
  certificate   Certificate  @relation(fields: [certificateId], references: [id])

  recordId      Int?
  record        Record       @relation(fields: [recordId], references: [id])
}
```

Here we have the following info that we can use to track when we've sent notifications to users, specifically:

1. Who it is sent to (`user`)
2. When we sent it (`sentAt`)
3. What it relates to (a `certficateId` OR `recordId`).  There is probably a cleaner way to do this.  Do we need to use 2 tables instead?  Maybe that's better, but seems like a lot of extra infrastructure to me.

Using this, we could solve various problems. For example:

1. I need to send a notification about a specific expiring certificate to a user.  I can check to see if I've already sent one in the past X days, and either send it or ignore the request to send it based on the presence/absence of the notification record.
2. Similar to 1., I need to notify about an expiring record
3. Similar to 2., I need to notify about an expiring record that has already been renewed (e.g., there is already a notification record for this DNS record, but we need to send another one).  To do this, I can use the user + record + sent date, and figure out if I need to send another one (e.g., has it been X days or something).

I think we can improve on what I'm suggesting above, and hopefully @cychu42, @dadolhay and others can chime in with suggestions, too.

app/queues/notifications/certificate-expiration-monitor-worker.server.ts
```diff
@@ -0,0 +1,53 @@
+import { Worker } from 'bullmq';
+import { redis } from '~/lib/redis.server';
+import logger from '~/lib/logger.server';
+import type { DomainStatusData } from './record-cert-expiration-data.interface';
+import { addNotification } from './notifications.server';
+import { prisma } from '~/db.server';
```
humphd, line 6: Please separate your imports into 3 sections:

```ts
// Third-party modules
import { Worker } from 'bullmq';
...

// Internal modules
import { redis } from '~/lib/redis.server';
...

// Types
import type ...
```

This makes the code much easier to read and understand where things are coming from, and how they are being used.

app/queues/notifications/certificate-expiration-monitor-worker.server.ts
```diff
@@ -0,0 +1,53 @@
...
+
+// name for the queue
+export const certExpirationQueueName = 'notifications-certExpiration';
```
humphd, line 9: See discussion in https://github.com/DevelopingSpace/starchart/pull/289#pullrequestreview-1320483880.  I think we can likely co-locate all notification logic into a single queue, since the process is going to be very similar in all cases (e.g., do a query, based on results create notification jobs).  The worker can use an if/switch to allow for slightly different logic in each case.

app/queues/notifications/certificate-expiration-monitor.server.ts
```diff
@@ -0,0 +1,53 @@
+import { Queue } from 'bullmq';
```
humphd, line 1: Same comment here about ordering imports.

app/queues/notifications/certificate-expiration-monitor.server.ts
```diff
@@ -0,0 +1,53 @@
...
+  certExpirationQueueName,
+} from './certificate-expiration-monitor-worker.server';
+
+const { NOTIFICATIONS_REPEAT_PATTERN } = process.env;
+
+// fetch records with by status
+const getCertificatesByExpiration = async () => {
+  let monthFromToday = new Date();
+  monthFromToday.setMonth(monthFromToday.getMonth() + 1);
+  return await prisma.certificate.findMany({
```
humphd, line 17: You don't need `async` and `await` in this code, because you aren't consuming the result of the `Promise`.  You can return the `Promise` directly and let the caller `await` it.

app/queues/notifications/certificate-expiration-monitor.server.ts
```diff
@@ -0,0 +1,53 @@
...
+    },
+    select: {
+      user: true,
+      id: true,
+    },
+  });
+};
+
+export const certificateExpirationCheckQueue = new Queue<DomainStatusData[]>(
+  certExpirationQueueName,
```
humphd, line 36: This is another example of where co-locating code would make this easier, since I could look above this and find the name vs. having to jump between files.

app/queues/notifications/record-expiration-monitor.server.ts
```diff
@@ -0,0 +1,52 @@
...
+  recordExpirationQueueName,
+} from './record-expiration-monitor.worker';
+
+const { NOTIFICATIONS_REPEAT_PATTERN } = process.env;
+
+// fetch records with by status
+const getRecordsByExpiration = async () => {
+  let monthFromToday = new Date();
+  monthFromToday.setMonth(monthFromToday.getMonth() + 1);
+  return await prisma.record.findMany({
```
humphd, line 17: Same issue with `async` and `await`.  Also, this code is very similar to the above, with only the query being different.  We can consolidate this into a single worker/queue that does a different query.

humphd (changes requested):
OK, I've done a first pass through this code.  I have a few design issues I'd like to see figured out, and also gave some feedback on specific bits of code.

As discussed on Slack, I think you are over-complicating this code by breaking it up into separate files.  For example, you have to manually start the worker processes, have to export internal types, etc.  The pattern we have already established in other parts of the code is to combine queue/worker/type logic in the same file.  There isn't so much of it that we are benefiting from the separation.

prisma/schema.prisma
dadolhay: Why don't we add a date field here, `notifyExpirationAt: Date`

When we create/renew a record, we set it 1 month before the expiry
If the date is in the past, we can send the notification, and mve this notifyExpirationAt to a future date or null if we don't want to send anymore emails

This would not require an extra table and related complexity

SerpentBytes: @dadolhay, thanks for the feedback. I'd wait to hear @cychu42 thoughts on this and then implement the changes.

humphd: I think the idea of using a date field is clever, since it might be enough to allow us to do this all in a the main tables (i.e. not adding new ones).

If it was a date of when we last notified a user about an expiry, we could do the following:

1. We add a `notifiedAt` field (name TBD) to the dns records and certs tables
2. When we create a record/cert, we'd leave it `notifiedAt` empty
3. When a record/cert is going to expire (e.g., 1 month out?) we send an email and update `notifiedAt` to current date time
4. When we renew a record/cert, we update the `expiry` and clear `notifiedAt`
5. The process repeats

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,189 @@
...
+enum Type {
+  Certificate = 'certificate',
+  Record = 'record',
+}
+
+// name for the queue
+const expirationNotificationQueueName = 'expiration-notification';
+
+const updateNotificationStatus = (type: string, id: number) => {
+  // throw if any of the arguments are missing
```
humphd, line 23: The way your types are defined in 22, this can never happen.

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,189 @@
...
+
+enum Type {
+  Certificate = 'certificate',
+  Record = 'record',
+}
+
+// name for the queue
+const expirationNotificationQueueName = 'expiration-notification';
+
+const updateNotificationStatus = (type: string, id: number) => {
```
humphd, line 22: Should `type` not be `Type` or a union type: `'certificate | 'record'`?

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,189 @@
...
+        },
+      });
+  }
+};
+
+// fetch records with by status
+const getRecordsByExpiration = (type: string) => {
+  let monthFromToday = new Date();
+  monthFromToday.setMonth(monthFromToday.getMonth() + 1);
+  if (!type) {
```
humphd, line 55: Can't happen based on your type in 52

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,189 @@
...
+expirationNotificationQueue.on('error', (err) => {
+  logger.warn('Error running check for DNS Records/Certificates about to expire', err);
+});
+
+// function to add jobs
+export const addExpirationNotification = async (type: string) => {
+  let jobName: string;
+  let data: ExpirationStatusPayload[];
+
+  if (!type) {
```
humphd, line 114: Can't happen

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,189 @@
...
+    throw new Error('argument: type for addExpirationCheckJobs cannot be null or undefined');
+  }
+  switch (type) {
+    case Type.Certificate:
+      data = await getRecordsByExpiration(Type.Certificate)!;
+      if (data.length) {
+        logger.info('Entering if');
+        jobName = `${expirationNotificationQueueName}-${
+          Type.Certificate
+        }:${new Date().toLocaleString('en-CA')}`;
```
humphd, line 124: We have to be careful here.  The job name needs to be unique, but, it also needs to be the same per job.  The reason for this is that 5 mins later, we don't want to add yet another job for the same task.  Also, if we have multiple instances of the server running, we don't want to have both servers start identical jobs.  Make the job name unique by using the user's name, certificate id, etc. but not a date.

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,189 @@
...
+        return expirationNotificationQueue.add(jobName, data, {
+          repeat: { every: 5 * 60 * 1000 },
+        });
+      }
+      break;
+    case Type.Record:
+      data = await getRecordsByExpiration(Type.Record)!;
+      if (data.length) {
+        jobName = `${expirationNotificationQueueName}-${Type.Record}:${new Date().toLocaleString(
```
humphd, line 133: Same thing here

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,189 @@
...
+import { Worker, Queue } from 'bullmq';
+
+// Internal modules
+import { redis } from '~/lib/redis.server';
+import logger from '~/lib/logger.server';
+import { addNotification } from './notifications.server';
+import { prisma } from '~/db.server';
+
+// Types
+import type { ExpirationStatusPayload } from './record-cert-expiration-data.interface';
```
humphd, line 11: Because this type info is only used in this file, move the types into this file.

app/queues/notifications/record-cert-expiration-data.interface.ts
```diff
@@ -0,0 +1,14 @@
...
+interface User {
+  username: string;
+  firstName: string;
+  lastName: string;
+  email: string;
+  createdAt: Date;
+  updatedAt: Date;
+}
+export interface ExpirationStatusPayload {
+  user: User;
```
humphd, line 11: When you move this into the other file, you can also import the `User` type from Prisma, and use that vs. redefining it yourself.

humphd (changes requested):
This is looking better, nice work.  I've made some more comments in the code.

Regarding failing a job and having it sync to db.  Two thoughts:

1. I would not cache the data in the job, but have your worker go back to the database and check the current state of the data, then update based on that.
2. If you ever need to throw an error that will stop the job, you can do it like we do in https://github.com/DevelopingSpace/starchart/blob/main/app/queues/certificate/order-creator-worker.server.ts#L72-L81

app/queues/notifications/expiration-notification.server.ts
SerpentBytes: changed this to `RecordType`

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,160 @@
...
+export interface ExpirationStatusPayload {
+  type: RecordType;
+}
+
+// name for the queue
+const expirationNotificationQueueName = 'expiration-notification';
+
+const updateNotificationStatus = (type: RecordType, id: number) => {
+  // throw if any of the arguments are missing
+  if (!type || !id) {
```
humphd, line 26: I'm confused about the need to include these.  The types you define in 24 don't allow them to be null/undefined.  Can you clarify why you do this extra check?

app/queues/notifications/expiration-notification.server.ts
humphd: Same question

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,160 @@
...
+};
+// worker instance to process DNS Record/Certificates expiration notification jobs
+const expirationNotificationWorker = new Worker<ExpirationStatusPayload>(
+  expirationNotificationQueueName,
+  async (job) => {
+    const { type } = job.data;
+    try {
+      logger.debug(`Processing job: ${job.name}`);
+      let records = await getRecordsByExpiration(type);
+      records.map(async (record) => {
```
humphd, line 135: You are mixing synchronous and asynchronous code paths here.  Your `.map()` will complete immediately, but the array will contain a bunch of Promises that need to get awaited.  You can fix it like so:

```ts
await Promise.all(records.map(async (record) => {
  ...
  await updateStatusAndNotify(...);
}))
```

humphd (changes requested):

app/queues/notifications/expiration-notification.server.ts
SerpentBytes: That's an error. 

To clarify: if I infer the type, there is no need to check because the value can't be null or undefined, as Typescript won't allow it.

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,152 @@
...
+const expirationNotificationWorker = new Worker<ExpirationStatusPayload>(
+  expirationNotificationQueueName,
+  async (job) => {
+    const { type } = job.data;
+    try {
+      logger.debug(`Processing job: ${job.name}`);
+      let records = await getRecordsByExpiration(type);
+      await Promise.all(
+        records.map(async (record) => {
+          if (!record.lastNotified) {
```
humphd, line 126: We're going to need something more here, so we can deal with renewals.  The first time we notify a user for a record/cert, it will work.  But if they renew, it won't ever notify again in the future.  I wonder if this should really be a check to see if it's a) not null/undefined; or b) a date more than 30 days in the past?

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,152 @@
...
+            await updateStatusAndNotify(type, record.id, {
+              emailAddress: record.user.email,
+              subject: `Sample ${type} expiration notice`,
+              message: `Sample ${type} expiration notice message`,
```
humphd, line 130: We should file a follow-up to figure out what this email subject/message will be later on.  Can you do that before we land this?

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,152 @@
...
+            });
+          }
+        })
+      );
+
+      logger.info(`Completed job: ${job.name}`);
```
humphd, line 136: Make this a `.debug`

humphd (changes requested):
Looking good.  A few things I noticed in this read through.

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,152 @@
...
+    connection: redis,
+  }
+);
+
+expirationNotificationQueue.on('error', (err) => {
+  logger.warn('Error running check for DNS Records/Certificates about to expire', err);
+});
+
+// function to add jobs
+export const addExpirationNotification = async (type: RecordType) => {
```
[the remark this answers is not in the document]
SerpentBytes: > Actually, another question: do we even need to expose this to callers? Since we'll do this on a repeated cycle, should we call it ourselves internally on startup in this file, and no one ever needs to interact with it?

Currently, we are using the /dev route to initiate repeating jobs, so I need to expose it to call it in the front end. It should be automated on startup. 

I could look into ways to start the process automatically without user interaction.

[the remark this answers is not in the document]
humphd: When you say "currently," it's being added in this PR.  I would skip putting it in `/dev` and do it on startup.  We have an `init()` model now (see how SAML and DNS do their startup logic once).

app/queues/notifications/expiration-notification.server.ts
SerpentBytes: > We're going to need something more here, so we can deal with renewals. The first time we notify a user for a record/cert, it will work. But if they renew, it won't ever notify again in the future. I wonder if this should really be a check to see if it's a) not null/undefined; or b) a date more than 30 days in the past?

We could do a combination of both. Initially, it's null, and it is our cue to tell that we have never notified a user. We should also check whether the value of `lastNotified` is set to a month in the past (as you are suggesting), as this would be our cue to re-notify the user.

humphd (approved):
Great!  I'll get @Genne23v and @dadolhay to review it too, since they are doing such similar stuff.

humphd (approved):

app/lib/notifications.server.ts
```diff
@@ -46,4 +51,10 @@ const sendNotification = async (emailAddress: string, subject: string, text: str
   }
 };
 
+export async function init() {
+  logger.debug('Notifications init: adding jobs for certificate/record expiration notices');
+  await addCertificateExpirationNotifications();
+  await addRecordExpirationNotifications();
```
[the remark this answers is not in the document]
sfrunza13: Actually I don't know that there is, this looks good.

sfrunza13 (approved):

app/lib/notifications.server.ts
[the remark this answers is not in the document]
SerpentBytes: It's helpful to know that background jobs are being added automatically via the log message, as it will happen on startup without any UI interaction starting the process.

SerpentBytes:
Fixed conflicts with the latest merged code.

Genne23v (dismissed):
Looking good!

humphd (dismissed):

SerpentBytes:
> It Looks like there's a conflict to sort out still.

Yes. I am just waiting for a few more reviews.

sfrunza13 (approved):

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,158 @@
...
+// fetch records by status based on type
+const getRecordsByExpiration = (type: RecordType) => {
+  switch (type) {
+    case RecordType.Certificate:
+      return prisma.certificate.findMany({
+        where: {
+          validTo: {
+            lte: dayjs().add(1, 'M').toDate(),
+          },
+          lastNotified: null,
```
dadolhay, line 57: @SerpentBytes So we only send a notification once?

humphd (approved):

app/queues/notifications/expiration-notification.server.ts
```diff
@@ -0,0 +1,172 @@
...
+        where: {
+          validTo: {
+            lte: dayjs().add(1, 'M').toDate(),
+          },
+          OR: [
+            {
+              lastNotified: null,
+            },
+            {
+              lastNotified: dayjs().subtract(30, 'd').toDate(),
```
dadolhay, line 62: I would suggest adding a constant NOTIFICATION_FREQUENCY, default 7 (days)
```suggestion
              lastNotified:{
                lte: dayjs().subtract(NOTIFICATION_FREQUENCY, 'd').toDate(),
              },
```
The above is just a tip, but I think it should work :)

SerpentBytes: Thank you. I have updated the code to reflect the changes.

dadolhay (approved):
Nice work!

humphd (approved):
Looks good to me.

PR #269, opened 2023-02-23
ci: autogenerate release

### LINKED ISSUE #249: Automate release at the end of milestone
## Description
Automate the process of:
- bumping project version
- releasing on GitHub with changelog

shawnyu5:
Closes #249

A GitHub actions to auto generate a release tag upon the push of a new tag

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,7 +3,7 @@ name: release
 on:
   push:
     tags:
-      - '*.*.*'
+      - '*'
 
 jobs:
   Release:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -13,7 +13,7 @@ jobs:
     # needs: [Build, Prettier, Unit-Tests, E2E-Tests, Dockerfile-Lint, ESLint, Type-Check]
     steps:
       - name: Generate release
-        uses: 'marvinpinto/action-automatic-releases@latest'
+        uses: 'marvinpinto/action-automatic-releases@v1.2.1'
         with:
           repo_token: '${{ secrets.GITHUB_TOKEN }}'
           automatic_release_tag: 'latest'
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,7 +3,7 @@ name: release
 on:
   push:
     tags:
-      - 'v*'
+      - '*.*.*'
 
 jobs:
   Release:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -12,10 +12,18 @@ jobs:
     # TODO: not sure if this will work
     # needs: [Build, Prettier, Unit-Tests, E2E-Tests, Dockerfile-Lint, ESLint, Type-Check]
     steps:
+      - name: Set output
+        id: vars
+        run: echo "tag=${GITHUB_REF#refs/*/}" >> $GITHUB_OUTPUT
+      - name: Check output
+        env:
+          RELEASE_VERSION: ${{ steps.vars.outputs.tag }}
+        run: |
+          echo $RELEASE_VERSION
+          echo ${{ steps.vars.outputs.tag }}
       - name: Generate release
         uses: 'marvinpinto/action-automatic-releases@v1.2.1'
         with:
           repo_token: '${{ secrets.GITHUB_TOKEN }}'
           automatic_release_tag: 'latest'
           prerelease: false
-          # title: "Development Build"
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     # don't allow for publishing release unless everything else passes
     # TODO: not sure if this will work
-    needs: [Build, Prettier, Unit-Tests, E2E-Tests, Dockerfile-Lint, ESLint, Type-Check]
+    # needs: [Build, Prettier, Unit-Tests, E2E-Tests, Dockerfile-Lint, ESLint, Type-Check]
     steps:
       - name: Generate release
         uses: 'marvinpinto/action-automatic-releases@latest'
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -12,18 +12,12 @@ jobs:
     # TODO: not sure if this will work
     # needs: [Build, Prettier, Unit-Tests, E2E-Tests, Dockerfile-Lint, ESLint, Type-Check]
     steps:
-      - name: Set output
+      - name: Get tag version
         id: vars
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,32 @@
...
+  contents: write
+  pull-requests: write
+  statuses: write
+
+jobs:
+  Release:
+    runs-on: ubuntu-latest
+    # don't allow for publishing release unless everything else passes
+    # TODO: not sure if this will work
+    # needs: [Build, Prettier, Unit-Tests, E2E-Tests, Dockerfile-Lint, ESLint, Type-Check]
```
humphd, line 19: There are two ways to approach this.

1. Require that we only tag a commit that is known to be good (i.e., it's already landed on `main` and passed CI).  This is generally OK, since a release is a manual process (i.e., you tag).  Also, what if these fail?  You've already created the tag, so you need to do another patch release anyway.
2. You can separate your jobs out into separate files, and call them via `uses`.  We do this in Telescope, for example: https://github.com/Seneca-CDOT/telescope/blob/master/.github/workflows/release.yml#L14.  This is nice for re-use, but sucks for maintenance, since you can't update the jobs in a PR, and have to land changes first before you can test things.

I'd be tempted to do 1.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,32 @@
...
+      # get the version of the current git tag
+      - name: Get tag version
+        id: vars
+        run: echo "tag=${GITHUB_REF#refs/*/}" >> $GITHUB_OUTPUT
+      - name: Generate release
+        uses: 'marvinpinto/action-automatic-releases@v1.2.1'
+        with:
+          repo_token:
+            '${{ secrets.GITHUB_TOKEN }}'
+            # generate release name with the tag name that was just pushed
```
humphd, line 30: This is indented one too far I think

humphd (changes requested):
Did you also want to copy what I do in https://github.com/DevelopingSpace/starchart/blob/main/.github/workflows/ci.yaml#L311-L353 and have it also build and push a production image?  You'd probably use [`type=semver`](https://github.com/docker/metadata-action#typesemver) for this, and also bump `latest` and `main`

.github/workflows/release.yaml
shawnyu5: I agree, 1 seems like the better option. Since releasing is a manual process, we should have the freedom, if push comes to shove, to release with CI not passing (when that prettier job randomly started failing). I don't think its within the scope of this job to enforce all CI must pass before we can publish a release. That decision should be on the actual person creating the release.

.github/workflows/release.yaml
```diff
@@ -30,3 +27,42 @@ jobs:
             # generate release name with the tag name that was just pushed
           automatic_release_tag: ${{ steps.vars.outputs.tag }}
           prerelease: false
+  Docker-Build:
+    # We'll only build and push when landing on main, not for PRs
+    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
```
humphd, line 32: I don't think we need this, since we'll never run this in response to a PR, right (only on tag)?

.github/workflows/release.yaml
```diff
@@ -30,3 +27,42 @@ jobs:
...
+
+      - name: Extract Docker metadata from git info
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: ghcr.io/DevelopingSpace/starchart
+          flavor: latest=true
+          tags: |
+            type=semver,value=${{ steps.vars.outputs.tag }}
+      - name: Log in to the GitHub Container Registry
```
humphd, line 55: Blank line in between all steps, or none, but be consistent.

.github/workflows/release.yaml
humphd: This is still wrong

humphd (changes requested):

.github/workflows/release.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+permissions:
+  actions: write
+  contents: write
+  pull-requests: write
```
humphd, line 11: Can you educate me on this.  Why are these last 2 required?

.github/workflows/release.yaml
shawnyu5: Prettier formats this. This is not me...

.github/workflows/release.yaml
shawnyu5: I still need to mess around with those permissions. The github action to generate a tag needs some permissions to save the tag in the repo.

humphd: Would we not push the tag ourselves?  What is the workflow you imagine?

shawnyu5: Sorry, the gh action is used to generate the release when a tag is pushed. With default permissions given in a workflow, I get this error, that it doesnt have enough permissions to publish a new release

https://github.com/shawnyu5/starchart/actions/runs/4249711991/jobs/7390117298

humphd: That error makes it seem like it wants to create a new tag, which seems odd if we push our own.

humphd: I think you're right, and this is something I didn't fully understand.  I found [this](http://practicalseries.com/1002-vcs/09-06-github.html#js--090601) helpful:

> 9.6.1
> Tags and releases—the differences
Well a tag is just a unique symbolic name attached to a particular commit point. Git supports two types of tag ([§ 6.3.4](http://practicalseries.com/1002-vcs/06-03-example.html#js--060304)) a lightweight tag (without a message) and an annotated tag (with a message). In Brackets we only have lightweight tags.
>
>GitHub can display both. In [Figure 9.82](http://practicalseries.com/1002-vcs/09-06-github.html#) above, the tags are all lightweight, I can tell this because it tells me there is no release notes for any of the tags, this release note is what Git refers to as the tag message attached to the annotated tag. If we added a release note it would become an annotated tag.
>
>A release on the other hand is a GitHub object, it is still attached to a particular commit point, but this time it allows us to specify that a particular commit has a particular release status.
>
>GitHub converts all our tags to releases when we push them up from the local repository and that is why they are listed as both releases and tags.

So GitHub does want our manually created tag, and will use it to create a new GitHub Release (tag) object, which is why it wants elevated privileges.

shawnyu5: ah that makes sense. Tho I don't think it needs all those permissions there, I just put them there for now so I can move on to test the rest of the action. I should remove the ones the action does not need.

SerpentBytes:
@shawnyu5, just checking in to see if you are still working on this PR.

humphd:
Closing in favour of #466

PR #273, opened 2023-02-24
Seed 2 test users

### LINKED ISSUE #175: Create Users in Prisma Seed
As a follow up to #105 and #174 this issue aims to add 3 new users to seed. 2 of these users can be in the database and one of them should be used for the first time login case.

cychu42:
Closes #175 
This PR adds two more users to for seeding in `seed.ts` according to PR #174 as follow-up.

## Test
You can use `npm run db:seed` to seed, and then use `npm run db:studio` to see whether the users are created.
They should also be present after setting up a new instance of the project locally, since `npm run setup` will be used.

### DIFF
--- a/prisma/seed.ts
+++ b/prisma/seed.ts
@@ -19,13 +19,17 @@ async function seed() {
   await prisma.certificate.deleteMany().catch(() => {});
   await prisma.user.deleteMany().catch(() => {});
 
-  await prisma.user.create({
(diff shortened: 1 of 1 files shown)


sfrunza13 (approved):

SerpentBytes (approved):

cychu42:
Thank you.

Issue #270, opened 2023-02-23
Create DNS Flow skeleton like Certs Flow


humphd:
We've got two very similar, and related, processes to manage:

- DNS
- Certificates

Both now have an API:

- DNS: https://github.com/DevelopingSpace/starchart/blob/main/app/lib/dns.server.ts and 
- Certificates: https://github.com/DevelopingSpace/starchart/blob/main/app/lib/lets-encrypt.server.ts

The Certificates have an asynchronous flow, made-up of various sub-jobs: https://github.com/DevelopingSpace/starchart/tree/main/app/queues/certificate.

We need the same sort of thing for DNS.  There is some work in https://github.com/DevelopingSpace/starchart/blob/main/app/lib/domains.server.ts that might help with this, but what we really need to do is break down the process of working with DNS (Route53, MySQL, Notifications, etc) as a series of async jobs that can be performed in the background, and linked together as a flow.

Without this, we're going to waste a lot of time trying to hook everything together, so I'd like to prioritize the work, and get it started as soon as possible.

@dadolhay has the most experience thinking in terms of distinct, async jobs, so perhaps he can throw some thoughts in here.

@Genne23v are you interested in working on this?

Whoever does work on it, I'd *highly* recommend we do this in stages like @dadolhay did with the certs flow: build the skeleton, then build the individual workers to connect back to our existing APIs.  I'd like to avoid having someone go off for a month and try to build this without feedback.

To get the ball rolling, here's a draft of the process:

1. `addDnsRecord()` (or `update`, `delete`) takes record details and creates a new flow. A record is immediately put into MySQL with the state indicating that it's in process, so the front-end can begin to query for that data (and state)
3. A call to Route53 is made 
4. We ask Route53 when it's done using the SDK API (might take 2 or 3 tries)
5. When the record is done, we update the MySQL data with the new status (could be error or ok)
6. We send the user whatever notification(s) are necessary to let them know what has happened

What am I missing?

Genne23v:
@humphd I should've let you know that I'm working on this. It took much longer than I thought as I struggled with some errors. I just pushed #274 to resolve this issue. Although notification is missing in my code.

humphd:
Thanks @Genne23v.

Issue #112, opened 2023-02-03
Expire DNS records after a configurable amount of time


humphd:
We need to enforce and automate an expiry on DNS records.  When a user creates a DNS record, we need to specify an expiry date in the future.  The amount of time should be configurable via an environment variable.  Initially this can be 6 months in production, and we can use N days in staging (to make it easy to test).

We should run a query on a regular schedule (e.g., daily), which gets all DNS records from the database that have expired.  It then deletes these from Route53 and then from the database.  An email notification should be sent to the user indicating that the domain has expired and been removed.

Route53 has rate limits that we'll have to honour here.  We can't send hundreds of requests at once.  However, we can group a bunch of DELETE requests into a single change set and send that.

Where and how we run this job is open for debate.  We could do it in a child process of the main server.  We could also do it in a separate container that restarts every 24 hours.  We could also explore the worker-queue idea (but we might see if we can avoid this to start).

Issue #110, opened 2023-02-03
Enforce a configurable limit on the number of DNS records per user


humphd:
We need to be able to set an environment variable that imposes a limit on the number of DNS records a user can have.  When the user tries to add a new record, their current record count should be compared against this limit, and the operation should fail if it will mean exceeding the limit.

I'm not sure what the right way to enforce this is (e.g., in the db, in the model code, etc).

Genne23v:
In my opinion, if model declines the request, queue will continuously request although there is no second chance to make it happen. I think the dns worker should throw `UnrecoverableError`. Please let me know if it sounds like a plan.

PR #274, opened 2023-02-24
Add dns work flows

### LINKED ISSUE #217: Create a workflow to add domain to DNS and DB
Adding queue flow as below 

- create an initial record in MySQL for a domain with pending status
- talk to Route53 to create/update/delete a record
- poll Route53 to check connection status of the domain until it's ready
- update the MySQL record with the active or error status

Genne23v:
Close #217 and parts of #270

This PR adds workflows as below. 
1. Create a record in DB with `PENDING` status. Return `id` to update `status` later. 
2. Create a record in Route53 (`TTL` property added due to error)
3. Poll Route53 in every second to get the updated status (`INSYNC` or something else)
4. Update the DB either with `active` or `error` 

### To test this, it needs manual steps as below. 
- Create a `hostedZone` using unit test `dns.server.test.ts`. If you run all test, it will remove `hostedZone`. Run partial test. 
- Get `hostedZoneId` from `localhost:5053/moto-api` and set `AWS_ROUTE53_HOSTED_ZONE_ID` 
- Set `ROOT_DOMAIN` to `starchart.com`
- Go to `http://localhost:8080/dev` and hit Request DNS Record button


SerpentBytes (changes requested):
@Genne23v E2E tests are failing because WS_ROUTE53_HOSTED_ZONE_ID is missing.

Genne23v:
> @Genne23v E2E tests are failing because `WS_ROUTE53_HOSTED_ZONE_ID` is missing.



> @Genne23v E2E tests are failing because `WS_ROUTE53_HOSTED_ZONE_ID` is missing.

It shouldn't throw in testing environment. I will take a closer look tomorrow.

app/lib/dns.server.ts
```diff
@@ -86,6 +86,7 @@ export const upsertRecord = async (
                   Value: value,
                 },
               ],
+              TTL: 60 * 60 * 24,
```
dadolhay, line 89: Please use a shorter TTL

Context:
There are two "kinds" of DNS servers on the internet, "Authoritative" and " Recursive"

- Authoritative servers are the authority for some zone, I.e., our server will be the authority for senecastudent.ca
- Recursive (or recursor) DNS servers are only used to look up information from authoritative servers on the internet.

Since authoritative servers don't want to be bothered over and over again, they send out the TTL or Time To Live information with each response they give, saying "Here you go, and remember this for X seconds, don't bother me again with this until it expires"

This means, that if you change a value in the authoritative server (i.e. our route53) some DNS servers will still "remember" the old value for a period of time (no longer then the old value's TTL)

In practice this switchover follows a parabolic curve where the curve intersects with the X axis at TTL.

TL;DR
We don't want to leave the student's record in an inconsistent state for 24 hours after a change, so I suggest using like 5 minutes here (300)

app/lib/dns.server.ts
```diff
@@ -132,6 +133,7 @@ export const deleteRecord = async (
...
```
dadolhay, line 136: Same as above

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+  addDbRecordQueueName,
+  async (job) => {
+    logger.info('Creating a record in DB');
+    const { username, type, name, value } = job.data;
+
+    if (!username) {
+      throw new Error('Username is not provided');
+    }
+
+    const existingRecords = await prisma.record.count({
```
dadolhay, line 28: - I'd put in username just for good measure into the where too.
- There is a limit of how many records a user create in sum, should not we check that too?

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+      where: {
+        type,
+        name,
+        value,
+      },
+    });
+
+    if (existingRecords > 0) {
```
dadolhay, line 36:
```suggestion
    if (existingRecords) {
```
nit: You can rely on the falsyness of `0`

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    try {
+      const result = await prisma.record.create({
+        data: {
+          username,
+          type,
+          name,
+          value,
+          expiresAt: setMonthsFromNow(6),
```
dadolhay, line 47: Should not we get the 6 months from an env var? sounds like something that we might change later

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+          status: RecordStatus.pending,
+        },
+      });
+      return result.id;
+    } catch (error) {
+      logger.warn('Could not create a record in DB', { error });
```
dadolhay, line 53: You should rethrow the existing error object here to stop the process and indicate a failure
```suggestion
      logger.warn('Could not create a record in DB', { error });
      throw error;
```

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+export const createDnsRecordWorker = new Worker<
+  Pick<DomainRecord, 'username' | 'type' | 'name' | 'value'>
+>(
+  createDnsRecordQueueName,
+  async (job) => {
+    logger.info('Creating a record in Route 53');
+    const { username, type, name, value } = job.data;
+
+    try {
+      return await createRecord(username, type, name.toLowerCase(), value);
```
dadolhay, line 68: return automatically awaits always, no need to add it

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+    } catch (error) {
+      logger.warn('Could not create a record in Route53', { error });
```
dadolhay, line 70: please rethrow all errors, therwise this will only silently fail not stopping the process

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+  async (job) => {
+    logger.info('Checking Route53 readiness');
+
+    const values: any = await job.getChildrenValues();
+    const key = Object.keys(values)[0];
+    const changeId = values[key];
+    let status: string;
+
+    do {
+      status = await getChangeStatus(changeId);
```
dadolhay, line 87: This way you'll have a large number of active jobs. Instead of awaiting here, throw an error, and let the BullMQ retry logic re-run your job until it succeeds

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+export const syncDbStatusWorker = new Worker(
+  syncDbStatusQueueName,
+  async (job) => {
+    logger.info('Sync DB status with DNS');
+    const values = await job.getChildrenValues();
+    const keys = Object.keys(values);
+
+    let recordId, status;
+    if (keys.length !== 2) {
+      throw new Error('Invalid children values from previous workers');
```
dadolhay, line 104: This could be an UnrecoverableError from BullMQ

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+    } else {
+      if (keys[0].includes(addDbRecordQueueName)) {
```
dadolhay, line 106: Not sure if I understand this part. What does this do?

app/queues/dns/dns-flow.server.ts
```diff
@@ -0,0 +1,88 @@
...
+  username,
+  type,
+  name,
+  value,
+}: Pick<DomainRecord, 'username' | 'type' | 'name' | 'value'>) => {
+  // Step 1. Create a record in MySQL for a domain with pending status
+  const addDbRecord: FlowJob = {
+    name: `addDbRecord:${name}-${username}`,
+    queueName: addDbRecordQueueName,
+    data: { username, type, name, value } as Pick<
```
dadolhay, line 25: I added the `as` part here, so the worker itself can dictate what data it needs. I suggest to export the required props from the worker file itself and use that here, I don't see this pick adding any value (correct me if I'm wrong)

dadolhay: Please correct the other ones too in this file

dadolhay (changes requested):

humphd:
I filed #275 to improve the situation for local dev and needing a hosted zone in moto.

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
+import { delay, Worker } from 'bullmq';
+import { redis } from '~/lib/redis.server';
```
humphd, line 2: Separate your third-party, internal, and type imports with a blank line:

```ts
import { delay, Worker } from 'bullmq';
...

...

import type { DomainRecord } from '~/lib/domains.server';
...
```

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export const addDbRecordWorker = new Worker<
+  Pick<DomainRecord, 'username' | 'type' | 'name' | 'value'>
+>(
+  addDbRecordQueueName,
+  async (job) => {
+    logger.info('Creating a record in DB');
+    const { username, type, name, value } = job.data;
+
+    if (!username) {
```
humphd, line 24: Why would we not have a username?  Is it optional in `DomainRecord['username']`?  That seems odd to me.

app/queues/dns/create-record-worker.server.ts
humphd: #110 (could evolve it there, or solve that here).

app/queues/dns/create-record-worker.server.ts
humphd: I think with Winston (different from pino), you are supposed to use the `error` bare:

```ts
logger.warn('Could not create a record in DB', error);
```

app/queues/dns/create-record-worker.server.ts
humphd: I've read a few places in code today that used this magic number.  Agree we need to make it configurable, see #112.

app/queues/dns/create-record-worker.server.ts
humphd: Same comment on error interpolation with Winston.

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+  },
+  { connection: redis }
+);
+
+export const checkDnsStatusWorker = new Worker(
+  checkDnsStatusQueueName,
+  async (job) => {
+    logger.info('Checking Route53 readiness');
+
+    const values: any = await job.getChildrenValues();
```
humphd, line 81: Can we add some type info to the Worker definition above to avoid having to do all this untyped/unnamed property access?

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+    do {
+      status = await getChangeStatus(changeId);
+      await delay(1000);
+    } while (status === 'PENDING');
+    return status;
+  },
+  { connection: redis }
+);
+
+export const syncDbStatusWorker = new Worker(
```
humphd, line 95: Again, can we not type this?

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+        recordId = values[keys[0]];
+        status = values[keys[1]];
+      } else {
+        recordId = values[keys[1]];
+        status = values[keys[0]];
+      }
+    }
+
+    try {
+      switch (status) {
```
humphd, line 116: Simplify this, and move the condition into the value for the `status` vs. a switch:

```ts
await prisma.record.update({
  where: {
    id: recordId,
  },
  data: {
    status: status === 'INSYNC' ? RecordStatus.active : RecordStatus.error,
  },
});
```

humphd (changes requested):
Agree with @dadolhay's comments, and added some more of my own.

humphd:
Another thought I have while reading this.  We're starting to add a lot of queues.  Each queue requires more memory, Redis event handlers, etc.  At some point, we'll have to measure.  I know in the official docs, the BullMQ talks about queues not being a huge drain.

Related to the above, I also wonder if we should use the same queue or worker for all common tasks.  For example: all Route53 operations are going to get made against the same API, and share the same rate limiting.  We can set a [rate limit](https://docs.bullmq.io/guide/rate-limiting#manual-rate-limit) on a worker to avoid hitting this. 

From https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html, here are the Route53 API limits we need to respect:

> All Amazon Route 53 API requests
>For the [Amazon Route 53 APIs](https://docs.aws.amazon.com/Route53/latest/APIReference/API_Operations_Amazon_Route_53.html) five requests per second per AWS account. If you submit more than five requests per second, Amazon Route 53 returns an HTTP 400 error (Bad request). The response header also includes a Code element with a value of Throttling and a Message element with a value of Rate exceeded.
>
>Note
If your application exceeds this limit, we recommend that you implement exponential backoff for retries. For more information, see [Error Retries and Exponential Backoff in AWS](https://docs.aws.amazon.com/general/latest/gr/api-retries.html) in the Amazon Web Services General Reference.
>
>ChangeResourceRecordSets requests
>If Route 53 can't process a request before the next request arrives, it will reject subsequent requests for the same hosted zone and return an HTTP 400 error (Bad request). The response header also includes a Code element with a value of PriorRequestNotComplete and a Message element with a value of The request was rejected because Route 53 was still processing a prior request.

app/lib/dns.server.ts
Genne23v: Thanks for the detailed explanation. I didn't have much idea what is the right timeframe.

app/queues/dns/create-record-worker.server.ts
Genne23v: @humphd  Thank you for filing and sharing for me. I will update in next PR.

app/queues/dns/create-record-worker.server.ts
Genne23v: Yeah, I thought the same. I will handle this in another PR.

app/queues/dns/create-record-worker.server.ts
Genne23v: Will be updated

app/queues/dns/create-record-worker.server.ts
Genne23v: Will be updated

app/queues/dns/create-record-worker.server.ts
Genne23v: In my testing, values from `job.getChildrenValues()` always returns the values in order. So this is not necessary in my experience so far. But I decided to assume that it can be a different order just to be safe. So the first value in the object can be either one of `dbId` or `status`.

Genne23v:
> 

Thank you for the issue. I will assign myself and work as a next step.

app/queues/dns/create-record-worker.server.ts
Genne23v: Yes, it's definitely cleaner. Thank you

app/queues/dns/create-record-worker.server.ts
dadolhay: Thanks for the explanation

To be honest this looks a bit flaky, and hard(er) to maintain if the job dependency chain chnages. What I'd do is:
```js
    const childJobResults = await job.getChildrenValues();
    const addRecordJobKey = Object.keys(childJobResults).find(item => item.includes(addDbRecordQueueName))
    const recordId = childJobResults[addRecordJobKey]
```

This way no matter what the order was, how many steps you had before, you always get the correct result

This is just a suggestion, so if you don't agree, feel free to ignore it

app/queues/dns/create-record-worker.server.ts
Genne23v: I'm not sure if I understand the intention correct. Please let me know if I misunderstood your intention. 

```js
    const status = await getChangeStatus(changeId);

    if (status === 'PENDING') {
      throw new Error('Change status is still pending');
    }
    return status;
``

Comment (humphd) [reply app/queues/dns/create-record-worker.server.ts:87]: He means, don't do the loop and retry logic here, that's what a queue is for.  Fail this job, and BullMQ will retry it

**Comment by Genne23v** [comment]
> Another thought I have while reading this. We're starting to add a lot of queues. Each queue requires more memory, Redis event handlers, etc. At some point, we'll have to measure. I know in the official docs, the BullMQ talks about queues not being a huge drain.
> 
> Related to the above, I also wonder if we should use the same queue or worker for all common tasks. For example: all Route53 operations are going to get made against the same API, and share the same rate limiting. We can set a [rate limit](https://docs.bullmq.io/guide/rate-limiting#manual-rate-limit) on a worker to avoid hitting this.
> 
> From https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html, here are the Route53 API limits we need to respect:
> 
> > All Amazon Route 53 API requests
> > For the [Amazon Route 53 APIs](https://docs.aws.amazon.com/Route53/latest/APIReference/API_Operations_Amazon_Route_53.html) five requests per second per AWS account. If you submit more than five requests per second, Amazon Route 53 returns an HTTP 400 error (Bad request). The response header also includes a Code element with a value of Throttling and a Message element with a value of Rate exceeded.
> > Note
> > If your application exceeds this limit, we recommend that you implement exponential backoff for retries. For more information, see [Error Retries and Exponential Backoff in AWS](https://docs.aws.amazon.com/general/latest/gr/api-retries.html) in the Amazon Web Services General Reference.
> > ChangeResourceRecordSets requests
> > If Route 53 can't process a request before the next request arrives, it will reject subsequent requests for the same hosted zone and return an HTTP 400 error (Bad request). The response header also includes a Code element with a value of PriorRequestNotComplete and a Message element with a value of The request was rejected because Route 53 was still processing a prior request.

I will take a look at this and write an issue for follow-up. Thank you

File: app/queues/dns/create-record-worker.server.ts (hunk 16 above)
Comment (Genne23v) [reply app/queues/dns/create-record-worker.server.ts:95]: I'm not sure if this is referring to `new Worker`. This `Worker` receives the data from children, not from its own job. Do I still need to type this?

File: app/queues/dns/create-record-worker.server.ts (hunk 11 above)
Comment (Genne23v) [reply app/queues/dns/create-record-worker.server.ts:106]: It's true that my code is less maintainable. Thanks for the suggestion.

File: app/queues/dns/dns-flow.server.ts (hunk 12 above)
Comment (Genne23v) [reply app/queues/dns/dns-flow.server.ts:25]: I'm not sure what is the best approach, but my thinking was that I have `...Record` types with similar props for DB, Route53 and Job. So I was thinking this is good to modify something that is available rather than thinking about reduced scope of props with ambiguous name. I should name my new type `JobRecord`.

File: app/queues/dns/create-record-worker.server.ts (hunk 3 above)
Comment (humphd) [reply app/queues/dns/create-record-worker.server.ts:29]: @Genne23v so to confirm, you're going to also add `username` in the follow-up?  Seems like an easy win here in this PR.

File: app/queues/dns/create-record-worker.server.ts [hunk 18]
```diff
@@ -0,0 +1,130 @@
...
+  { connection: redis }
+);
+
+export const checkDnsStatusWorker = new Worker(
+  checkDnsStatusQueueName,
+  async (job) => {
+    logger.info('Checking Route53 readiness');
+
+    const values: { [jobKey: string]: string } = await job.getChildrenValues();
+    const key = Object.keys(values)[0];
```
humphd, line 83: Maybe this is OK, but it makes me nervous, as you're depending on ordering.  I haven't dug into this, but is there really not a way to type the return value of a job?  I won't block the PR on this, but we should dig into this more deeply.

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,130 @@
...
+  { connection: redis }
+);
+
+export const syncDbStatusWorker = new Worker(
+  syncDbStatusQueueName,
+  async (job) => {
+    logger.info('Sync DB status with DNS');
+
+    const childJobValues = await job.getChildrenValues();
+    if (Object.keys(childJobValues).length !== 2) {
```
humphd, line 102: You use this repeatedly, you might as well cache it and reuse:

```ts
const values = Object.keys(childJobValues);
```

package.json
```diff
@@ -20,10 +20,10 @@
     "db:studio": "prisma studio",
     "setup": "run-s db:generate db:push db:seed",
     "start": "cross-env NODE_ENV=production node ./build/server.js",
-    "start:e2e": "cross-env NODE_ENV=testing node --require dotenv/config ./build/server.js",
+    "start:e2e": " node --require dotenv/config ./build/server.js",
```
humphd, line 23: Why this change?

humphd (changes requested):
I think this looks like something we could take.  I've made a few more comments, but in general, we could start with this and improve going forward.

package.json
Eakam1007: I suggested this change ([slack conversation here](https://seneca-open-source.slack.com/archives/C04HR4WV5P0/p1677286047709169)) to get the e2e tests to work as intended. 

To summarize, I found that `testing` is not an acceptable option for `NODE_ENV`, but `test` is.
Also, the `pretest:e2e:run` script that does the build will replace all instance of `process.env.NODE_ENV` with its actual value. So, we needed to pass it there instead because it assumed production otherwise. Though there is an extra empty space on this line

humphd: Thanks for this, sounds good.  Yes, let's fix the blank line.  Question: will the `NODE_ENV` carry forward to this script from the `pre*` script, or should we also add it here?

Eakam1007: I am not sure if it matters, since the `pre*` script will replace all instances of `process.env.NODE_ENV` in the code with `test`.  From my testing, it most likely won't carry forward

app/queues/dns/create-record-worker.server.ts
dadolhay: Happy if it helped :)

app/queues/dns/create-record-worker.server.ts
dadolhay: I agree, it makes the code brittle. If you introduce an other step or swap steps or whatever this can result unexpected problems. I'd use the find method here as well if possible

app/queues/dns/create-record-worker.server.ts
dadolhay: @humphd I don't think there is a way to type data that has been returned by child job, at least not here as a ganaric

@Genne23v What I'd do is:
- Export the return type of the child job worker's JS file
- Import it into this file

In L112 - L113 do this:
```ts
    const recordId = childJobValues[addDbRecordJobKey] as SomeType;
    const status = childJobValues[checkStatusJobKey] as SomeOtherType;
```

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    const addDbRecordJobKey = Object.keys(childJobValues).find((item) =>
+      item.includes(addDbRecordQueueName)
+    );
+    const checkStatusJobKey = Object.keys(childJobValues).find((item) =>
+      item.includes(checkDnsStatusQueueName)
+    );
+    const recordId = childJobValues[addDbRecordJobKey!];
+    const status = childJobValues[checkStatusJobKey!];
```
dadolhay, line 113:
```suggestion
    if (!addDbRecordJobKey) {
        throw new Error(`Did not receive data from child job ${addDbRecordQueueName}`)
    }

    if (!checkStatusJobKey) {
        throw new Error(`Did not receive data from child job ${checkDnsStatusQueueName}`)
    }

    const recordId = childJobValues[addDbRecordJobKey] as AddDbRecordJobReturnType;
    const status = childJobValues[checkStatusJobKey] as CheckStatusJobReturnType;
```

app/queues/dns/create-record-worker.server.ts
Genne23v: I'm not sure how cache should be done. Can I do with `global`?

Genne23v: I totally misunderstood the comment. Sorry for the wrong question.

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,145 @@
...
+export const syncDbStatusQueueName = 'sync-db-status';
+
+export interface JobRecord {
+  username: Record['username'];
+  type: Record['type'];
+  name: Record['name'];
+  value: Record['value'];
+}
+
+export const addDbRecordWorker = new Worker<JobRecord>(
```
dadolhay, line 24:
```suggestion

export type JobRecordResult = Record['id'];

export const addDbRecordWorker = new Worker<JobRecord, JobRecordResult>(
```

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,145 @@
...
+        data: {
+          username,
+          type,
+          name,
+          value,
+          expiresAt: setMonthsFromNow(6),
+          status: RecordStatus.pending,
+        },
+      });
+      return result.id as Record['id'];
```
dadolhay, line 53:
```suggestion
      return result.id;
```
Not needed because of the above generic

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,145 @@
...
+      item.includes(checkDnsStatusQueueName)
+    );
+
+    if (!addDbRecordJobKey) {
+      throw new Error(`Could not receive data from child job ${addDbRecordQueueName}`);
+    }
+    if (!checkStatusJobKey) {
+      throw new Error(`Could not receive data from child job ${checkDnsStatusQueueName}`);
+    }
+    const recordId = childJobValues[addDbRecordJobKey] as Record['id'];
```
dadolhay, line 127:
```suggestion
    const recordId = childJobValues[addDbRecordJobKey] as JobRecordResult;
```

This way you can be 100% sure that what you return and what you get are always in sync type-wise, and if you change the interface in the `add db record` worker, it wil also be changed here too

Same thing goes with `status job`, I just didn't want to repeat the example

dadolhay (changes requested):

dadolhay (approved):
Looks nice!

Genne23v:
> Looks nice!

Thank you for all your valuable feedback. It really helps me a lot!

app/queues/dns/dns-flow.server.ts
```diff
@@ -0,0 +1,77 @@
...
+      failParentOnFailure: true,
+      attempts: 5,
+      backoff: {
+        type: 'exponential',
+        delay: 30_000,
+      },
+    },
+  };
+
+  return await flowProducer.add(syncDbStatus);
```
humphd, line 76: not: you don't need `async/await` in this function, and can return this `Promise` directly.

package.json
humphd, line 23: Let's re-add `cross-env NODE_ENV=test` here and remove that extra whitespace.

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,148 @@
...
+  value: Record['value'];
+}
+
+export type addDbRecordJobRecordResult = Record['id'];
+export type addDnsRecordJobRecordResult = string;
+
+export const addDbRecordWorker = new Worker<JobRecord, addDbRecordJobRecordResult>(
+  addDbRecordQueueName,
+  async (job) => {
+    logger.info('Creating a record in DB');
```
humphd, line 30: All these worker logs should be `.debug()` vs. `.info()`: this info is great to have in development and testing, but not needed in production.

app/queues/dns/create-record-worker.server.ts
```diff
@@ -0,0 +1,148 @@
...
+      throw error;
+    }
+  },
+  { connection: redis }
+);
+
+export const createDnsRecordWorker = new Worker<JobRecord, addDnsRecordJobRecordResult>(
+  createDnsRecordQueueName,
+  async (job) => {
+    logger.info('Creating a record in Route 53');
```
humphd, line 68: Same here, and all similar instances below

humphd (changes requested):
Looks good, a few small things and it's ready.

humphd (approved):

dadolhay:
> > Looks nice!
> 
> Thank you for all your valuable feedback. It really helps me a lot!

Cool, I'm happy if you find it useful :)

dadolhay (approved):

humphd:
> * Create a `hostedZone` using unit test `dns.server.test.ts`. If you run all test, it will remove `hostedZone`. Run partial test.
> * Get `hostedZoneId` from `localhost:5053/moto-api` and set `AWS_ROUTE53_HOSTED_ZONE_ID`
> * Set `ROOT_DOMAIN` to `starchart.com`

I wrote #283 to help with this logic.

SerpentBytes (approved):

app/queues/dns/dns-flow.server.ts
```diff
@@ -0,0 +1,77 @@
...
+const flowProducer = new FlowProducer({ connection: redis });
+
+export const addDnsRequest = async ({ username, type, name, value }: JobRecord) => {
+  // Step 1. Create a record in MySQL for a domain with pending status
+  const addDbRecord: FlowJob = {
+    name: `addDbRecord:${name}-${username}`,
+    queueName: addDbRecordQueueName,
+    data: { username, type, name, value } as JobRecord,
+    opts: {
+      failParentOnFailure: true,
```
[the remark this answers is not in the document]
dadolhay: Parent runs after child(s) so step 1 is a child of step 2 and so on

humphd (approved):

SerpentBytes (approved):

Issue #272, opened 2023-02-23
Record renewal in records table


Myrfion:
Just connect existing back-end functionality to from the back end and make sure it works properly

Myrfion:
Resolved in #324

Issue #205, opened 2023-02-14
Provide tooltips for domains table icons


humphd:
We have a bunch of icons in each row for a domain:

<img width="324" alt="Screenshot 2023-02-13 at 8 31 16 PM" src="https://user-images.githubusercontent.com/427398/218614966-383fd6e1-3aa5-4d67-987c-8e3d64232f24.png">

Hovering over these doesn't tell you what they are.  Let's include a tooltip or HTML title or something.

Myrfion:
Fixed in #276

PR #276, opened 2023-02-24
Feat: integrate domain delete + renew with DB + tooltips

### LINKED ISSUE #271: Record Deletion in Records Table
Just connect existing back end functionality to from end and make sure it works properly

Myrfion:
- Add integration of deleting domain
- Add integration of renewing domain (expiry date + 6 months)
- Add tooltips for action buttons
- Add loading on table row action submission

Here is what the table row in the loading state looks like: 
Screenshot 2023-02-24 at 11 13 53

Closes #271 #272 #205


humphd (commented):
This is going to require #274, since these operations require background async jobs to complete.  I wonder if we should wait to get that in, so you can call the top-level API bits of those flows?

cc @Genne23v.

app/routes/__index/domains/index.tsx
```diff
@@ -15,20 +20,51 @@ export const loader = async ({ request }: LoaderArgs) => {
...
+  const { id } = Object.fromEntries(await request.formData());
+  const formalizedID = Number(id);
+
+  if (request.method === 'PUT') {
+    const record = await getRecordById(formalizedID);
+    if (record) {
+      const newExpiresAt = new Date(record.expiresAt);
+      newExpiresAt.setMonth(newExpiresAt.getMonth() + 6);
+
+      await updateRecordById(
```
dadolhay, line 33: This signature is extremely hard to use and maintain. Please pass in an object instead. You can check the certificate and record models for an example if you would like to

app/routes/__index/domains/index.tsx
```diff
@@ -15,20 +20,51 @@ export const loader = async ({ request }: LoaderArgs) => {
...
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        newExpiresAt
+      );
+    }
+  } else if (request.method === 'DELETE') {
+    await deleteRecordById(formalizedID);
```
dadolhay, line 46: If I understand this correctly, this is now what we should do.

If you take the record and delete it from the FE, that will not actually delete the record from route53, it will only delete our knowledge of it.

Instead, we should kick off a worker, that removes the record from route53, then delete it from the db as well

dadolhay (changes requested):

app/routes/__index/domains/index.tsx
Myrfion: I feel like we don't have an api for making a proper deletion yet, I guess it makes sense to wait till some related PRs gonna be merged as like #274 as @humphd mentioned

dadolhay: To be honest, I'd just add a TODO into the code instead of this DB tie and ship this PR.
If a PR is not merged, it's constant work to resolve conflicts and keep it up to date and ready to merge while others are working on the same piece of code

Myrfion: @dadolhay makes sense, I did what you say, just left todos comments and opened issues to resolve them when the appropriate code for all these things gonna be shipped #280 #281

humphd (approved):

SerpentBytes (approved):

dadolhay (approved):

PR #279, opened 2023-02-24
Dynamically load the SSO IdP metadata at startup

humphd:
This adjusts our SAML code a bit to dynamically load the IdP's metadata at startup vs. requiring us to pass it in.  This also allows us to check that the IdP is reachable.

This data is not a secret (the metadata is always public), so we don't need to keep it in secrets.  And storing the entire file seems like a waste, so this uses the URL.  This change will make it easier to setup staging and production too.

One note about the way I'm doing this.  This code intentionally re-uses a Promise, so I haven't forgotten the `await` in my `fetch` call.  It's too bad Samlify doesn't support this use case directly.

I've put this in 0.5 since I missed code freeze for 0.4.


humphd:
I had to update CI for the e2e tests to include the login container.  I tried to run it as a service container like MySQL and Redis, but apparently you can't mount volumes when you do this, because your code isn't checked-out when the containers start.  So I use docker compose to do it, and it seems fine.

dadolhay (approved):

Eakam1007 (approved):
Looks good. I am just curious about why we reuse the promise instead of waiting for it to resolve. Does the await take too long?

humphd:
@Eakam1007 the way this code is written.  We'd have to do an async `init()` or something.

PR #283, opened 2023-02-28
Add server init logic for SAML and DNS

### LINKED ISSUE #275: Add npm script to create Route53 hosted zone for local dev
In #274 we're starting to need access to an existing Route53 hosted zone.  There is some existing [code in our DNS tests](https://github.com/DevelopingSpace/starchart/blob/main/test/unit/dns.server.test.ts#L13-L21) to do this using the [`createHostedZone()`](https://github.com/DevelopingSpace/starchart/blob/main/app/lib/dns.server.ts#L56-L72) function.

We should provide a way to accomplish this in development.  Because we already have a [`setup`](https://github.com/DevelopingSpace/starchart/blob/main/package.json#L21) script in `package.json`, I suggest we include an additional step to create a hosted zone by running some node code.

Similar to how the db seeding works, this call should be idempotent, that is, calling it over and over again should result in the same, single hosted zone being setup.

humphd:
Fixes #275.

We have a number of modules that require some kind of asynchronous setup step before the app can start.  Currently, this includes:

- SAML: we need to download the IdP's metadata XML and use it to create an `IdentityProvider` instance
- DNS: in development, we need to create a hosted zone on the mock Rout53 server

In order to facility this kind of logic, I'm adding a step before the server is started that does any `init()` that a module might need done.  All you have to do is `export` an `async` function, and we'll call it at startup.

The way that Remix hot-reloads work, I had to do a bit of gymnastics to preserve module state after a server-side reload.  The current `db.server.ts` file shows a pattern to use: define a `global` variable and stick the instance data there, then restore it when the module is reloaded.

I've also moved the SAML code into `app/lib/` like the other server modules.

To test this, do `npm run build` and `npm run dev` and you'll see something like this:

```
info: 🔌 setting up prisma client to localhost:3306
info: app initializing...
debug: SAML init: downloading IdP metadata from http://localhost:8081/simplesaml/saml2/idp/metadata.php
debug: DNS init: created hosted zone EPE35EFYJZJFNT8 for starchart.invalid
Watching Remix app in development mode...
💿 Built in 311ms
info: ✅ app ready: http://localhost:8080
```

Now the metadata is downloaded (SAML login should work) and a DNS hosted zone has been created (the DNS stuff in #274 should work in development).


app/lib/dns.server.ts
```diff
@@ -53,7 +78,7 @@ export const route53Client = new Route53Client({
   credentials: credentials(),
 });
 
-export const createHostedZone = async (domain: string) => {
+export async function createHostedZone(domain: string) {
```
Genne23v, line 81: Just for my own curiosity. Does this make a difference?

humphd: It has to do with [JS hoisting](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting) and how a `function` vs. a `const` appears in the current scope.  I need to call this function before it gets declared, so I'm relying on it being hoisted to the top of the scope.

app/lib/saml.server.ts
```diff
@@ -0,0 +1,103 @@
...
+import type { IdentityProvider } from 'samlify/types/src/entity-idp';
+
+samlify.setSchemaValidator(validator);
+
+let idp: IdentityProvider;
+
+// this is needed because in development we don't want to restart
+// the server with every change, but we want to reuse the already
+// initialized idp that we created at startup, and cached on the global.
+declare global {
```
sfrunza13, line 17: Does this ever get used apart from dev? Could it also be included under the below conditional?

sfrunza13 (approved):

app/lib/saml.server.ts
humphd: Apparently not:

> An ambient module declaration is only allowed at the top level in a file.ts (1234)

Genne23v (approved):

PR #289, opened 2023-03-01
Add dns update workflow

Genne23v:
This PR adds DNS update workflow as below.

1. Update the record by ID in DB with PENDING status
2. Upsert the record in Route53
3. Repeat polling Route53 until it gets the updated status (INSYNC or something else)
4. Update the DB either with active or error

### To test this, it needs manual steps as below.
Create a hostedZone using unit test dns.server.test.ts. If you run all test, it will remove hostedZone. Run partial test.
Get hostedZoneId from localhost:5053/moto-api and set AWS_ROUTE53_HOSTED_ZONE_ID
Set ROOT_DOMAIN to starchart.com
Go to http://localhost:8080/dev and hit Request DNS Record button


humphd (commented):
A general comment, after quickly scanning the code in this PR.  For DNS, doing Create, Update, and Delete is all very similar.  Instead of creating separate queues/workers for each of these, can we create a single set of queues/workers that take the operation, etc. to run (e.g., create vs. update) and are able to run any of the jobs like this?

I think we should be able to use/leverage the existing queues/workers you already created.

Genne23v:
> 
> I think we should be able to use/leverage the existing queues/workers you already created.

I did some testing to reuse existing worker, somehow it didn't work as expected. And the actually only `checkDnsStatus` would be identical and the rest would be little different. I will spend more time to reduce the repeated worker/queue.

humphd:
The way you'd do it is to have your worker be more elaborate, and include all the logic of each case.  You'd send a arg with the job data that tells it what to do:

For example:

```ts
// Names are made up, don't read anything into them...
export async function genericWorker(job: DnsJob) {
  const { type, ...data } = job.data;

  switch(type) {
    case 'create':
      return doCreate(data); 
    case 'update':
      return doUpdate(data); 
    case 'delete':
      return doDelete(data); 
  }
}
```

humphd (changes requested):
Can you rebase this on main so it is easier to test?  Also I created #298 to make it easier to deal with dns record creation for a user (i.e., we can just use the initial part of a name vs. the entire thing).

humphd:
@Genne23v that wasn't a rebase, you added those commits in by mistake.

Genne23v:
> @Genne23v that wasn't a rebase, you added those commits in by mistake.

Sorry, my bad. I'm fixing it now.

app/routes/dev.tsx
```diff
@@ -56,6 +57,30 @@ export const action = async ({ request }: ActionArgs) => {
...
+        name: 'osd700-a2.user1.starchart.com',
+        value: '192.168.0.2',
+        username: 'user1',
+      });
+      return json({
+        result: 'ok',
+        message: 'DNS record creation requested',
+      });
+    case 'delete-dns-record-request':
+      await deleteDnsRequest({
```
Myrfion, line 73: Is there any necessity of passing so many arguments to `deleteDnsRequest`?

Since this is deletion maybe the function could just take an id and just make a deletion by id?

app/queues/dns/dns-flow.server.ts
```diff
@@ -1,23 +1,24 @@
...
   checkDnsStatusQueueName,
   syncDbStatusQueueName,
-} from './create-record-worker.server';
+} from './dns-record-worker.server';
 import { redis } from '~/lib/redis.server';
 
 import type { FlowJob } from 'bullmq';
-import type { JobRecord } from './create-record-worker.server';
+import type { JobRecord } from './dns-record-worker.server';
+import { WorkType } from './dns-record-worker.server';
```
humphd, line 12: `import type ...`

app/queues/dns/dns-flow.server.ts
```diff
@@ -1,23 +1,24 @@
...
 
 const flowProducer = new FlowProducer({ connection: redis });
 
 export const addDnsRequest = async ({ username, type, name, value }: JobRecord) => {
   // Step 1. Create a record in MySQL for a domain with pending status
   const addDbRecord: FlowJob = {
```
humphd, line 18: Question: should we be adding the initial record to the db *before* we put this job in the queue?  I ask because the front-end is going to want to be able to query for this record immediately, and poll on it to see when it's completed (i.e., when `status` changes).

In other words, should we put the record in the db with `pending` status, then add the job to the queue and return?

app/queues/dns/dns-record-worker.server.ts
```diff
@@ -0,0 +1,211 @@
...
+  value: Record['value'];
+}
+
+export type addDbRecordJobRecordResult = Record['id'];
+export type dnsRecordUpdateJobRecordResult = string;
+
+export const dbRecordWorker = new Worker<JobRecord>(
+  dbRecordUpdateQueueName,
+  async (job) => {
+    logger.debug('DB record update in DB');
```
humphd, line 36: `DNS record update in DB`?

app/queues/dns/dns-record-worker.server.ts
```diff
@@ -0,0 +1,211 @@
...
+    const { workType, id, username, type, name, value } = job.data;
+
+    try {
+      switch (workType) {
+        case 'create':
+          const existingRecords = await prisma.record.count({
```
humphd, line 42: Is there any way we can collapse all this code into a function you call in your switch, like you do in 104-116 below?  This is pretty hard to read with it all expanded inline like this.

app/queues/dns/dns-record-worker.server.ts
```diff
@@ -0,0 +1,211 @@
...
+          return createRecord(username, type, name.toLowerCase(), value);
+
+        case 'update':
+          return upsertRecord(username, type, name.toLowerCase(), value);
+
+        case 'delete':
+          return deleteRecord(username, type, name.toLowerCase(), value);
+
+        default:
+          throw new UnrecoverableError('Invalid work type');
```
humphd, line 115: Same comment here: include the `workType` in the error message so it's easier to debug.

app/routes/dev.tsx
```diff
@@ -13,8 +13,9 @@ import { requireUser, requireUsername } from '~/session.server';
 import { addNotification } from '~/queues/notifications.server';
 import { addCertRequest } from '~/queues/certificate/certificate-flow.server';
 
+import { addDnsRequest, updateDnsRequest, deleteDnsRequest } from '~/queues/dns/dns-flow.server';
 import type { LoaderArgs, ActionArgs } from '@remix-run/node';
-import { addDnsRequest } from '~/queues/dns/dns-flow.server';
+import { WorkType } from '~/queues/dns/dns-record-worker.server';
```
humphd, line 18: This can be removed

app/routes/dev.tsx
```diff
@@ -56,6 +57,30 @@ export const action = async ({ request }: ActionArgs) => {
         result: 'ok',
         message: 'DNS record creation requested',
       });
+    case 'update-dns-record-request':
+      await updateDnsRequest({
+        id: 62,
+        type: 'A',
+        name: 'osd700-a2.user1.starchart.com',
```
humphd, line 64: I think that we should only require passing the first part of the `name`, and not the base domain, which we can get from the user (see #298).

humphd (changes requested):

app/routes/dev.tsx
Genne23v: `id` is to remove a record in DB. But I need type, name, value to remove in Route53. It doesn't use unique identifier.

humphd: It's an interesting point though: could you not do a query to the db by the `id` to get the rest?  Not critical to change it all, just thinking out loud.

Genne23v: It could remove without `id` and with type, name, value, it should be unique combination. But I'm not sure if we should do in a less reliable way.

app/queues/dns/dns-record-worker.server.ts
Genne23v: I should change it to `logger.debug('${workType} DNS record in DB');`

app/queues/dns/dns-record-worker.server.ts
Genne23v: Will be updated!

app/routes/dev.tsx
Genne23v: I think Prisma does not allow me to use delete without unique key. 

`Type '{ type: RecordType; name: string; value: string; }' is not assignable to type 'RecordWhereUniqueInput'.
  Object literal may only specify known properties, and 'type' does not exist in type 'RecordWhereUniqueInput'.ts(2322)`

app/queues/dns/dns-flow.server.ts
Genne23v: Totally makes sense. It should be working as you stated. I will think about how it should be done.

Genne23v:
@humphd @Myrfion  I just updated the PR. Sorry that I force pushed. Please note that I haven't refactored the code to separate DB update and Route53 update in this PR. I think the refactoring amount is very big for this one PR. I will quickly  push another PR once this gets merged.

app/models/record.server.ts
```diff
@@ -1,4 +1,4 @@
-import type { Record } from '@prisma/client';
+import { Record, RecordStatus } from '@prisma/client';
```
SerpentBytes, line 1: I would import Record as a type to eliminate the warning raised by ESLint.

app/routes/__index/domains/new.tsx
```diff
@@ -1,5 +1,5 @@
 import { Container, Heading, Text } from '@chakra-ui/react';
-import { RecordType } from '@prisma/client';
+import { RecordStatus, RecordType } from '@prisma/client';
```
SerpentBytes, line 2: It seems like `RecordStatus` is never being used. I would remove it, so ESLint doesn't complain.

SerpentBytes (approved):

app/models/record.server.ts
```diff
@@ -12,40 +13,68 @@ export async function getRecordById(id: Record['id']) {
...
       status,
+      expiresAt: status === RecordStatus.active ? expireToSet : undefined,
+    },
+  });
+}
+
+export async function doesRecordExist(data: Pick<Record, 'username' | 'type' | 'name' | 'value'>) {
+  const { username, type, name, value } = data;
+
+  return prisma.record.count({
```
humphd, line 72: I would return a boolean, based on the name of this.

```ts
const count = await prisma.record.count({...});
return count > 0;
```

app/queues/dns/dns-flow.server.ts
```diff
@@ -1,23 +1,26 @@
...
+import { WorkType } from './dns-record-worker.server';
 
 import type { FlowJob } from 'bullmq';
-import type { JobRecord } from './create-record-worker.server';
+import type { JobRecord } from './dns-record-worker.server';
 
 const flowProducer = new FlowProducer({ connection: redis });
 
 export const addDnsRequest = async ({ username, type, name, value }: JobRecord) => {
+  const fullRecordName = `${name}.${username}.${process.env.ROOT_DOMAIN}`;
```
humphd, line 17: Similar to what I told @dadolhay, we should probably have a common function in `app/utils.ts` or something for doing this kind of thing.  I had to write something similar in https://github.com/DevelopingSpace/starchart/blob/main/app/models/user.server.ts#L27.  One reason to centralize this is that we have to deal a `.` in a `username`.

Maybe we could do something like this:

```ts
// app/utils.ts

/**
 * Remove invalid/unwanted characters from a username. In the case
 * of faculty/admins, we will have a `.` in the username, which we
 * don't want to use as part of domain names.
 * @param username The user's username (e.g., `jsmith` or `john.smith`)
 */
function cleanUsername(username: PrismaUser['username']) {
  return username.replace(/\./g, '');
}

/**
 * Create the domain for a user, using their username
 * @param username The user's username (e.g., `jsmith` or `john.smith`)
 * @returns string jsmith -> jsmith.starchart.com
 */
function buildUserBaseDomain(username: PrismaUser['username']) {
  return `${cleanUsername(username)}.${process.env.ROOT_DOMAIN}`;
}

export function buildDomain(username: PrismaUser['username'], name?: string) {
  if(name) {
    return `${name}.${buildUserBaseDomain(username)}`
  }
  return buildUserBaseDomain(username);
}
```

app/routes/__index/domains/new.tsx
```diff
@@ -48,7 +48,7 @@ export const action = async ({ request }: ActionArgs) => {
 
   // Create a pending record... for now
```
humphd, line 49: This comment probably needs to get updated to reflect the new reality.

humphd (changes requested):
Are you going to do the switch to create/update the record in the DB when you add it to the queue vs. when the worker starts?  I think you should, so that the front-end can rely on the record existing early.

app/queues/dns/dns-record-worker.server.ts
```diff
@@ -0,0 +1,176 @@
...
+    } catch (error) {
+      logger.warn('Could not update the record in DB', error);
+      throw error;
+    }
+  },
+  { connection: redis }
+);
+
+const setMonthsFromNow = (months: number) =>
+  new Date(new Date().setMonth(new Date().getMonth() + months));
```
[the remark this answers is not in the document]
Genne23v: Will be removed

app/routes/__index/domains/new.tsx
Genne23v: I think the comment is still valid. I just move the status set from here to `/models/record.server.ts`

Genne23v:
@humphd I did a quick prototype to see how it should work and considered to add it to this PR, but the amount of refactoring is very big. So I changed my mind to do that separation in next release. If it needs to be in this iteration, I will update the PR.

humphd:
> @humphd I did a quick prototype to see how it should work and considered to add it to this PR, but the amount of refactoring is very big. So I changed my mind to do that separation in next release. If it needs to be in this iteration, I will update the PR.

That makes sense to me, but you didn't mention it in the PR.  It's good to call out plans like that, and even better to file follow-up issues to clearly communicate intent.

app/utils.ts
```diff
@@ -67,6 +67,13 @@ export function buildUserBaseDomain(username: PrismaUser['username']) {
   return `${cleanUsername(username)}.${process.env.ROOT_DOMAIN}`;
 }
 
+export function buildDomain(username: PrismaUser['username'], name?: string) {
```
humphd, line 70: You can rebase to use what @dadolhay already landed in https://github.com/DevelopingSpace/starchart/commit/6719cee3c2d0d1d5d29022c6dc04d8fa1f7322e1#diff-9ba22a4ee607f76b338e5793ffbb2eae933b1b432aba0fbcd3a3fd14c9efbb1bR66

humphd (changes requested):

Genne23v:
> 

Here's the issue number for follow-up. #315

humphd (dismissed):
Thanks for filing that issue.

Myrfion (approved):

SerpentBytes (approved):

SerpentBytes (approved):

SerpentBytes (approved):

app/models/record.server.ts
```diff
@@ -12,42 +13,71 @@ export async function getRecordById(id: Record['id']) {
...
       ports,
       expiresAt,
+    },
+  });
+}
+
+export async function updateRecordStatusById(id: Record['id'], status: Record['status']) {
+  const now = new Date();
+  const expireToSet = new Date();
+  expireToSet.setMonth(now.getMonth() + 6);
```
SerpentBytes, line 56: I would use `dayjs` to manipulate dates. See issue #311

SerpentBytes (changes requested):
I left some feedback. In issue #311, we request to replace the logic related to manipulating dates with built-in functions from `dayjs`, as it does it in fewer lines of code. While this PR is still open, could you use that library instead?

app/models/record.server.ts
Genne23v: Great suggestion! Thanks

Myrfion (approved):

SerpentBytes (approved):

PR #294, opened 2023-03-06
Tests for logging in and redirecting to index and to dev

### LINKED ISSUE #98: Create e2e auth test
Create a playwright test to authenticate as starchartdev.

sfrunza13:
Fixes #98 

This is the first few playwright tests that I made using codegen. They can be tested by running `npm run test:e2e:dev`

These two are just the first two, I would like some advice on how to proceed as it pertains to testing all of the requests in the flow, should I be doing each as there are many. For example after clicking login there is a post and two gets before landing on the next visible page which is the localhost:8081 simple SAML login. Also how will I test requests to and from localhost:8081 when in playwright.config.ts we `use` our base URL.

```
use: {
    /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
    actionTimeout: 0,
    /* Base URL to use in actions like `await page.goto('/')`. */
    baseURL: `http://localhost:${process.env.PORT}`,

    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
    trace: 'on-first-retry',
    /* Record video on first retry */
    video: 'on-first-retry',
  },
```

Please advise @humphd


humphd:
The failures here are related to you expecting things to be running on port `8080` and in CI it uses `8911`.

I would not hard-code the port in the test.  Instead, I'd have Playwright wait on the page having some title you expect (e.g., in an `h1`).  Then you don't care about the URLs, only the place where you land.

I also filed #295 to document how to use all the amazing debugging tools we have with Playwright.  It's so good!

humphd (commented):
I think you need to use https://playwright.dev/docs/api/class-page#page-wait-for-url to wait on the redirects to finish.

sfrunza13:
@humphd I'm thinking this might be what it has to do with:

```
  # SSO Identity Provider test service, https://simplesamlphp.org
  # Access to the login page available at http://localhost:8081
  login:
    image: kristophjunge/test-saml-idp
    container_name: 'login'
    ports:
      - '8081:8080'
    environment:
      - SIMPLESAMLPHP_SP_ENTITY_ID=http://host.docker.internal:8080/sp
      - SIMPLESAMLPHP_SP_ASSERTION_CONSUMER_SERVICE=http://localhost:8080/login/callback
      - SIMPLESAMLPHP_SP_SINGLE_LOGOUT_SERVICE=http://localhost:8080/logout/callback
    volumes:
      - ./config/simplesamlphp-users.php:/var/www/simplesamlphp/config/authsources.php
      - ./config/saml20-idp-hosted.php:/var/www/simplesamlphp/metadata/saml20-idp-hosted.php
```

How can I change this to make it use the testing port?

app/lib/saml.server.ts
```diff
@@ -67,13 +67,13 @@ const sp = samlify.ServiceProvider({
   assertionConsumerService: [
     {
       Binding: 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST',
-      Location: process.env.HOSTNAME + '/login/callback',
+      Location: '/login/callback',
```
humphd, line 70: Is this allowed (relative URL)?  I suspect we need an absolute, and what you had before was likely correct.

test/e2e/login.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { test, expect } from '@playwright/test';
+
+test('Login w User 1', async ({ page }) => {
```
humphd, line 3: s/w/with/

test/e2e/login.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+  await page.goto('/login?redirectTo=%2F');
+  await page.getByRole('button', { name: 'Sign In' }).click();
+  await page.getByLabel('Username').click();
+  await page.getByLabel('Username').fill('user1');
+  await page.getByLabel('Password').click();
+  await page.getByLabel('Password').fill('user1pass');
+  await page.getByRole('button', { name: 'Login' }).click();
+  await page.waitForURL(/.*\/$/);
```
humphd, line 11: Can you use something better here?  It should be `http://localhost:8080/...` right?  Using a regex like this is a bit risky.

test/e2e/login.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+  const locator = page.getByText('user1', { exact: true });
+  const starchartHeading = page.getByRole('heading', { name: 'Starchart' });
+  await expect(locator).toHaveCount(1);
```
humphd, line 14: Can you use `.toContainText()` vs. doing a count?

humphd (changes requested):

app/lib/saml.server.ts
sfrunza13: Yeah I was messing around to see if it makes a difference and everything still worked, I think this hints at a problem with what I wrote. I will change this back though.

test/e2e/login.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  await page.getByRole('button', { name: 'Login' }).click();
+  await page.waitForURL('http://localhost:8080');
+  const locator = page.getByText('user1', { exact: true });
+  const starchartHeading = page.getByRole('heading', { name: 'Starchart' });
+  await expect(locator).toContainText('user1');
+  await expect(starchartHeading).toContainText('Starchart');
+});
+
+test('Login with User 1 dev redirect', async ({ page }) => {
+  await page.goto('/login?redirectTo=%2Fdev');
```
humphd, line 17: We are likely to remove `/dev` at some point, so I'd use `/domains` or something that is more stable.

humphd (changes requested):
One small thing and this is good

test/e2e/login.spec.ts
Eakam1007: Is there no other way to get this locator? Seems redundant to get by text and then check the same text

humphd (approved):
Looks great, nice work.

test/e2e/login.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+test('Login with User 1', async ({ page }) => {
+  await page.goto('/login?redirectTo=%2F');
+  await page.getByRole('button', { name: 'Sign In' }).click();
+  await page.getByLabel('Username').fill('user1');
+  await page.getByLabel('Password').fill('user1pass');
+  await page.getByRole('button', { name: 'Login' }).click();
+  await page.waitForURL('http://localhost:8080');
+  const locator = page.getByText('user1', { exact: true });
+  const starchartHeading = page.getByRole('heading', { name: 'Starchart' });
+  await expect(locator).toContainText('user1');
```
Eakam1007, line 12: Is it possible to use some other locator for the username? 
This reads like expect(element with `this text`) to contain(`this text`)

Eakam1007 (commented):
Looks good, just a small thing

test/e2e/login.spec.ts
sfrunza13: `<p class="chakra-text cha-0">user1</p>` do you have any recommendations on how to select this? It is the user1 that shows up in the top right on the header. 

further down the same class is used on another paragraph

also do you know why the css is wonky when using the codegen script? It will sometimes look different, very frequently will come back like this:

image

humphd: Add an `id` or `class` to the React element so you can query it more easily in the test.

humphd (approved):

Eakam1007 (approved):

PR #297, opened 2023-03-07
Only add final exception logging once, increase emitter max

humphd:
This fixes 2 things:

1. Increases the max for registering event emitter listeners. We get all kinds of warnings due to how BullMQ works
2. Only registers the handlers for logging fatal errors once, which helps with 1. too

### DIFF
--- a/app/lib/logger.server.ts
+++ b/app/lib/logger.server.ts
@@ -15,16 +15,20 @@ const logger = winston.createLogger({
   transports: [new winston.transports.Console()],
 });
 
-// https://nodejs.org/api/process.html#event-uncaughtexception
-process.on('uncaughtException', (err, origin) => {
-  logger.error('uncaughtException', err, origin);
-  throw err;
-});
+export async function init() {
+  logger.debug(`logger init: adding logging for unhandled process errors`);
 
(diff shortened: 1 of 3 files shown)


humphd:
Winston isn't as flexible as Pino for logging custom objects.  I think you can only pass one object in.  What if you do this:

```ts
// Combine user and samlResponse into a single object for logging
logger.debug('user and samlResponse', { user, samlResponse });
```

sfrunza13:
> Winston isn't as flexible as Pino for logging custom objects. I think you can only pass one object in. What if you do this:
> 
> ```ts
> // Combine user and samlResponse into a single object for logging
> logger.debug('user and samlResponse', { user, samlResponse });
> ```

but then what about 

> Winston isn't as flexible as Pino for logging custom objects. I think you can only pass one object in. What if you do this:
> 
> ```ts
> // Combine user and samlResponse into a single object for logging
> logger.debug('user and samlResponse', { user, samlResponse });
> ```
This works
what about lines 23 and 29 in logger.server.ts? would the same thing happen there?

sfrunza13 (approved):

Genne23v (approved):
I don't have anything to add. It looks good to me.

PR #305, opened 2023-03-08
feat: DNS propagation wait worker

### LINKED ISSUE #169: Create queue and worker for ACME DNS propagation await
Use the static method of the Let's encrypt class (do not instantiate, as it requires the API to be hit)

This fills out step 2 of
https://github.com/Seneca-CDOT/starchart/issues/167

dadolhay:
- Alters Let's encrypt class to handle DNS error
- Alters DNS worker, enhances error logging
- Adds request to DNS worker to inject appropriate records
- Implements DNS propagation waiter worker
- Simplifies Cert order worker, simplifies response (no need for additional data)

Closes #169

### DIFF
--- a/app/lib/dns.server.ts
+++ b/app/lib/dns.server.ts
@@ -113,11 +113,23 @@ export const upsertRecord = async (
 ) => {
   try {
     if (!isNameValid(name, username)) {
-      throw new Error('Invalid name provided');
+      logger.error('Invalid record name provided', {
+        name,
+        username,
+        baseDomian: `${username}.${process.env.ROOT_DOMAIN}`,
+      });
+
+      throw new Error('Invalid record name provided');
     }
 
     if (!isValueValid(type, value)) {
-      throw new Error('Invalid value provided');
+      logger.error('Invalid record value provided', {
+        name,
+        username,
+        type,
+        value,
+      });
+      throw new Error('Invalid record value provided');
     }
 
     const command = new ChangeResourceRecordSetsCommand({
@@ -226,8 +238,8 @@ export const getChangeStatus = async (changeId: string) => {
 export const isNameValid = (name: string, username: string) => {
   const rootDomain = process.env.ROOT_DOMAIN!;
 
-  /* Domain name must end with username and root domain. 
-  Here it removes username and root domain, 
+  /* Domain name must end with username and root domain.
+  Here it removes username and root domain,
   to validate only subdomain that user has input */
   const toRemove = `.${username}.${rootDomain}`;
   if (!name.endsWith(toRemove)) {
--- a/app/lib/dns.server.ts
+++ b/app/lib/dns.server.ts
@@ -9,6 +9,7 @@ import isIP from 'validator/lib/isIP';
 
 import logger from '~/lib/logger.server';
 import secrets from '~/lib/secrets.server';
+import { buildUserBaseDomain } from '~/utils';
 
 import type {
   CreateHostedZoneResponse,
@@ -116,7 +117,7 @@ export const upsertRecord = async (
       logger.error('Invalid record name provided', {
         name,
         username,
-        baseDomian: `${username}.${process.env.ROOT_DOMAIN}`,
+        baseDomain: buildUserBaseDomain(username),
       });
 
       throw new Error('Invalid record name provided');
@@ -236,12 +237,12 @@ export const getChangeStatus = async (changeId: string) => {
 4. Domain name cannot contain multiple consecutive '-' or '_'
 5. Domain name can contain uppercase in UI but it is converted to lowercase before validation */
 export const isNameValid = (name: string, username: string) => {
-  const rootDomain = process.env.ROOT_DOMAIN!;
+  const baseDomain = buildUserBaseDomain(username);
 
   /* Domain name must end with username and root domain.
   Here it removes username and root domain,
   to validate only subdomain that user has input */
-  const toRemove = `.${username}.${rootDomain}`;
+  const toRemove = `.${baseDomain}`;
   if (!name.endsWith(toRemove)) {
     return false;
   }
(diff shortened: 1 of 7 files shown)


app/utils.ts
```diff
@@ -63,3 +64,29 @@ export function useUser(): User {
...
+}: {
+  queueName: string;
+  job: Job;
+}): Promise<{
+  [jobKey: string]: CT;
+}> {
+  const childrenValues = await job.getChildrenValues();
+  const filteredChildrenValues = Object.keys(childrenValues)
+    .filter((jobKey) => jobKey.includes(queueName))
+    .reduce((acc, jobKey) => ({ ...acc, [jobKey]: childrenValues[jobKey] }), {});
```
Genne23v, line 89: I'm not sure if I understand correctly. Is this assuming that multiple values can be returned?

app/lib/dns.server.ts
```diff
@@ -113,11 +113,23 @@ export const upsertRecord = async (
 ) => {
   try {
     if (!isNameValid(name, username)) {
-      throw new Error('Invalid name provided');
+      logger.error('Invalid record name provided', {
```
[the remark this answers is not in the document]
dadolhay: This should be unexpected, as the FE should only send correct domains

app/utils.ts
dadolhay: In theoriy, you could have multiple of the same child processes, so we are selecting all of those that match, later we can get the [0] element if there has to be a single one

app/utils.ts
```diff
@@ -63,3 +64,29 @@ export function useUser(): User {
...
+ */
+
+export async function getChildrenValuesOfQueueName<CT>({
+  queueName,
+  job,
+}: {
+  queueName: string;
+  job: Job;
+}): Promise<{
+  [jobKey: string]: CT;
```
SerpentBytes, line 84: Just curious about what the data type name CT is implying here. Is it related to certificates?

app/lib/dns.server.ts
```diff
@@ -113,11 +113,23 @@ export const upsertRecord = async (
...
+        name,
+        username,
+        baseDomian: `${username}.${process.env.ROOT_DOMAIN}`,
```
SerpentBytes, line 119: There is a small typo in spelling here; I wonder if it will be a problem now or later.

SerpentBytes (dismissed):

app/queues/certificate/order-creator-worker.server.ts
```diff
@@ -14,28 +15,34 @@ export interface OrderCreatorData {
...
+}) => {
   const challengeInsertPromises = bundles.map(async ({ domain, value: challengeKey }) => {
     logger.info(`Adding challenge to DNS`, { domain, challengeKey });
 
     /**
-     * TODO actually add challenge to DNS
+     * add challenge to DNS
      */
+    await addDnsRequest({
+      username,
```
Genne23v, line 41: @dadolhay  I'm going to break down this function into 1. Add a record to DB with pending status 2. Route53 update and update DB status once Route53 work is complete. Let me know if you see any concern with this.

app/lib/dns.server.ts
dadolhay: Nice catch!

No, it's just logging, I'll fix it in the next PR

app/queues/certificate/order-creator-worker.server.ts
dadolhay: @Genne23v 
It's still going to be a flow, right? so I create a job to handle this, and the job flow adds the db entry, handles the route53 api, then updates the status to complete

Am I understanding you correctly?

app/lib/dns.server.ts
humphd, line 119: We should leverage the function I wrote in https://github.com/DevelopingSpace/starchart/blob/main/app/models/user.server.ts#L22-L29 to do this, since we need to remove the `.` from a faculty name.  We could move that function to `app/utils.ts` and you could call it from there?

app/queues/certificate/dns-waiter-worker.server.ts
```diff
@@ -9,11 +15,65 @@ export interface DnsWaiterData {
...
+    logger.info('Checking challenges in DNS', {
+      rootDomain,
+      certificateId: orderCreatorRetval.certificateId,
+    });
+
+    const challenges = await challengeModel.getChallengesByCertificateId(
+      orderCreatorRetval.certificateId
+    );
+
+    // Using a for .. of loop here instear of forEach to be able to await
```
humphd, line 48: instead (typo)

app/queues/certificate/order-creator-worker.server.ts
```diff
@@ -14,28 +15,34 @@ export interface OrderCreatorData {
...
+      type: 'A',
```
humphd, line 42: Should this be `TXT`?

app/routes/dev.tsx
```diff
@@ -29,7 +29,7 @@ export const action = async ({ request }: ActionArgs) => {
       // Because of the foreign key constraint, it needs to be an existing user
       await addCertRequest({
         username: 'starchartdev',
-        rootDomain: 'testing.starchartdev.starchart.com',
+        rootDomain: 'testing.user1.starchart.com',
```
humphd, line 32: We should probably start using the actual username (variable), since we have multiple users in the system.

humphd (changes requested):
Looks good.  A few small things and questions.

app/queues/certificate/order-creator-worker.server.ts
dadolhay: Yep, correcting

app/lib/dns.server.ts
dadolhay: Sure, done!

app/queues/certificate/order-creator-worker.server.ts
Genne23v: Initial DB record creation won't be a part of flow as it completes the work immediately. Then the rest of jobs will be run as flow.

SerpentBytes (approved):

humphd (approved):

Issue #290, opened 2023-03-05
Sync Seneca SSO SAML claims with our Prisma/Auth code


humphd:
In order to ship the app to staging (#43) we need to sync our Prisma schema and auth code to be ready to use the [claims](https://learn.microsoft.com/en-us/windows-server/identity/ad-fs/technical-reference/the-role-of-claims) that Seneca is going to provide (see [metadata.xml](https://login.microsoftonline.com/eb34f74a-58e7-4a8b-9e59-433e4c412757/federationmetadata/2007-06/federationmetadata.xml?appid=9b6e9159-c5ab-462b-8efa-3ecb46e8b6df), from which I've extracted the relevant claims below).

Currently we use the [following claims](https://github.com/DevelopingSpace/starchart/blob/main/app/routes/login/callback.tsx#L58-L68):

1. `sAMAccountName`: the username (e.g., `jsmith`, the portion before the `@` in the user's full email address)
2. `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname`: the first name (e.g., `John`)
3. `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname`: the last name (e.g., `Smith`)
4. `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress`: the email address (e.g., `jsmith@myseneca.ca`)

This is good, but we need to adjust in order make use of an additional claim, `http://schemas.microsoft.com/ws/2008/06/identity/claims/groups`.  The `groups` claim tells us whether the user is member of the following groups:

1. Students: `group=mycustomdomain-dev-students` (staging) or `mycustomdomain-students` (prod)
2. Faculty: `group=mycustomdomain-dev-faculty` (staging) or `mycustomdomain-faculty` (prod)
3. Admins: `group=mycustomdomain-dev-admins` (staging) or `mycustomdomain-admins` (prod)

I'm not 100% clear on how the data will be sent to us.  Typically, `groups` (plural) means you can have more than one (array vs string?).  However, I suspect our users will only be in a single group at once, though the data may arrive as an array.

We need to:

1. Figure out if storing a single `group` string is OK, or whether we need to include multiple rows per user
2. Update the Prisma schema
3. Update the Prisma seed code to include various users in different groups
4. Update the simplesaml config to add more claims
5. Add some code to determine if a user `isStudent()`, `isFaculty()`, or `isAdmin()
6. Anything else I'm missing

[...]

PR #307, opened 2023-03-08
Add groups in User model

cychu42:
This is for database part of issue #290
This is meant to have `groups` as a column to store role(s)/group(s) they have, as well as related changes.

## Changes:
* Add groups in `User` model as String in `prisma\schema.prisma`
* Add examples in `prisma\seed.ts`
* Modify `createUser()` in `app\models\user.server.ts` accordingly
* Add `isStudent()`, `isFaculty()`, and `isAdmin()` in `app\models\user.server.ts`.
* Combine `firstName` and `lastName` into `displayName`
* Modify arguments of `createUser()` in `app\routes\login\callback.tsx`

`isStudent()`, `isFaculty()`, and `isAdmin()` accept `username` as parameter. They would return `true` if they have the role/group, `false` if not.

## Testing:
You make a local copy of the project branch. If you follow the contribution document's steps to set it up, you can then check whether the schema and seeding work, via `npm run db:studio`.
You can use the functions `isStudent()`, `isFaculty()`, and `isAdmin()` to test them in your code, or test in a [query file](https://github.com/DevelopingSpace/starchart/wiki/Prisma#3-write-your-first-query).

### DIFF
--- a/app/models/user.server.ts
+++ b/app/models/user.server.ts
@@ -50,17 +50,35 @@ export async function checkUsernameExists(username: PrismaUser['username']) {
 
 export async function createUser(
   username: PrismaUser['username'],
-  firstName: PrismaUser['firstName'],
-  lastName: PrismaUser['lastName'],
-  email: PrismaUser['email']
+  displayName: PrismaUser['displayName'],
+  email: PrismaUser['email'],
+  group: PrismaUser['group']
 ) {
   logger.info(`Creating new user ${username}`);
   return prisma.user.create({
     data: {
       username,
-      firstName,
-      lastName,
+      displayName,
       email,
+      group,
     },
   });
+}
+
+export async function isStudent(username: PrismaUser['username']) {
+  const { group } = await prisma.user.findUniqueOrThrow({ where: { username } });
+  // The group will have -dev in it on staging but not on prod
+  return /mycustomdomain(-dev)?-students/.test(group);
+}
+
+export async function isFaculty(username: PrismaUser['username']) {
+  const { group } = await prisma.user.findUniqueOrThrow({ where: { username } });
+  // The group will have -dev in it on staging but not on prod
+  return /mycustomdomain(-dev)?-faculty/.test(group);
+}
+
+export async function isAdmin(username: PrismaUser['username']) {
+  const { group } = await prisma.user.findUniqueOrThrow({ where: { username } });
+  // The group will have -dev in it on staging but not on prod
+  return /mycustomdomain(-dev)?-admins/.test(group);
 }
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -8,13 +8,12 @@ generator client {
 }
 
 model User {
-  username  String @id
-  firstName String
-  lastName  String
-  email     String @unique
-
-  createdAt   DateTime     @default(now())
-  updatedAt   DateTime     @updatedAt
+  username    String        @id
+  displayName String
+  email       String        @unique
+  group       String
+  createdAt   DateTime      @default(now())
+  updatedAt   DateTime      @updatedAt
   record      Record[]
   certificate Certificate[]
 }
--- a/prisma/seed.ts
+++ b/prisma/seed.ts
@@ -5,9 +5,9 @@ const prisma = new PrismaClient();
 
 async function seed() {
   const username = 'starchartdev';
-  const firstName = 'Starchart';
-  const lastName = 'Developer';
+  const displayName = 'Starchart Developer';
   const email = 'dev@starchart.invalid';
+  const group = 'mycustomdomain-admins';
   let recordExpDate = new Date();
   recordExpDate.setMonth(recordExpDate.getMonth() + 6); // a record expires after 6 months
   let certExpDate = new Date();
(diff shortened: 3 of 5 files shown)


app/models/user.server.ts
```diff
@@ -52,7 +52,8 @@ export async function createUser(
   username: PrismaUser['username'],
   firstName: PrismaUser['firstName'],
```
humphd, line 53: Drop `firstName` and `lastName` and use `displayName` or just `name` if you want (both parts are given to us in 1 single name).

app/models/user.server.ts
```diff
@@ -52,7 +52,8 @@ export async function createUser(
...
   lastName: PrismaUser['lastName'],
-  email: PrismaUser['email']
+  email: PrismaUser['email'],
+  groups: PrismaUser['groups']
```
humphd, line 56: Change this to `group` (singular)

app/models/user.server.ts
```diff
@@ -61,6 +62,51 @@ export async function createUser(
...
       lastName,
       email,
+      groups,
     },
   });
 }
+
+export async function isStudent(username: PrismaUser['username']) {
+  let result;
+  prisma.user.findUnique({ where: { username } }).then((user) => {
```
humphd, line 72:
```ts
const { group } = await prisma.user.findUniqueOrThrow({ where: { username } });
// The group will have -dev in it on staging but not on prod
return /mycustomdomain(-dev)?-students/.test(group);
```

app/models/user.server.ts
```diff
@@ -61,6 +62,51 @@ export async function createUser(
...
+      result = true;
+    } else {
+      result = false;
+    }
+  });
+  return result;
+}
+
+export async function isFaculty(username: PrismaUser['username']) {
+  let result;
```
humphd, line 85: See pattern I wrote above...

app/models/user.server.ts
```diff
@@ -61,6 +62,51 @@ export async function createUser(
...
+      result = true;
+    } else {
+      result = false;
+    }
+  });
+
+  return result;
+}
+
+export async function isAdmin(username: PrismaUser['username']) {
```
humphd, line 99: See pattern I wrote above...

app/models/user.server.ts
```diff
@@ -61,6 +62,51 @@ export async function createUser(
...
+  let result;
+  prisma.user.findUnique({ where: { username } }).then((user) => {
+    if (!user) {
+      // Leave it undefined
+    } else if (user.groups.includes('mycustomdomain-Admins')) {
```
humphd, line 104: lower case 'a' in admins

prisma/seed.ts
```diff
@@ -26,9 +27,29 @@ async function seed() {
...
+        lastName: 'Galilei',
+        email: 'user2@myseneca.ca',
+        groups: 'mycustomdomain-students',
+      },
+      {
+        username: 'user3',
+        firstName: 'Adam',
+        lastName: 'Mason',
+        email: 'user3@myseneca.ca',
+        groups: 'mycustomdomain-faculty, mycustomdomain-admins',
```
humphd, line 51: It will only ever be a single string

humphd (changes requested):
See my comment in https://github.com/DevelopingSpace/starchart/issues/290#issuecomment-1460882772

SerpentBytes (changes requested):
Type-checks seem to fail because fewer than expected arguments are being passed to a function call in `app/routes/login/callback.tsx`. Please take a look at the failing tests for more details.

app/models/user.server.ts
cychu42: Oh, I see what you meant by staging vs prod in that issue.

prisma/seed.ts
cychu42: Are you saying an user will only have one group?

sfrunza13: This is also what I understood from our conversations yesterday. I think users will only ever have 1 group they belong to.

cychu42:
This piece of code is failing in typcheck, due to changes to `createUser()`'s parameter for this PR.
```
   if (!(await checkUsernameExists(username))) {
    await createUser(
      username,
      samlResponse.attributes['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname'],
      samlResponse.attributes['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname'],
      samlResponse.attributes['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress']
    );
  }
```

app/models/user.server.ts
cychu42: Got it!

app/models/user.server.ts
cychu42: Ah, good catch.

cychu42:
I modified those arguments to what Stefan will be using after he implements changes to IDP side.

SerpentBytes (dismissed):

sfrunza13 (approved):

prisma/schema.prisma
```diff
@@ -8,11 +8,10 @@ generator client {
...
 model User {
-  username  String @id
-  firstName String
-  lastName  String
-  email     String @unique
-
+  username    String       @id
+  displayName String
+  email       String       @unique
+  group      String
```
humphd, line 14: Please from `prisma format` to clean up this formatting.

prisma/seed.ts
```diff
@@ -23,12 +23,28 @@ async function seed() {
...
+      },
+      {
+        username: 'user2',
+        displayName: 'Galileo Galilei',
+        email: 'user2@myseneca.ca',
+        group: 'mycustomdomain-students',
+      },
+      {
+        username: 'user3',
+        displayName: 'Adam Mason',
```
humphd, line 44: cc @sfrunza13 - we better keep this seed list and the config for simplesaml and our docs in sync.

humphd (changes requested):
Nice work.  One nit.

sfrunza13:
Once this goes in I will pull it and rebase my feature branch then make my PR right after this.

prisma/schema.prisma
cychu42: Thanks. Done!

cychu42:
OK. I formatted the schema file.

sfrunza13 (approved):

SerpentBytes (approved):

prisma/seed.ts
humphd: Fixed in #312

humphd (approved):
👍

cychu42:
Thank you all for the reviews.

Issue #292, opened 2023-03-06
Access Denied / Login Failed Page


sfrunza13:
I would like to have a front end representation of a failed login. We made it to the callback but did not succeed in creating a user / session so then I need to redirect to a problem.

```
  // Try and extract the username and see if there is an existing user by that name
  if (!samlResponse.attributes.sAMAccountName) {
    // TODO: Make this redirect to access denied page
    return redirect('/');
  }
```

The redirect here would be to the new route that would display an error message for the user.

Ririio:
How do I recreate this process on my local repo?

sfrunza13:
```
'broken:ithurts' => array(
        'uid' => array('2'),
        'eduPersonAffiliation' => array('group2'),
        'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress' => 'broken.heart@myseneca.ca',
        'email' => 'hansolo@myseneca.ca',
        'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname' => 'broken',
        'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname' => 'heart',
        'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name'=> 'broken.heart@myseneca.ca',
        'http://schemas.microsoft.com/identity/claims/displayname' => 'Broken Heart',
        
    ),
```

Add this to the users config, it does not have a samaccountname and will hit exactly that redirect

sfrunza13:
The saml response comes from the IDP as part of one of the bindings, it's posted to us at this callback route. The following function on the callback is what is responsible to parsing the body and returning the samlResponse object that we later work with:
`const { samlResponse, relayState } = await parseLoginResponse(body);`

sfrunza13:
Something similar to that would be fine, I just want a page that displays to the user that something went wrong when signing in. Keeping it vague is ok.

Issue #204, opened 2023-02-14
Implement landing-page experience


humphd:
When a user logs in, we take them to an empty page at `/`.  A user essentially has `/domains` and `/certificate`.  What should we do at `/`?

We could provide a landing page that does something similar to Amazon Cognito (i.e., provide some text, and a "cards" that take you to "Domains" and "Certificate"):

![image](https://user-images.githubusercontent.com/427398/218614543-332747a3-9416-4aeb-adea-b862d588f0c5.png)

We could also drop you directly into your Domains page.  We need to think about it.

Ririio:
Just wondering. The header still appears at this page right?

Eakam1007:
Yup, I am pretty sure it should

Ririio:
I like the idea of providing information within this page. Either as a small boxes referring to what "certificate" and "domain" does? We can have a "simplify" version of what they are, and split it across a "next" and "previous" functions. If they want a much deeper understanding we send them to a link with a much more robust information?

Ririio:
What information are we suppose to provide within the card that you mentioned @humphd?

humphd:
@Ririio you could include the icons and titles for DNS Records and Certificates, and below, provide a sentence or two about what each one is for.

Ririio:
You're referring to icons like these right?

![image](https://user-images.githubusercontent.com/32577022/223574278-b01adb3e-f1be-4610-a216-9339f3939a60.png)

humphd:
The current icons you use in the header.

PR #310, opened 2023-03-09
Landing Page UI | Access Denied/Login Failed Page

Ririio:
For Issue: #292 #204 

- This pull request added a new UI for the landing page, with two new cards, website name and description.
- I tweaked the header to now have "absolute" position. This is to prevent overflowing when developers want to set the height to "full"
- The previous "Loading" file, has now change to "DisplayPage"
  - This is to have a more global message page
- I remove the name of the website on the header, and moved it to the landing page instead
  - The icon feels much cleaner compared to a text with aligning
- Domain and Certficate cards displays a brief description of what they are
  - A button for the "information" page
  - A button to redirect them to the main page
  
Demo #292 

https://user-images.githubusercontent.com/32577022/224070226-166a625b-8be2-42c0-a2f4-caa1f9d7e40f.mp4

Demo #204 

https://user-images.githubusercontent.com/32577022/224070287-91ecbc92-5e82-43a0-af8e-e3a3ee03cc91.mp4


app/routes/login.tsx
```diff
@@ -16,6 +16,8 @@ import { Form } from '@remix-run/react';
 
 import { getUsername } from '~/session.server';
 import { createLoginRequest } from '~/lib/saml.server';
+import DisplayPage from '~/components/display-page';
+import undraw from '~/assets/undraw_server_down_s-4-lk.svg';
```
SerpentBytes, line 20: This is another unused import

SerpentBytes (changes requested):

app/components/header.tsx
```diff
@@ -32,24 +31,36 @@ export default function Header() {
...
-            </Flex>
-          </Link>
+        <HStack color="white" justifyContent="flex-start" gap="10">
+          <Flex justifySelf="flex-start">
+            <IconButton
+              as={Link}
+              to={{ pathname: '/' }}
+              aria-label="Redirect to main"
+              _hover={{ backgroundColor: 'transparent' }}
+              icon={<LockIcon color="gray.600" />}
```
Eakam1007, line 46: Instead of adding another icon as a link to home, we could keep `My.Custom.Domain` in the header, and add a link to that. So clicking on it would take you to `/` instead

Eakam1007 (changes requested):

app/components/header.tsx
Ririio: I'm refraining from using text on header unless it's on the edges is due to alignment issues. I noticed that it never stays in the middle on different scale

Eakam1007: Yeah, but this will also remove it from the other pages. I think we can create another issue to fix the alignment for the header for different viewports

app/components/header.tsx
```diff
@@ -80,10 +91,14 @@ export default function Header() {
...
+        <Link to={{ pathname: '/' }}>
+          <Flex justifySelf="flex-start">
+            <LockIcon color="gray.600" />
+          </Flex>
+        </Link>
+      </Show>
 
-      <Heading as="h1" size={{ base: 'md', xs: 'lg', sm: 'xl' }} color="white">
-        My.Custom.Domain
-      </Heading>
```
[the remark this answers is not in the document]
Ririio: Sure... you made me realize it's better being up there

app/components/landing-page/landing-page-card.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <Heading size="lg" color="brand.500">
+        {cardName}
+      </Heading>
+
+      <Flex width={{ md: 'xs' }} height={{ md: 'xl' }}>
+        <Text fontSize={{ base: 'sm', md: 'md' }}>{cardDescription}</Text>
+      </Flex>
+      <ButtonGroup size={{ base: 'sm', md: 'md' }} gap="5">
+        <Button as={Link} to={{ pathname: `${path}/instructions` }}>
+          Information
```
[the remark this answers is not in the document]
Eakam1007: I think `Instructions` would be fine since the path itself is `/instructions`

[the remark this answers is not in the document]
Ririio: Alrighty

app/components/header.tsx
```diff
@@ -91,15 +82,13 @@ export default function Header() {
...
-          <Flex justifySelf="flex-start">
-            <LockIcon color="gray.600" />
-          </Flex>
-        </Link>
-      </Show>
+      <Show below="lg"></Show>
+
+      <Heading as="h1" size="xl" color="white" width="100%" textAlign="center">
+        My.Custom.Domain
+      </Heading>
```
Eakam1007: I think you can now link this to `/`

Ririio: oh yeah I have. It also has a functionality that I can assign it `as={Link}` and `to={{pathname: '/'}}`, but  I couldn't make it so the test doesn't fail

SerpentBytes (dismissed):

app/components/header.tsx
Eakam1007: I tried running this but it does not link to anything

Ririio: oh you caught into it, I forgot to add the the Link component

humphd (changes requested):
This is looking great.

One thing about the 2 buttons in the cards.  I think we should consider making the entire card clickable, and not have any buttons at all.  I'd move the "Instructions" to the individual pages and include it as a link in the intro paragraphs on those pages (i.e., `to learn more, see these instructions...`).

Ririio:
> One thing about the 2 buttons in the cards. I think we should consider making the entire card clickable, and not have any buttons at all. I'd move the "Instructions" to the individual pages and include it as a link in the intro paragraphs on those pages (i.e., `to learn more, see these instructions...`).

That was my first iteration of the card actually, I spent a few hours trying to make it look better, but having a title and desc just doesn't work unless I add effects. After a while I decided to go with buttons to make it not empty

Eakam1007:
I think you can wrap the entire card in a `Link` Chakra UI component

Ririio:
> I think you can wrap the entire card in a `Link` Chakra UI component

I did create the first iteration as a clickable card, but I decided against it due to not having enough values to justify its existance

app/components/header.tsx
```diff
@@ -26,65 +26,73 @@ export default function Header() {
...
-      height="auto"
       width="100%"
       alignItems="center"
       shadow="2xl"
       fontSize="18"
       paddingY="2"
-      paddingX={{ base: '1', md: '10' }}
+      paddingX={{ base: '1', md: '5' }}
       justifyContent="space-between"
+      position="absolute"
```
Eakam1007, line 36: Position absolute causes empty space on the domain creation page:
image
Maybe something else can be used to get the same result?

Eakam1007 (changes requested):

app/components/header.tsx
Eakam1007: Could you please give an example of the overflow you fixed by setting `position: absolute` ?

Ririio: That's actually due to adding a marginTop on the container which is the header is located on the route. By adding the absolute it basically makes the header as part of the container instead of being outside of it.

Eakam1007: Sorry, could you please show which marginTop you are talking about here?

Ririio: image

Add a paddingY on the container, and that should solve the issue

Eakam1007: Removing margin top hides the title (`Create new DNS Record`):
image

I believe this is because `position: absolute` makes it so that the header is no longer considered when rendering other elements. Essentially, it allows the header to overlap other elements. For example, on the domains table page:
With `position: absolute`:
image

Without `position: absolute`:
image

The content is pushed up because the header no longer takes that space when the page is rendered

Eakam1007: Is there a use case for `height: 100%` on the containers? Shouldn't `height: auto` work as well?

Ririio: I guess currently we don't really have any use of setting height to 100%, sure I'll remove it

app/components/certificate/description.tsx
```diff
@@ -17,7 +17,7 @@ export default function DescriptionSection({
     <Flex
       flexDirection="column"
       gap="3"
-      marginTop={{ base: '5', md: '14' }}
+      marginTop="5"
```
Eakam1007, line 20: Did you mean to include this?

Ririio: yep I noticed there's no need to have a large margin there so I went with the base as default now

app/components/landing-page/landing-page-card.tsx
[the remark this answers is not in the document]
Eakam1007: So, did you end up not changing this?

Eakam1007 (approved):
You could change the button text from `Information` to something else if you like but looks good to me

humphd:
> > One thing about the 2 buttons in the cards. I think we should consider making the entire card clickable, and not have any buttons at all. I'd move the "Instructions" to the individual pages and include it as a link in the intro paragraphs on those pages (i.e., `to learn more, see these instructions...`).
> 
> That was my first iteration of the card actually, I spent a few hours trying to make it look better, but having a title and desc just doesn't work unless I add effects. After a while I decided to go with buttons to make it not empty

OK.  I'd still move the "instructions" bit to the page itself, and only have a single call-to-action button for each card.

Eakam1007 (changes requested):
A few things:
- You should remove the unused `ButtonGroup` import in `langing-page-card.tsx` Line 1
- Would it be better if the button took the entire width of the card?
  image
- Also, you should add a `Link` to the instructions page in the `card description`

Eakam1007 (approved):

dadolhay:
@Ririio
Maybe I'm just missing it, but what is the license for that SVG image? Where is it from, what conditions is it come with? IMO we should document this (maybe a file next to the svg or something)

Ririio:
> @Ririio Maybe I'm just missing it, but what is the license for that SVG image? Where is it from, what conditions is it come with? IMO we should document this (maybe a file next to the svg or something)

https://github.com/DevelopingSpace/starchart/blob/main/app/assets/LICENSE

The license is under here

dadolhay (approved):

dadolhay:
Thanks, approved!

PR #312, opened 2023-03-09
New user info in IDP

sfrunza13:
to do with #307 and #290 

This changes the claims that the IDP sends us to better match what staging and prod will do and adds a user that exists in seed now that did not exist in the IDP previously. Also changes persistent name id to sAMAccountName.


config/simplesamlphp-users.php
```diff
@@ -19,36 +19,34 @@
...
+            'displayname' => 'Galileo Galilei',
             'email' => 'user2@myseneca.ca',
-            'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname' => 'Galileo',
-            'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname' => 'Galilei',
-            'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name'=> 'user2@myseneca.ca',
             'sAMAccountName'=> 'user2',
-            'http://schemas.microsoft.com/identity/claims/displayname' => 'Galileo Galilei',
+            'group'=>'mycustomdomain-faculty',
         ),
+        'user3:user3pass' => array(
```
humphd, line 35: Please add this user/password to the Markdown file where we keep track of all user accounts for testing/dev.

humphd (changes requested):

humphd (approved):

config/simplesamlphp-users.php
```diff
@@ -13,42 +13,32 @@
...
         'exampleauth:UserPass',
         'user1:user1pass' => array(
-            'uid' => array('1'),
-            'eduPersonAffiliation' => array('group1'),
             /*
               NOTE: we add a bunch of claims that we expect to get back from Seneca,
               and need to simulate here.
             */
-            'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress' => 'user1@myseneca.ca',
+            'displayname' => 'Johannes Kepler',
```
Ririio, line 20: Is the display name the value we'll be displaying on the header?

sfrunza13: Currently we are showing sAMAccountName but we can and maybe should show display name instead, I don't know really.

Ririio: Also one more thing, what were the purpose of the schema links you removed?

Ririio (approved):

Issue #277, opened 2023-02-24
Add validation for new DNS Record form


Eakam1007:
Based on https://github.com/DevelopingSpace/starchart/pull/243#discussion_r1117075338, we could add different validation for the DNS Record form fields based on what `type` of DNS Record it is.

Alongside this, we should also show `.{username}.domain.com` on the right side of the `Domain Name` field. This can be accomplished using [`<InputRightAddon>`](https://chakra-ui.com/docs/components/input#left-and-right-addons)

humphd:
This is going to require us to pass down the `ROOT_DOMAIN` from the back-end (`env`) to the front-end, probably via a loader.  Another option that might be better: in the back-end, when we create the user in the session, we could add a `.domain` property, which would be `{username}.{root_domain}` and then send that along in the loader that the root calls, so it's always available to the front-end code.

Ririio:
I suggest we specify minimum characters for each form field to prevent the user from being able to create domains with a single character. We can add a regex for `Ports` form to check if the user has provided with numbers that are comma separated. We can also enforce a format for course that is similar to what seneca does i.e., 3-char followed by 3-nums

humphd:
I wouldn't bother doing format checks on the optional fields.  They are purely informational. Let's not hard-code in some logic for what a course code looks like, and then have some bug later where people can't enter their course.  For ports, course, etc, we don't care what they enter.

Ririio:
From what I can see it just doesn't seem like there's point having those optional fields... If we're keeping them for informational purposes, at minimum it's a good idea to have a standard format students has to follow when adding them. 

Maybe in the future we can provide faculty members the ability to filter out the table to only display dns records with course code "web322" as an example. Should make it a lot easier for viewing students websites for the course.

PR #313, opened 2023-03-10
Add new DNS Record form tests

### LINKED ISSUE #278: Add end to end tests for DNS Record creation form
As a follow up to #243, we need to add tests for the DNS Record creation form. 
Tests could not be added as part of the PR because of [a bug in the sign up process that happens in the playwright test environment](https://github.com/DevelopingSpace/starchart/pull/243#issuecomment-1443548631). This would need to be investigated and fixed or an alternate process could be used to mock a signed in user (maybe setting the right cookies?)

These should test the following:
- User must be logged in to access the form
- Field validations (also see #277)
- Form submit is successful (for now this is a redirect to the edit page for the newly created DNS Record)

### LINKED ISSUE #296: Setup shared session for end to end tests
We need a use a shared authenticated state for end to end tests.

More details here: https://playwright.dev/docs/auth

Eakam1007:
Resolves #278, resolves #296 

Since writing the tests relied on setting up the shared auth session, these have been combined into one PR. The following changes have been made:
- Added auth setup file for e2e tests to login as a user and save session data in a JSON file
- Updated playwright config to add a setup project which uses the auth setup file. Added this setup project as a dependency to all other projects. This means the setup will run before the tests are run
- Added above JSON file to `.gitignore`. This file is generated before the tests are run so it is not necessary to save it
- Added tests for new DNS Record form. Here the saved session generated by the setup is used for the authenticated tests
- Updated Mobile Safari config as the tests pass locally but fail on CI, as per comment below

### Steps to test
You can check the CI to confirm that the tests pass. Alternatively,
- Ensure that app is setup according to [setup instructions](https://github.com/DevelopingSpace/starchart/blob/main/CONTRIBUTING.md#development)
- Run `npm run docker`
- Run `npm run build`
- Run `npm run test:e2e:run`
- Check that tests added in this PR pass. **Note**: Some of the tests for Login will fail locally. This is not related to this PR, and can be investigated in another issue


Eakam1007:
Hmmm, not quite sure why tests only fail on `Mobile Safari` on CI. They pass locally

Eakam1007:
CI issue with Mobile Safari seems to be related to this open issue in playwright.
So, using the workaround specified in issue discussion for now: https://www.github.com/microsoft/playwright/issues/11812#issuecomment-1462829766

test/e2e/new-dns-record.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+test.describe('not authenticated', () => {
+  test('redirects to login page', async ({ page }) => {
+    await page.goto('/');
+    await expect(page).toHaveURL(/.*login.*/);
+  });
+});
+
+test.describe('authenticated as user', () => {
+  // Use shared state for being logged in as user
+  test.use({ storageState: 'test/e2e/.auth/user.json' });
```
humphd, line 12: Because every "already logged in" test is going to have to do this, should we create `test/e2e/utils.ts` and make a function, `alreadyLoggedIn()` (or something) that does this for you?

Also, a question: does this need to go in a `beforeAll()` so the order is never an issue?

test/e2e/new-dns-record.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    await page.getByLabel('Course').fill('test course');
+    await page.getByLabel('Description').fill('test description');
+    await page.getByRole('button', { name: 'Create' }).click();
+    await expect(page).toHaveURL(/.*domains\/[0-9]*/);
+  });
+
+  test('does not create dns record if required fields are empty', async ({ page }) => {
+    await page.goto('/domains/new');
+    await page.getByRole('button', { name: 'Create' }).click();
+    await expect(page).toHaveURL(/.*domains\/new/);
```
humphd, line 38: Do we want to check for an error message in the page?

test/e2e/new-dns-record.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  test('redirects to edit dns record page when all fields are filled', async ({ page }) => {
+    await page.goto('/domains/new');
+    await page.getByLabel('Record Name*').fill('test');
+    await page.getByRole('combobox', { name: 'Type' }).selectOption('A');
+    await page.getByLabel('Value*').fill('test');
+    await page.getByLabel('Ports').fill('port1, port2');
+    await page.getByLabel('Course').fill('test course');
+    await page.getByLabel('Description').fill('test description');
+    await page.getByRole('button', { name: 'Create' }).click();
+    await expect(page).toHaveURL(/.*domains\/[0-9]*/);
```
humphd, line 32: Question: do we not need to wait on a navigation here?  Or is the fake routing we're doing in Remix such that we'll not need to worry about it?

humphd (changes requested):
This is great to see.  cc @Myrfion for another reviewer, since he wrote this.

Should we add tests for validation errors on required fields?  For example, if we put an IP address for a CNAME?  Could do that in follow-ups too.

test/e2e/new-dns-record.spec.ts
Eakam1007: There isn't any way to check the error messages since front end validation exists through Chakra UI components:
image
I tried to grab it but I am pretty sure it is not possible

test/e2e/new-dns-record.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  });
+
+  test('redirects to edit dns record page when all fields are filled', async ({ page }) => {
+    await page.goto('/domains/new');
+    await page.getByLabel('Record Name*').fill('test');
+    await page.getByRole('combobox', { name: 'Type' }).selectOption('A');
+    await page.getByLabel('Value*').fill('test');
+    await page.getByLabel('Ports').fill('port1, port2');
+    await page.getByLabel('Course').fill('test course');
+    await page.getByLabel('Description').fill('test description');
```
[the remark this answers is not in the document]
Eakam1007: Yeah, that's a good idea! The form is missing validations. I think we can add this point to #277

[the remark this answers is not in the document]
Ririio: I know ports are optional, but should we consider enforcing that it's numerical?

test/e2e/new-dns-record.spec.ts
humphd: In a follow-up, we could [style the control](https://chakra-ui.com/docs/components/input#changing-the-focus-and-error-border-colors) and check for that I guess.

test/e2e/new-dns-record.spec.ts
Eakam1007: We don't need to worry about it because the `toHaveURL` should wait until the page navigates to that URL, up to the configured timeout seconds

test/e2e/new-dns-record.spec.ts
Eakam1007: It sets that option for the entire describe block. Actually, using `test.use` in a `beforeEach` or `beforeAll` would result in an error ([relevant docs](https://playwright.dev/docs/api/class-test#test-use))

As for the function, we could create one but all we need is one line per test file I believe. Not sure if that is worth it. 
Would the function simply return this line?

test/e2e/new-dns-record.spec.ts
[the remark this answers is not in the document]
Eakam1007: We are storing ports as a string, and we ask the user to enter them in this format:
`port1, port2, port3`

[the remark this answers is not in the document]
Ririio: yeah that could work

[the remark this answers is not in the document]
Eakam1007: Do you want to add these to issue 277?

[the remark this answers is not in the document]
Ririio: Sure

Eakam1007:
> Should we add tests for validation errors on required fields? For example, if we put an IP address for a CNAME? Could do that in follow-ups too.

Hmm, I don't those validations exist yet. I think those can be added after #277 or as part of it

playwright.config.ts
```diff
@@ -67,12 +78,19 @@ const config: PlaywrightTestConfig = {
...
+      dependencies: ['setup'],
     },
     {
       name: 'Mobile Safari',
       use: {
-        ...devices['iPhone 12'],
+        userAgent: devices['iPhone 12'].userAgent,
+        viewport: devices['iPhone 12'].viewport,
+        deviceScaleFactor: devices['iPhone 12'].deviceScaleFactor,
+        isMobile: false,
```
humphd, line 89: Please add a comment with a link to the Issue where you got this tip discussing why we special-case mobile safari.

test/e2e/new-dns-record.spec.ts
humphd: I think the value of the function over this is that it can better communicate what is happening here, hide the specifics of the file path, etc.  It's not critical, but if it was me, I'd wrap a named function around this so that others can use it without understanding it.  Up to you.

humphd (approved):

test/e2e/new-dns-record.spec.ts
Eakam1007: Yea, that makes sense

humphd (approved):
Awesome

test/e2e/new-dns-record.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+});
+
+test.describe('authenticated as user', () => {
+  loggedInAsUser();
+
+  test('redirects to edit dns record page when required fields are filled', async ({ page }) => {
+    await page.goto('/domains/new');
+    await page.getByLabel('Record Name*').fill('test');
+    await page.getByRole('combobox', { name: 'Type' }).selectOption('A');
+    await page.getByLabel('Value*').fill('test');
```
SerpentBytes, line 19: Please let me know if I am wrong, but adding an asterisk to the value being passed is the playwright-specific syntax to indicate the field is not optional.

SerpentBytes (approved):

test/e2e/new-dns-record.spec.ts
Eakam1007: Oh, the label has an asterisk next to the required fields on the page. This test was autogenerated by using the VS extension for playwright. So, it grabs the input with that label (`Record Name*`) and fills in the value

SerpentBytes: Thanks for clarifying.

sfrunza13 (approved):

Issue #60, opened 2023-01-30
Do not require a `build` step before `dev`


dadolhay:
The current setup (probably because of the run-p causes a race condition) require a build step before dev successfully runs.

We should alter the script so it completes correctly even on the first time `dev` is executed

humphd:
I wonder if we should update the [setup script](https://github.com/Seneca-CDOT/starchart/blob/main/package.json#L22) so it does this properly for people.  It's already using `run-s` so it won't race.  We'd just have to put the steps in the right order in there.

After that, it should allow the other steps to run in parallel as they do now.

humphd:
Filed an issue upstream https://github.com/remix-run/blues-stack/issues/159

humphd:
This is no longer required with the delay I added.  Not perfect, but way better.

Issue #142, opened 2023-02-06
Provide Emergency Shutdown (Panic Button) functionality


ctyler:
In the event of a systems breach (e.g., through compromised secrets, or holes in auth libraries), a Bad Actor could potentially wreak a lot of chaos. It would be useful to have a mechanism to globally shut down the system in an emergency in such a way that it can't be overridden by the (compromised) API, and which would more positively disable active resource records (as opposed to no longer serving them authoritatively and waiting for them to eventually timeout across layers of caches).

humphd:
Thanks for filing this.  What sort of mechanism do you imagine this being?  What does the workflow look like?

humphd:
I'm not clear what action to take here and don't have further feedback.  Closing for now, but feel free to re-open or file something new if you want to explore.

PR #321, opened 2023-03-10
Rework startup and async init logic

### LINKED ISSUE #319: Fix startup init logic to work in dev and production
I was testing the app on staging, and the server `init` logic I wrote for development doesn't work in production.  I need to figure out a better solution for handling `async` startup logic.  I've been playing with some solutions, but none of them are perfect.  I'll have to get this sorted out for 0.6

humphd:
Fixes #319, part of #60.  This can happen in 0.6.

This changes a bunch of things about how startup happens, in particular how async startup logic is handled.  The way that Remix does its initialization is different than I'm used to.  It works in 2 parts, with `server.ts` (the "server") and then everything in `app/*` (the "remix" bundle).  These produce `build/server.js` and `build/index.js` respectively.

The way that live reload happens on the server side, everything in `require.cache` gets deleted when the server restarts.  This means that module-level, in-memory variables (e.g., instances) get deleted in development (but not production).  To combat this, they recommend you cache instances on the `global` in development, and reattach them when loading again.

This is fine for instances created using synchronous code paths, but breaks down for asynchronous code.  We have a bunch of the latter, and needed a solution for setting it up once.

I've adjusted a few things to accomplish this:

1. Altered the way that `saml.server.ts` loads the IDP metadata.  We now do it from a file synchronously vs. fetching from a URL.  This turns out to be the recommended way anyway, from a security point of view.  I've had to update the `env` and Docker setup to deal with this new build arg/env requirement.  I've added the XML for dev and staging to `config/`.
2. Altered the way that `dns.server.ts` creates the hosted zone in development.  I'm not creating it dynamically on first use, inside existing `async` functions.  Only the first call will incur the extra cost of creation, and others will use the same zone id.
3. Altered the way that `logger.server.ts` was setting up the termination logging.  I only want to add these listeners once, so this manages that via a `global` in dev.

I fixed some other things along the way:

1. Bug in how the AWS URL endpoint was being set for development.  It wouldn't have worked in production
2. I've increased the default max listeners for `EventEmitter` so we don't keep get warnings.
3. I've removed the extra `await` in `saml.server.ts` for `parseLoginResponse` on `idp`
4. Improved a race condition running `npm run dev` by alter the order we call the various `dev:*` scripts vs. letting it decide on its own.

Later I'll do a bit more on this, but I don't want to change the worker queue stuff until we land @SerpentBytes's stuff in #268.


sfrunza13 (commented):
Would it be any better or safer to have the SAML metadata as a secret instead of a file under config?

app/lib/redis.server.ts
```diff
@@ -5,8 +5,9 @@ import type { Redis as RedisType, RedisOptions } from 'ioredis';
 
 // Using Redis means we'll need a lot more event listeners for BullMQ
 // or we will hit the default max limit of 10 listeners. The preferred
-// fix for this warning is to increase that default.
-EventEmitter.defaultMaxListeners = 64;
+// fix for this warning is to increase that default. Setting this high
+// per https://github.com/OptimalBits/bull/issues/615#issuecomment-897051838
+EventEmitter.defaultMaxListeners = 10_000;
```
SerpentBytes, line 10: I am guessing as we added more Worker/Queue logic, we started getting this error more frequently. With the addition of the first worker/queue, I recall that the warning related to hitting the limits of  MaxListeners was occasional. Setting the value to max should eliminate the warning for good.

Genne23v (approved):
Looks good to me!

SerpentBytes (dismissed):

humphd:
> Would it be any better or safer to have the SAML metadata as a secret instead of a file under config?

It's not a secret, so I don't think it belongs there, no.

humphd:
I've rebased this in @SerpentBytes' work that just landed, and gotten rid of the last of the `init` logic in the server.  I've also moved the shutdown logic for each worker to the same file where it happens, so there isn't a module dependency between the server and remix bundle.

Due to a race condition in the startup logic with `dev`, I've stopped using the `--watch` on the server build, and instead pre-build it before we start `dev`.

sfrunza13 (approved):

app/queues/certificate/challenge-completer-worker.server.ts
```diff
@@ -17,3 +17,5 @@ export const challengeCompleterWorker = new Worker<ChallengeCompleterData>(
   },
   { connection: redis }
 );
+
+process.on('SIGINT', () => challengeCompleterWorker.close());
```
SerpentBytes, line 21: How is closing connections to workers based on signal interruption better than how we were doing things previously? By doing this, are we building towards the panic feature requested in #142.

humphd: It's the same idea.  The graceful shutdown would do the same thing, but it would require us to create a connection between the server bundle and remix bundle (i.e., we'd have to get a reference to the workers in another bundle).  This way, we can do it all in the same bundle.

The SIGINT is a POSIX signal, see https://dsa.cs.tsinghua.edu.cn/oj/static/unix_signal.html#:~:text=The%20SIGINT%20signal%20is%20sent,break%22%20key%20can%20be%20used.&text=The%20SIGKILL%20signal%20is%20sent,to%20terminate%20immediately%20(kill).

SerpentBytes: Thanks for clarifying.

SerpentBytes (approved):

humphd:
@Genne23v can you sign-off on this before I merge?

Genne23v (approved):

Genne23v:
> @Genne23v can you sign-off on this before I merge?

I just signed off

Issue #245, opened 2023-02-19
Standardize on naming for Domain vs. Record


humphd:
All over the code we use Record and Domain in various contexts.  I'm reviewing code right now that does this:

```ts
function onDomainAction(domain: Record, action: DomainsTableAction) {
```

This seems odd to me.  Can we come-up with a standard way to deal with this through the code and normalize our types/names?

Eakam1007:
We probably should use `Record` instead of `Domain`

humphd:
I think `Record` is too vague.  A `Record` makes sense only if you already understand the context, but it can be confused with a record in a database, etc.  I propose we call these `DNS Records` (in UI) and `DnsRecord` in code and the database.

Prior art:

- https://www.cloudflare.com/learning/dns/dns-records/
- https://ns1.com/resources/dns-records-explained

humphd:
We should get this done in the next release if possible, so that it doesn't drag out.

PR #324, opened 2023-03-12
Integrate domain create,delete,prolong with server + minor UI hintsgs

### LINKED ISSUE #280: Add appropriate connect to records service to make deletion work
We don't have appropriate server-side functionality that would actually delete the record from the route56, but when it is gonna be ready it should be integrated in the records page

Myrfion:
## Changes
- integration for domain creation with server
- integration for domain deletion and prolong
- add refresh domains button
- remove records from seeding, because they are invalid because they are not connected to any hosted zone and in development, it is created at the start of dev environment, so when you seed db it doesn't exist yet. Since the creation of records is available now with UI people can do that for testing purposes. Invalid records provide more confusion than benefit (because you can't perform server-side operations with them properly) so removing them
- Highlight names based on the domains table 
- Add `react-icons` package for icons, this is the one that chakra show as the example of 3-rd party package so I think it should be good, the chakra's icons library itself is very poor 

The screen recording of how it looks now is attached

https://user-images.githubusercontent.com/32075241/224517240-32b1c48e-dd19-49dd-b994-06412e7b6184.mov

closes #280 #272


app/queues/dns/dns-flow.server.ts
```diff
@@ -157,13 +156,13 @@ export const updateDnsRequest = async ({ id, username, type, name, value }: JobR
 };
 
 export const deleteDnsRequest = async ({ id, username, type, name, value }: JobRecord) => {
-  const fullRecordName = buildDomain(username, name);
+  logger.debug(process.env.AWS_ROUTE53_HOSTED_ZONE_ID);
```
Ririio, line 159: If you don't mind me asking, what does this logger do in this line?

app/queues/dns/dns-flow.server.ts
```diff
@@ -82,8 +83,6 @@ export const addDnsRequest = async ({ username, type, name, value }: JobRecord)
 };
 
 export const updateDnsRequest = async ({ id, username, type, name, value }: JobRecord) => {
-  const fullRecordName = buildDomain(username, name);
```
Ririio, line 85: Question: Why did we have this line in the first place?

app/routes/__index/domains/index.tsx
```diff
@@ -1,13 +1,18 @@
 import { AddIcon } from '@chakra-ui/icons';
-import { Button, Container, Flex, Heading, Text } from '@chakra-ui/react';
+import { Button, Container, Flex, Heading, Icon, IconButton, Text } from '@chakra-ui/react';
```
Ririio, line 2:
```suggestion
import { Button, Container, Flex, Heading, Icon, Text } from '@chakra-ui/react';
```

remove the button if you aren't going to use it

app/routes/__index/domains/index.tsx
```diff
@@ -1,13 +1,18 @@
...
 import { Link, useNavigate, useSubmit, useTransition } from '@remix-run/react';
 import { typedjson, useTypedLoaderData } from 'remix-typedjson';
+import { FaRedoAlt } from 'react-icons/fa';
 import type { DomainsTableAction } from '~/components/domains-table';
 import DomainsTable from '~/components/domains-table';
-import { getRecordsByUsername } from '~/models/record.server';
+import { getRecordById, getRecordsByUsername } from '~/models/record.server';
 import { requireUsername } from '~/session.server';
 import type { Record } from '@prisma/client';
+import { addDnsRequest, deleteDnsRequest, updateDnsRequest } from '~/queues/dns/dns-flow.server';
```
Ririio, line 12:
```suggestion
import { deleteDnsRequest, updateDnsRequest } from '~/queues/dns/dns-flow.server';
```

You forgot to remove `addDnsRequest` from the imports

app/routes/__index/domains/index.tsx
```diff
@@ -1,13 +1,18 @@
...
 import type { LoaderArgs, ActionArgs } from '@remix-run/node';
 import { Link, useNavigate, useSubmit, useTransition } from '@remix-run/react';
 import { typedjson, useTypedLoaderData } from 'remix-typedjson';
+import { FaRedoAlt } from 'react-icons/fa';
 import type { DomainsTableAction } from '~/components/domains-table';
 import DomainsTable from '~/components/domains-table';
-import { getRecordsByUsername } from '~/models/record.server';
+import { getRecordById, getRecordsByUsername } from '~/models/record.server';
 import { requireUsername } from '~/session.server';
 import type { Record } from '@prisma/client';
```
Ririio, line 11: I would recommend grouping the imports depending on whether they are dependencies or file imports like `@prisma/client` should not be grouped with the other prisma deps

app/routes/__index/domains/index.tsx
```diff
@@ -16,13 +21,51 @@ export const loader = async ({ request }: LoaderArgs) => {
...
+  invariant(intent, 'Missing `intent` in posted form data');
+  invariant(id, 'Missing `id` in posted form data');
+
+  const recordID = Number(id);
+  const record = await getRecordById(recordID);
+
+  if (!record) {
+    throw new Response('The record is not found', {
+      status: 404,
+    });
```
Ririio, line 37: Question: In regards to this condition here, does this throw a status 404 if a specific record is not found? If that's the case you might want to throw a status [422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) instead to be more accurate

app/queues/dns/dns-flow.server.ts
Myrfion: It was probably copied from the record creation, but since the name is already formatted there and saved in that format there is no need to format (build) it again

app/queues/dns/dns-flow.server.ts
Myrfion: I was actully debugging something and forgot to remove it :) good catch

app/queues/dns/dns-flow.server.ts
```diff
@@ -11,6 +11,7 @@ import { WorkType } from './dns-record-worker.server';
 import type { FlowJob } from 'bullmq';
 import type { JobRecord } from './dns-record-worker.server';
 import { buildDomain } from '~/utils';
+import logger from '~/lib/logger.server';
```
Ririio, line 14: You might have forgotten to remove the import for logger
```suggestion
import logger from '~/lib/logger.server';
```

app/routes/__index/domains/index.tsx
Myrfion: Good catch, 422 would be in fact better

app/components/domains-table.tsx
```diff
@@ -89,6 +90,25 @@ export default function DomainsTable(props: DomainsTableProps) {
     }
   }
 
+  function renderDomainName(domainName: string) {
```
SerpentBytes, line 93: Would you mind renaming the function to `renderDnsRecordName`, since it's being added to this PR? I have to standardize the naming convention as specified #245 for the existing code.

SerpentBytes (commented):
Just curious as to why we are deleting all this code. Shouldn't we keep it for local testing?

prisma/seed.ts
```diff
@@ -50,80 +50,6 @@ async function seed() {
     ],
   });
 
-  await prisma.record.createMany({
-    data: [
```
SerpentBytes, line 54: Just curious as to why we are deleting all this code. Shouldn't we keep it for local testing?

app/components/domains-table.tsx
```diff
@@ -89,6 +90,25 @@ export default function DomainsTable(props: DomainsTableProps) {
     }
   }
 
+  function renderDnsRecordName(domainName: string) {
+    const words: string[] = domainName.split('.');
```
humphd, line 94: Don't bother including types when they can be inferred.  Here, the value will automatically be `:string[]` due to the assignment.

app/components/domains-table.tsx
```diff
@@ -89,6 +90,25 @@ export default function DomainsTable(props: DomainsTableProps) {
...
+    const nameBase: string = words[0];
+    const restOfName: string = words.slice(1).join('.');
```
humphd, line 96:
```ts
const [nameBase, ...restOfName] = words;
```

app/components/domains-table.tsx
```diff
@@ -89,6 +90,25 @@ export default function DomainsTable(props: DomainsTableProps) {
...
+
+    return (
+      <Flex alignItems="flex-end" flexDirection="row">
+        <Text>
+          <Text as="span" sx={{ fontSize: 'md', fontWeight: 'medium' }}>
```
humphd, line 101: I wonder if we should only do the bold/gray change, but leave the size the same?  I think it looks a bit off to have the name be larger.

app/queues/dns/dns-flow.server.ts
humphd: cc @Genne23v.  I wonder if we should document this and/or alter the name of the variable, since a name portion and a full domain are different.  It can happen in a follow-up vs here, but we should clarify so that `name` doesn't mean 2 different things.

app/routes/__index/domains/index.tsx
```diff
@@ -16,13 +22,51 @@ export const loader = async ({ request }: LoaderArgs) => {
...
+  const id = formData.get('id');
+  invariant(intent, 'Missing `intent` in posted form data');
+  invariant(id, 'Missing `id` in posted form data');
+
+  const recordID = Number(id);
+  const record = await getRecordById(recordID);
+
+  if (!record) {
+    throw new Response('The record is not found', {
+      status: 422,
```
humphd, line 37: Why is this not `404`?  The client has requested a resource with a given `id`, and the database doesn't have it.  Isn't that the definition of `404`?

app/routes/__index/domains/index.tsx
```diff
@@ -16,13 +22,51 @@ export const loader = async ({ request }: LoaderArgs) => {
 };
 
 export const action = async ({ request }: ActionArgs) => {
-  if (request.method === 'PUT') {
-    // TODO: Call the appropriate code to prolong record
-  } else if (request.method === 'DELETE') {
-    // TODO: Call the appropriate code to delete record
+  const username = await requireUsername(request);
+  const formData = await request.formData();
+  const intent = formData.get('intent');
```
humphd, line 27: Use zod to do the form parsing, like we do in https://github.com/DevelopingSpace/starchart/blob/3fa7f7079b2df10aad33ef3d96450c3fe9dc9c37/app/routes/__index/domains/new.tsx

app/routes/__index/domains/index.tsx
```diff
@@ -16,13 +22,51 @@ export const loader = async ({ request }: LoaderArgs) => {
...
+  if (!record) {
+    throw new Response('The record is not found', {
+      status: 422,
+    });
   }
 
-  return null;
+  switch (intent) {
+    case 'renew-record':
+      await updateDnsRequest({
```
humphd, line 43: Question: is a "renew" trying to update the expiry date?  It doesn't look like we actually do that in the db: https://github.com/DevelopingSpace/starchart/blob/3fa7f7079b2df10aad33ef3d96450c3fe9dc9c37/app/queues/dns/dns-flow.server.ts#L87-L107. cc @Genne23v

app/routes/__index/domains/index.tsx
```diff
@@ -37,10 +81,10 @@ export default function DomainsIndexRoute() {
         navigate(domain.id.toString());
         break;
       case 'DELETE':
-        submit({ id: domain.id.toString() }, { method: 'delete' });
+        submit({ id: domain.id.toString(), intent: 'delete-record' }, { method: 'delete' });
```
humphd, line 84: Question: can you not put the `id` in a `<Input type="hidden" />` in the form?  That's the more common way to store it.

app/routes/__index/domains/index.tsx
```diff
@@ -57,6 +101,11 @@ export default function DomainsIndexRoute() {
           a galley of type and scrambled it to make a type specimen book.
         </Text>
         <Flex justifyContent="flex-end">
+          <form method="get">
```
humphd, line 104: `form` should be `Form` to get the benefits of Remix.

prisma/seed.ts
humphd: From the description, he's removing it because the seeded records are not valid, since they are only in the db and not Route53.  From this point on, we have to create the records properly in both at runtime.

humphd (changes requested):
I'm excited to see this happening!

app/components/domains-table.tsx
Myrfion: I don't mind making it the same font size

app/routes/__index/domains/index.tsx
Myrfion: Mario had this comment on this, which seemed to make sense for me:

@[Ririio](https://github.com/Ririio) Ririio [yesterday](https://github.com/DevelopingSpace/starchart/pull/324#discussion_r1133184112)
Question: In regards to this condition here, does this throw a status 404 if a specific record is not found? If that's the case you might want to throw a status [422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) instead to be more accurate

app/routes/__index/domains/index.tsx
Myrfion: `updateRecordStatusById` also updates the expiry date to the date now + 6 months

Genne23v: In that case, you can add this function to `record.server.ts`. 
```js
export async function renewRecordById(id: Record['id']) {
  const expiresAt = dayjs().set('month', 6).toDate();
  return prisma.record.update({
    where: {
      id,
    },
    data: {
      expiresAt,
    },
  });
}
```

app/components/domains-table.tsx
```diff
@@ -89,6 +90,24 @@ export default function DomainsTable(props: DomainsTableProps) {
     }
   }
 
+  function renderDnsRecordName(domainName: string) {
+    const words = domainName.split('.');
```
humphd, line 94: nit:

```ts
const [nameBase, ...restOfName] = domainName.split('.');
```

app/components/domains-table.tsx
```diff
@@ -141,7 +161,7 @@ export default function DomainsTable(props: DomainsTableProps) {
                         <Td>{renderDomainStatus(domain.status)}</Td>
                         <Td>
                           <Flex justifyContent="space-between" alignItems="center">
-                            {domain.name}
+                            {renderDnsRecordName(domain.name)}
```
humphd, line 164: I'm confused why this isn't a component?

app/models/record.server.ts
```diff
@@ -82,3 +82,15 @@ export async function doesRecordExist(data: Pick<Record, 'username' | 'type' | '
 export async function deleteRecordById(id: Record['id']) {
   return prisma.record.delete({ where: { id } });
 }
+
+export async function renewRecordById(id: Record['id']) {
```
humphd, line 86: You don't need `async` here, since you aren't `await`'ing

humphd (changes requested):

app/components/domains-table.tsx
Myrfion: I don't have perfect answer for this, I just used to use this pattern of table components with render functions which I have seen in many other places.

humphd: I don't think this is a good pattern.  React provides functional components for exactly this use case.  Why step outside of its component system?

Where I'd use a function call is to some non-React code that isn't returning a component.  But that's exactly what you're doing here, so it doesn't make sense to avoid React.

humphd (changes requested):

app/components/dns-record/dns-record-name.tsx
```diff
@@ -0,0 +1,27 @@
+import type { FC } from 'react';
+import React from 'react';
+import { Flex, Text } from '@chakra-ui/react';
+
+interface DnsRecordNameProps {
+  name: string;
+}
+
+const DnsRecordName: FC<DnsRecordNameProps> = (props) => {
+  const { name } = props;
```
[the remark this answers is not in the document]
Myrfion: I don't see how is it cleaner. In my opinion my version is more cleaner, since it show that the props object is passed and then decompose or for certain params

app/components/record-delete-alert-dialog.tsx
```diff
@@ -31,9 +34,13 @@ export default function RecordDeleteAlertDialog(props: RecordDeleteAlertDialogPr
             <Button colorScheme="gray" onClick={onCancel} ref={cancelRef}>
               Cancel
             </Button>
-            <Button colorScheme="brand" onClick={onConfirm} ml="3">
-              Delete
-            </Button>
+            <Form method="delete" style={{ margin: 0 }}>
+              <input type="hidden" name="id" value={dnsRecord?.id} />
+              <input type="hidden" name="intent" value="delete-record" />
```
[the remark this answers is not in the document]
Myrfion: I pretty sure Chakra UI Input provides some nice styling, unlike the input which has basic styles. Since inputs above have type `hidden` there is no need for them to be fancy cuz they are not visible

app/components/dns-record/dns-record-name.tsx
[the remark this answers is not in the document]
Ririio: Would that be the same if you specify it on the params? When you write it from there it already specifies that it's a prop? I was just wondering why add an extra line for a process you can do directly from the parameter?

app/models/record.server.ts
```diff
@@ -82,3 +82,15 @@ export async function doesRecordExist(data: Pick<Record, 'username' | 'type' | '
 export async function deleteRecordById(id: Record['id']) {
   return prisma.record.delete({ where: { id } });
 }
+
+export function renewRecordById(id: Record['id']) {
```
SerpentBytes, line 86: Similar to one of my previous comments: could you change it to `renewDnsRecordById`?

app/components/dns-record/dns-record-name.tsx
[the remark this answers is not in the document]
humphd: `FC` has fallen out of favour https://github.com/facebook/create-react-app/pull/8177

humphd (changes requested):
One thing from me, and I think this is good.  Fix the other issues people are calling out, too.

Ririio (approved):
I don't see any other issues

humphd (approved):

Issue #291, opened 2023-03-05
Prepare for, document, and automate our database migration and deployment strategy


humphd:
In order to ship the app to staging(#43) we need to get ready to handle database migrations.

Currently, we are working with a development database locally, which can be blown away whenever we need to fix something.  That won't work as soon as we move this to staging and production (hopefully in 0.5).  In those cases, we'll have to create migrations and run them.

I am not an expert on this, but here is my current understanding of the problem and solution.  Hopefully others will help with the research and implementation.  I am basing this largely on the [Prisma Migrate Mental Model](https://www.prisma.io/docs/concepts/components/prisma-migrate/mental-model):

1. In development, we use [`prisma db push`](https://www.prisma.io/docs/concepts/components/prisma-migrate/mental-model#prototype-your-schema) to force the local development database to use our schema.  This has been fine for development up to now, but won't work going forward
2. In staging and production (henceforth referred to as "prod"), we need to use [`prisma migrate deploy`](https://www.prisma.io/docs/concepts/components/prisma-migrate/mental-model#prisma-migrate-in-a-staging-and-production-environment) to run migrations against the database and synchronize it with our desired schema.
3. Doing 2. requires us to first create a set of [migrations in `prisma/migrations`](https://www.prisma.io/docs/concepts/components/prisma-migrate/mental-model#how-prisma-migrate-tracks-the-migration-state), which are SQL files that describe transformations from one database state to another (e.g., adding a table, creating an index, etc).  We do that using [`prisma migrate dev --name init`](https://www.prisma.io/docs/concepts/components/prisma-migrate/get-started#get-started-with-prisma-migrate).
4. Now we can use `prisma migrate deploy` in prod to update our database (i.e., push schema and/or apply migrations).
5. Whenever we need to alter our schema after we establish our baseline migration, we need to create a new migration SQL file using [`prisma migrate dev --name <name_of_migration>`](https://www.prisma.io/docs/concepts/components/prisma-migrate/get-started#get-started-with-prisma-migrate).  This SQL migration would then be applied in prod using `prisma migrate deploy`
6. We need to include `prisma migrate deploy` in our webhook deploy step
7. Running `prisma migrate deploy` automatically in prod will require updates to our Dockerfile, in order to create an image layer that has the necessary dependencies and migrations
8. This entire process needs to be carefully document in our `CONTRIBUTING` and `DEPLOY` docs

What else am I missing?

[...]

PR #325, opened 2023-03-12
Add prisma Docker target + CI to setup db in staging/prod

humphd:
Part of #291

In order to ship to staging/production, we need to be able to set up the database.  Doing so requires running Prisma with our schema from the staging/production VMs.  This creates a Docker build stage that can be optionally run in a stand-alone way to accomplish this.

Every time we push to `main`, CI will create both `startchart` and `starchart-prisma` images.  The latter is used to push the current schema to the database (there might be a better set of commands for us to run, let me know what you think).

We won't do this automatically, but I need this to exist so I can run it manually via Docker.

This change also does the following:

1. Updates the SHA of the node 18 image we use (they shipped a security update recently)
2. Adds a `db:setup` npm script that does everything except seeds the db
3. Removes QEMU from the Docker jobs, since we aren't cross-compiling anymore

This doesn't need to get looked at until next week as part of 0.6.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -314,7 +314,7 @@ jobs:
             # cache node modules using the same key as restore.
           key: ${{ steps.node_modules-cache-restore.outputs.cache-primary-key }}
 
-  Docker-Build:
+  Starchart-Docker:
     # We'll only build and push when landing on main, not for PRs
     if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
     runs-on: ubuntu-latest
@@ -326,9 +326,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
@@ -350,7 +347,7 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build and push Docker image
+      - name: Build and push Starchart Docker image
         uses: docker/build-push-action@v4
         with:
           context: .
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 legacy-peer-deps=true
+update-notifier=false
--- a/DEPLOY.md
+++ b/DEPLOY.md
@@ -43,3 +43,15 @@ The following secrets must be added to the Docker engine using [Docker Swarm sec
 | `NOTIFICATIONS_USERNAME`               | The SMTP username to use for sending notifications                                                                                                                           |
 | `NOTIFICATIONS_PASSWORD`               | The SMTP password to use for sending notifications                                                                                                                           |
 | `DATABASE_URL`                         | The MySQL database connection string URL. NOTE: this is needed as an environment variable only when doing database setup commands, but read as a secret when running the app |
+
+## Running the App via Docker
+
+The app can be started via Docker. If it is the first time you are starting the app, or if there are database changes that need to be applied first, you can set `DATABASE_SETUP=1` in the environment to run Prisma and sync the schema and database:
+
+```sh
+# Run the app without doing anything to the database
+$ docker run ghcr.io/developingspace/starchart
+
+# OR, run the app, but sync the datbase first
+$ docker run -e DATABASE_SETUP=1 ghcr.io/developingspace/starchart
+```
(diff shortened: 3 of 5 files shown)


dadolhay:
@humphd Interesting approach, but do we have to build a separate docker image just for this?

What if we used something like this lib:
https://www.npmjs.com/package/if-env
and decide to run the db:setup based on an external env?

humphd:
> @humphd Interesting approach, but do we have to build a separate docker image just for this?
> 
> What if we used something like this lib: https://www.npmjs.com/package/if-env and decide to run the db:setup based on an external env?

It's not a good idea to run `npm` in production (i.e., I don't want to add the extra process invocation when we run the app normally).  For the case of the manual setup step, it's not a huge deal, because it is meant to exit on its own (i.e., it's not something we need to manage with signals).

I have to create a different image because I'm not using the usual stages to end at the production image.

Let me know if I'm misunderstanding your comment.

humphd:
@dadolhay thinking about your comment more.  What we could do is create a `start.sh` script, which does what you suggest, and then if you include `DATABASE_SETUP=1 DATABASE_URL=mysql://...` in the env, we could run the prisma commands for you before starting the app.  That's cleaner than what I'm doing.  I'll try an update to this.

dadolhay:
@humphd Thanks, that sounds like a much simpler solution

humphd:
I've updated this PR and made some changes:

1. I've removed the extra docker build step from CI. We'll continue to only have a single image
2. I've added `update-notifier=false` to our `.npmrc` so that npm doesn't warn about newer versions when running in Docker.  I've also copied `.npmrc` into the final Docker stage, so it uses it.
3. I've copied the `prisma/` directory to the final Docker stage, so we have access to the schema and migrations
4. I've added a `docker-entrypoint.sh` bash script, which is called in our `ENTRYPOINT`.  This script checks to see if `DATABASE_SETUP=1` is set, and if it is, first runs `npx prisma db push` before starting the app.

Some questions I still have:

1. is `npx prisma db push` the correct thing to do?  Is there anything more I should do here, or different?
2. does it make sense to start the container after we run the db setup?  I'm not sure if it does.  I feel like doing the database setup is a discrete step, separate from running the app, but I'm not sure.

To test this:

1. Update `dev-secrets/DATABASE_URL` to be `mysql://starchart:starchart_password@host.docker.internal:3306/starchart` since you need to connect to the host from within a container (i.e., don't use localhost or 127.0.0.1)
2. Build the image: `docker build -t starchart .`
3. Run a container based on the image, simulating the secrets and setting the env var: `docker run --rm -e DATABASE_SETUP=1 -v $PWD/dev-secrets:/run/secrets starchart`.  You can also run it without the env: `docker run --rm starchart`

NOTE: it currently crashes on startup because the env var for `DATABASE_URL` needs to be secrets instead.  I've already fixed this, but need to rebase.

dadolhay (dismissed):

dadolhay:
> * is `npx prisma db push` the correct thing to do?  Is there anything more I should do here, or different?

I think intil we reach the DB baseline, this should be OK. If we encounter errors, we can change it to a full reset, but that could be destruptive to manual testers

> * does it make sense to start the container after we run the db setup?  I'm not sure if it does.  I feel like doing the database setup is a discrete step, separate from running the app, but I'm not sure.

I think it does, spares an extra step, but this is not a strong opinion

Approved, I like this approach much more, it's a lot simpler

dadolhay (dismissed):

package.json
```diff
@@ -19,7 +19,8 @@
     "db:reset": "prisma migrate reset",
     "db:format": "prisma format",
     "db:studio": "prisma studio",
-    "setup": "run-s db:generate db:push db:seed",
+    "db:setup": "run-s db:generate db:push",
+    "setup": "run-s db:setup db:seed",
```
cychu42, line 23: Nit. I'm wondering if there are better names to distinguish the two setups. They are both database related, and the only difference is whether we seed, not whether it's database related (`db`).
Maybe `db:setup` can be `setup:seedless`?

cychu42:
> I think intil we reach the DB baseline, this should be OK. If we encounter errors, we can change it to a full reset, but that could be destruptive to manual testers

Agree. If we don't need to worry about migration files or checking for database drift (via shadow database), `npx prisma db push` should be fine.

> does it make sense to start the container after we run the db setup? I'm not sure if it does. I feel like doing the database setup is a discrete step, separate from running the app, but I'm not sure.

I suppose you can look at container as part of the app and therefore make sense to start it as part of starting the app. Let me know if I'm misunderstanding you.

package.json
humphd: I'm not actually using this in my code, so I think I'll revert it back to what it was.

dadolhay (approved):

cychu42 (approved):
LGTM!

PR #334, opened 2023-03-14
Add docker-staging.yml for deploymen to staging

### LINKED ISSUE #43: Deploy the app to staging

humphd:
Fixes #43 

This defines a Docker Swarm Stack for staging.  Some notes on what's included:

- Redis configured to only run on the manager node (not worker(s)), and to use a named volume
- Starchart with secrets and staging env vars configured, 2 instances will be run in parallel

I'm going to manually test this on staging and report back with fixes and info on how it goes.


humphd:
Rebased on the other fixes we did this morning.  If this passes tests, I think it's ready to land.

humphd:
OK, this is all green, ready for review!

DEPLOY.md
```diff
@@ -19,6 +19,7 @@ The following configuration values must be set via environment variables.
 
 | Variable Name                | Description                                                                                                                                                                        |
 | ---------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `HOSTNAME`                   | The hostname of the server (e.g., `https://mycustomdomain.senecacollege.ca`)                                                                                                       |
```
sfrunza13, line 22: host.docker.internal now, not sure if we should add to description

app/lib/saml.server.ts
```diff
@@ -18,21 +18,32 @@ const idp = samlify.IdentityProvider({
   metadata: xml,
 });
 
+const { NODE_ENV, HOSTNAME } = process.env;
+
+/**
+ * In production, we require the hostname
+ */
+if (NODE_ENV === 'production') {
```
sfrunza13, line 26: why not always check for hostname

sfrunza13 (approved):

DEPLOY.md
humphd: Good idea, updated docs to be more clear.

app/lib/saml.server.ts
humphd: Good idea, updated in new push.

sfrunza13 (approved):

SerpentBytes (approved):

Genne23v (approved):
It looks good to me!

docker-staging.yml
```diff
@@ -0,0 +1,67 @@
...
+    restart: unless-stopped
+    depends_on:
+      - redis
+    ports:
+      - 8080:8080
+    environment:
+      - HOSTNAME=https://mycustomdomain-dev.senecacollege.ca
+      - AWS_ROUTE53_HOSTED_ZONE_ID=Z0228625ICAL609E0BBT
+      - LETS_ENCRYPT_ACCOUNT_EMAIL=mycustomdomain-dev@senecacollege.ca
+      - LETS_ENCRYPT_DIRECTORY_URL=https://acme-staging-v02.api.letsencrypt.org/directory
```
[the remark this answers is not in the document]
humphd: I'm going to land, but if this is needing to get changed, let me know if a follow up.

PR #337, opened 2023-03-14
Took array and made it string

### LINKED ISSUE #336: Flatten groups array
Group comes back as array for admins, use join to make it a string to match prisma schema.

sfrunza13:
joined group claims into string, added array of groups to hans since he's not in seed
Closes #336


humphd (approved):
Did you test logging in as one of our built-in admins locally?  Does that work?

The code looks good.

sfrunza13:
> 

This is why I added array of groups to hans, I tested and logged the string that the join creates before pushing

Ririio (approved):

PR #339, opened 2023-03-14
Use username and baseDomain in /certificate

humphd:
When we show the base domain for a user, we're currently hard-coding it.  This uses the logged in user's info instead.


Ririio (approved):
oh I didn't even consider that when writing mb

Eakam1007 (approved):

PR #340, opened 2023-03-14
Retry playwright tests at least once

Eakam1007:
Videos and traces are configured to be generated on the first retry. So, changing the number of retries outside of the `CI` environment to 1 would allow for easier debugging when running tests locally


sfrunza13 (approved):
Cool, all for this. Where does it put the files that are generated?

Eakam1007:
It puts them under the `test-results` folder. This is setup here: https://github.com/DevelopingSpace/starchart/blob/445361213802eafaaf3c813f0b28ed918b81308f/playwright.config.ts#L119

humphd (approved):

PR #343, opened 2023-03-14
Logout is only our session logout

sfrunza13:
To be tested on staging, but I think this will only effect the starchart session.


humphd (changes requested):
Looks good.  One thing to touch up.  Thanks for fixing this.

humphd (dismissed):

app/routes/logout.tsx
```diff
@@ -1,17 +1,16 @@
...
 
 export const action = async ({ request }: ActionArgs) => {
   const user = await getUsername(request);
 
-  // Do the opposite of login
+  // Invalidate the Starchart session but do not log out from Seneca IDP.
   if (user) {
-    const context = createLogoutRequest(user);
-    return redirect(context);
+    return await logout(request);
```
Genne23v, line 11: I learned this only weeks ago in code review. `await` is redundant when returning a promise.

Genne23v (changes requested):

app/routes/logout.tsx
Ririio: seems like the only viable time to use [await](https://eslint.org/docs/latest/rules/no-return-await) is when you're trying to catch and error from another function

sfrunza13: > seems like the only viable time to use [await](https://eslint.org/docs/latest/rules/no-return-await) is when you're trying to catch and error from another function

I don't think this is true

sfrunza13: However I do think Won might be right about it being redundant here.

humphd (approved):

Genne23v (approved):

PR #349, opened 2023-03-15
Refactor dns flows and workers

### LINKED ISSUE #28: Filter bad words in domain names

- Filter inappropriate domain name

### LINKED ISSUE #110: Enforce a configurable limit on the number of DNS records per user

### LINKED ISSUE #315: Break DNS update into two phases
Currently all of the steps to add/update record immediately go into queue which gives frontend long time to see the response. 

Instead the system updates DB immediately without waiting in a queue whole time. Then, queue will be run to update Route 53, poll the status, and update the Route 53 status in DB. 

This will give much better experience to the user.

### LINKED ISSUE #333: DNS flow refactors
I'd like to start a discussion about this file's TS declarations, because I'm quite certain it's wrong:
https://github.com/DevelopingSpace/starchart/blob/main/app/queues/dns/dns-flow.server.ts
First, we are trying to use `JobRecord`  for absolutely everything. Basically all the workers and exported flow initiator functions all use this, and we are trying to twist it to suite the specific use-case. This causes some very hard to follow code, and inappropriate use.
I.e.,  Line 17, `name`  refers to a subdomain only (which was a breaking change, but let's not get bogged down)
In Line 18, using name (subdomain) we create `fullRecordName` which is an FQDN
In line 24, we pass this FQDN down as `name`,
so in line 17 `name` was a subdomain, in line 24 `name` is FQDN and they are both using the same TS interface. This is wrong.

In my humble opinion:
- Each worker should be it's own file and should export an interface, of what it exactly expects as input params (`data`)
- [dns-flow.server.ts](https://github.com/DevelopingSpace/starchart/blob/main/app/queues/dns/dns-flow.server.ts) should import those interfaces and use the correct interface for each `data`  attribute in the flow generators
- There also should be a separate TS interface/type for each flow initiator. It can reuse the worker's interface, but should be specific.
- We should start using `subdomain`  and `fqdn`  atribute and variable names to signify what we are storing in it, `name` is not specific enough (see above issue)
- We need a way to pass down either FQDN or subdomain to the appropriate flow initiators, flow initiators should generate the fqdn if subdomain was provided. This is needed for the ACME integration (and probably other later code as well)
- Input validation should be part of the flow initiator, not the worker. This way we can synchronously return an error to the caller, and it's not the subsequent async job that's going to fail

Genne23v:
Close #315 
Close #333 

- Split DNS module into two phases (DB and DNS)
- Specify type to each flow, worker and job
- Change name to subdomain or fqdn to distinguish 

### How to test
Currently integration is only set up in `/dev` route. You don't need any environment setup to test. 

I'm not sure how DNS handling should be initiated after DB update (same frontend call or background job). 
@dadolhay This update requires DB table ID in `addDnsRequest` call in your function. Please let me know if you have any concern with this approach.


humphd:
@Genne23v it would have been better to separate the new functionality you've added from the refactoring.  Is there any way to split this PR into smaller PRs?

If not, I'll review it as is.  But this isn't ideal.

Genne23v:
I can separate. I will push an update soon.

Genne23v:
@humphd  I have removed new functionalities although they are very small. I would appreciate your review.

Genne23v:
@humphd @dadolhay I connected DNS module with frontend. It's functional, but it needs more backend error validation and display the error somewhere in frontend. Let me know if anything needs to be added.

app/routes/__index/domains/new.tsx
```diff
@@ -45,16 +47,28 @@ export const action = async ({ request }: ActionArgs) => {
...
-  // Create a pending record... for now
-  // This will most likely be replaced by some other logic
-  await addDnsRequest({
-    username: user.username,
-    type: data.type,
-    name: data.name,
-    value: data.value,
-  });
+  try {
+    const recordId = await createUserRecord({
```
dadolhay, line 51: This complexity should be abstracted away from the rest of the system. You could move this into the addDnsRequest flow creator (not in a worker, in the flow creator itself)

app/routes/dev.tsx
```diff
@@ -12,8 +12,14 @@ import logger from '~/lib/logger.server';
 import { requireUser, requireUsername } from '~/session.server';
 import { addNotification } from '~/queues/notifications/notifications.server';
 import { addCertRequest } from '~/queues/certificate/certificate-flow.server';
+import { addDnsRequest } from '~/queues/dns/add-record-flow.server';
+import { updateDnsRequest } from '~/queues/dns/update-record-flow.server';
+import { deleteDnsRequest } from '~/queues/dns/delete-record-flow.server';
+import { createUserRecord, updateUserRecord } from '~/lib/record-db.server';
```
dadolhay, line 18: Both create and update flow creator should handle the required DB changes. This should be abstracted away.

app/lib/record-db.server.ts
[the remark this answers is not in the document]
Genne23v: I see the point. I think showing the right error message is more important than any minor benefit by having this file.

app/queues/dns/workers/poll-dns-status-worker.server.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const checkDnsStatusQueueName = 'check-dns-status';
+
+export const checkDnsStatusWorker = new Worker(
+  checkDnsStatusQueueName,
+  async (job) => {
+    logger.debug('Checking Route53 readiness');
+
+    const childJobValues: { [jobKey: string]: string } = await job.getChildrenValues();
+    const dnsRecordUpdateJobKey = Object.keys(childJobValues).find((item) =>
```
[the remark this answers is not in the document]
Genne23v: Yeah, I saw the code. Thanks for reminding

app/queues/certificate/order-creator-worker.server.ts
```diff
@@ -40,8 +40,9 @@ const handleChallenges = ({
     await addDnsRequest({
       username,
       type: 'TXT',
-      name: domain,
+      subdomain: domain,
```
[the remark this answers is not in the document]
Genne23v: I declared the type as following. 
```js
export type Name = { subdomain: string } | { fqdn: string };
interface BaseAddDnsRequestData {
  username: Record['username'];
  type: Record['type'];
  value: Record['value'];
}
export type AddDnsRequestData = BaseAddDnsRequestData & Name;
```
But the behaviour is strange. It accepts `subdomain` in frontend, but not accepting `subdomain` in backend. Am I using it wrong?

Genne23v:
@dadolhay @humphd I updated the PR. I would appreciate another round of review!

app/queues/dns/add-record-flow.server.ts
```diff
@@ -0,0 +1,105 @@
...
+import { checkDnsStatusQueueName } from './workers/poll-dns-status-worker.server';
+import { syncDbStatusQueueName } from './workers/sync-db-status-worker.server';
+import { createUserRecord } from '~/models/record.server';
+
+import type { FlowJob } from 'bullmq';
+import type { Record } from '@prisma/client';
+import type { DnsUpdaterData } from './workers/dns-update-worker.server';
+import type { DbRecordSynchronizerData } from './workers/sync-db-status-worker.server';
+
+export type Name = { subdomain: string } | { fqdn: string };
```
humphd, line 14: `Name` is really generic, can we do better?  `DomainName`, or `DomainNameComponent` or something, which could be a `subdomain` or `fqdn`?  Doesn't have to be that, but `Name` is not going to work.  @dadolhay will have thoughts.

app/queues/dns/add-record-flow.server.ts
```diff
@@ -0,0 +1,105 @@
...
+const flowProducer = new FlowProducer({ connection: redis });
+
+export const addDnsRequest = async (data: AddDnsRequestData) => {
+  const { username, type, value } = data;
+
+  let subdomain: string;
+  if ('subdomain' in data) {
+    subdomain = data.subdomain;
+  } else {
+    throw new Error('fqdn is not expected');
```
humphd, line 40: You could use [`tiny-invariant`](https://www.npmjs.com/package/tiny-invariant) (already in our deps) here:

```ts
invariant('subdomain' in data, 'expected subdomain vs. fqdn');
const subdomain = data.subdomain;
```

app/queues/dns/delete-record-flow.server.ts
```diff
@@ -0,0 +1,89 @@
...
+  value: Record['value'];
+}
+export type DeleteDnsRequestData = BaseDeleteDnsRequestData & Name;
+
+const flowProducer = new FlowProducer({ connection: redis });
+
+export const deleteDnsRequest = async (data: DeleteDnsRequestData) => {
+  const { id, username, type, value } = data;
+
+  let subdomain: string;
```
humphd, line 28: Same deal here, use `tiny-invariant` to throw if your condition is not met.

app/queues/dns/delete-record-flow.server.ts
```diff
@@ -0,0 +1,89 @@
...
+import { syncDbStatusQueueName } from './workers/sync-db-status-worker.server';
+import { WorkType } from './add-record-flow.server';
+
+import type { FlowJob } from 'bullmq';
+import type { Record } from '@prisma/client';
+import type { DnsUpdaterData } from './workers/dns-update-worker.server';
+import type { DbRecordSynchronizerData } from './workers/sync-db-status-worker.server';
+import type { Name } from './add-record-flow.server';
+
+interface BaseDeleteDnsRequestData {
```
humphd, line 15: Instead of needing a name like this, you can do it all at once:

```ts
export type DeleteDnsRequestData = {
  id: Record['id'];
  username: Record['username'];
  type: Record['type'];
  value: Record['value'];
} & Name;
```

app/queues/dns/update-record-flow.server.ts
```diff
@@ -0,0 +1,108 @@
...
+interface BaseUpdateDnsRequestData {
+  id: Record['id'];
+  username: Record['username'];
+  type: Record['type'];
+  value: Record['value'];
+  description?: Record['description'];
+  course?: Record['course'];
+  ports?: Record['ports'];
+}
+export type UpdateDnsRequestData = BaseUpdateDnsRequestData & Name;
```
humphd, line 26: Same thing here, inline this type and get rid of the `Base*` naming

app/queues/dns/update-record-flow.server.ts
```diff
@@ -0,0 +1,108 @@
...
+
+const flowProducer = new FlowProducer({ connection: redis });
+
+export const updateDnsRequest = async (data: UpdateDnsRequestData) => {
+  const { username, type, value, id } = data;
+
+  let subdomain: string;
```
humphd, line 33: tiny-invariant

app/queues/dns/workers/dns-update-worker.server.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export interface DnsUpdaterData {
+  workType: WorkType;
+  username: Record['username'];
+  type: Record['type'];
+  fqdn: string;
+  value: Record['value'];
+}
+
+export type dnsRecordUpdateJobResult = string;
```
humphd, line 20: I think this should start with a `D` vs. `d`

app/queues/dns/workers/poll-dns-status-worker.server.ts
```diff
@@ -0,0 +1,39 @@
+import { Worker } from 'bullmq';
+import { redis } from '~/lib/redis.server';
+import logger from '~/lib/logger.server';
+import { getChangeStatus } from '~/lib/dns.server';
+import { dnsUpdateQueueName } from './dns-update-worker.server';
+import { getChildrenValuesOfQueueName } from '~/utils';
+
+export const checkDnsStatusQueueName = 'check-dns-status';
+
+// interface PollDnsStatus {
```
humphd, line 10: Can we remove this?

app/queues/dns/workers/poll-dns-status-worker.server.ts
```diff
@@ -0,0 +1,39 @@
...
+    logger.debug('Checking Route53 readiness');
+
+    const childrenValues = await getChildrenValuesOfQueueName<string>({
+      queueName: dnsUpdateQueueName,
+      job,
+    });
+
+    const [changeId] = Object.values(childrenValues);
+
+    // if (!dnsUpdateJobKey) {
```
humphd, line 26: Why is this here?

app/routes/__index/domains/new.tsx
```diff
@@ -45,16 +46,20 @@ export const action = async ({ request }: ActionArgs) => {
...
+      type: data.type,
+      subdomain: data.name,
+      value: data.value,
+    });
 
-  return redirect(`/domains`);
+    return redirect(`/domains`);
+  } catch (error) {
+    logger.warn('Add DNS request error', error);
+    //Need to display an error response
```
humphd, line 60: Please file that as an issue to fix later.

app/queues/dns/update-record-flow.server.ts
```diff
@@ -0,0 +1,108 @@
+import { FlowProducer } from 'bullmq';
+import { redis } from '~/lib/redis.server';
+import { buildDomain } from '~/utils';
+import { dnsUpdateQueueName } from './workers/dns-update-worker.server';
+import { checkDnsStatusQueueName } from './workers/poll-dns-status-worker.server';
+import { syncDbStatusQueueName } from './workers/sync-db-status-worker.server';
+import { WorkType } from './add-record-flow.server';
+import type { RecordUpdate } from '~/models/record.server';
```
humphd, line 8: Move this down to the types imports below

app/queues/dns/update-record-flow.server.ts
```diff
@@ -0,0 +1,108 @@
...
+    name: subdomain,
+    value: data.value,
+    username: data.username,
+    description: data.description,
+    course: data.course,
+    ports: data.ports,
+  };
+
+  // Before running workflow, update the record in DB
+  await updateRecordById(dbRecordUpdate);
```
humphd, line 53: If you avoid declaring the object before you use it, you don't need a type:

```ts
// {...} will get typed automatically, and you don't have to import/set the type
await updateRecordById({...});
```

app/models/record.server.ts
```diff
@@ -1,59 +1,77 @@
 import { RecordStatus } from '@prisma/client';
-import { prisma } from '~/db.server';
 import dayjs from 'dayjs';
+import { prisma } from '~/db.server';
 
 import type { Record } from '@prisma/client';
 export type { Record } from '@prisma/client';
 
-export async function getRecordsByUsername(username: Record['username']) {
+export interface RecordCreation {
```
humphd, line 8: These two types are clunky.  A type should describe a thing vs an action.

Instead of these names, can you:

1. Use inline types in this file, like `Pick<Record, 'username' | 'type' | 'name' | 'value'>`
2. Instead of using the exported type in other files, inline the objects you pass (I have a note below showing you what I mean)

humphd (changes requested):
I've made an initial set of comments.  Looking good.

app/models/record.server.ts
```diff
@@ -1,59 +1,77 @@
...
+  name: Record['name'];
+  value: Record['value'];
+}
+
+export interface RecordUpdate {
+  id: Record['id'];
+  type: Record['type'];
+  name: Record['name'];
+  value: Record['value'];
+  username?: Record['username'];
```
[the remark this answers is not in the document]
Genne23v: This is Data type only for DB update. So the username is not mandatory. I hope it makes sense. If this is not aligned with frontend or something, I will update.

app/models/record.server.ts
```diff
@@ -1,59 +1,77 @@
...
-      name,
-      value,
-      status,
-      username,
-      description,
-      course,
-      ports,
-      expiresAt,
-      lastNotified,
+      ...data,
```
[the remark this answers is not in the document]
Genne23v: Thank you for the recommendation. Will be updated!

app/queues/dns/add-record-flow.server.ts
Genne23v: It's been one of the most difficult tasks for me to name things related to domain. I would like to suggest `DomainType` as it's either partial domain of full domain.

app/queues/dns/delete-record-flow.server.ts
Genne23v: It's been so great for me to learn all these typing techniques. Thank you

app/queues/dns/workers/dns-update-worker.server.ts
Genne23v: Will be fixed

app/queues/dns/workers/poll-dns-status-worker.server.ts
Genne23v: Sorry that I forgot removing the comments

app/routes/__index/domains/new.tsx
Genne23v: I wrote #364

app/models/record.server.ts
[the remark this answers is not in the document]
humphd: It's fine.  Why would you ever include it if it's not required?  I'd remove it if it's optional.

app/queues/dns/add-record-flow.server.ts
humphd: > "There are only two hard things in Computer Science: cache invalidation and naming things." --Phil Karlton (Netscape)

You're a true programmer when you hit the hardest problems in CS. Welcome!

The more you can avoid needing to name things, the fewer names you have to come up with.

I don't think `DomainType` is wrong, but `DomainName` or even `Domain` are better imho, since both a `subdomain` and a `fully qualified domain name` are both `domains`.

dadolhay: I'd drobably name this `FqdnOrSubdomain`

dadolhay: and yes, separating stuff, race conditions, flaky tests, finding others` [monkey patches](https://en.wikipedia.org/wiki/Monkey_patch) and naming stuff are some of the hardest things in sw development

Genne23v: There could be many different opinions, but to me `FqdnOrSubdomain` makes sense a lot. Thanks for the suggestion.

app/queues/dns/add-record-flow.server.ts
```diff
@@ -0,0 +1,103 @@
...
+import { syncDbStatusQueueName } from './workers/sync-db-status-worker.server';
+import { createUserRecord } from '~/models/record.server';
+
+import type { FlowJob } from 'bullmq';
+import type { Record } from '@prisma/client';
+import type { DnsUpdaterData } from './workers/dns-update-worker.server';
+import type { DbRecordSynchronizerData } from './workers/sync-db-status-worker.server';
+
+export type FqdnOrSubdomain = { subdomain: string } | { fqdn: string };
+interface BaseAddDnsRequestData {
```
dadolhay, line 16: Can we use pick here too? Genuine question, I really don't know because of the optional attributes

app/queues/dns/add-record-flow.server.ts
```diff
@@ -0,0 +1,103 @@
...
+  update = 'update',
+  delete = 'delete',
+}
+
+const flowProducer = new FlowProducer({ connection: redis });
+
+export const addDnsRequest = async (data: AddDnsRequestData) => {
+  const { username, type, value } = data;
+
+  invariant('subdomain' in data, 'expected subdomain vs. fqdn');
```
dadolhay, line 37: If we are throwing an exception if fqdn is provided, then there is no point in accepting it in the first place.

If the whole system gravitates toward only working with a subdomain, then let's not accept fqdn here, I'll alter the dns worker send a subdomain instead

app/queues/dns/delete-record-flow.server.ts
```diff
@@ -0,0 +1,86 @@
...
+  type: Record['type'];
+  value: Record['value'];
+} & FqdnOrSubdomain;
+
+const flowProducer = new FlowProducer({ connection: redis });
+
+export const deleteDnsRequest = async (data: DeleteDnsRequestData) => {
+  const { id, username, type, value } = data;
+
+  invariant('subdomain' in data, 'expected subdomain vs. fqdn');
```
dadolhay, line 28: same as above

app/queues/dns/update-record-flow.server.ts
```diff
@@ -0,0 +1,101 @@
...
+  course?: Record['course'];
+  ports?: Record['ports'];
+} & FqdnOrSubdomain;
+
+const flowProducer = new FlowProducer({ connection: redis });
+
+export const updateDnsRequest = async (data: UpdateDnsRequestData) => {
+  const { username, type, value, id } = data;
+
+  invariant('subdomain' in data, 'expected subdomain vs. fqdn');
```
dadolhay, line 32: same here

app/queues/dns/workers/dns-update-worker.server.ts
```diff
@@ -0,0 +1,49 @@
...
+import type { WorkType } from '../add-record-flow.server';
+import type { Record } from '@prisma/client';
+
+export const dnsUpdateQueueName = 'dns-record-update';
+
+export interface DnsUpdaterData {
+  workType: WorkType;
+  username: Record['username'];
+  type: Record['type'];
+  fqdn: string;
```
dadolhay, line 16: If we rely on subdomain in the whole process, let's use that here too, and generate fqdn from the username and subdomain here. Should simplify the flow creator a bit

dadolhay (changes requested):

app/queues/dns/add-record-flow.server.ts
Genne23v: I'm not sure how much I understand the whole system. But I think there would not be many places using `fqdn`.

app/queues/dns/add-record-flow.server.ts
```diff
@@ -0,0 +1,94 @@
...
+import { checkDnsStatusQueueName } from './workers/poll-dns-status-worker.server';
+import { syncDbStatusQueueName } from './workers/sync-db-status-worker.server';
+import { createUserRecord } from '~/models/record.server';
+
+import type { FlowJob } from 'bullmq';
+import type { Record } from '@prisma/client';
+import type { DnsUpdaterData } from './workers/dns-update-worker.server';
+import type { DbRecordSynchronizerData } from './workers/sync-db-status-worker.server';
+
+export type SubDomain = { subdomain: string };
```
humphd, line 14: `Subdomain` vs. `SubDomain` please.

app/queues/dns/update-record-flow.server.ts
```diff
@@ -0,0 +1,91 @@
...
+import { updateRecordById } from '~/models/record.server';
+
+import type { Record } from '@prisma/client';
+import type { FlowJob } from 'bullmq';
+import type { DnsUpdaterData } from './workers/dns-update-worker.server';
+import type { DbRecordSynchronizerData } from './workers/sync-db-status-worker.server';
+import type { SubDomain } from './add-record-flow.server';
+
+export type UpdateDnsRequestData = Pick<Record, 'id' | 'username' | 'type' | 'value'> &
+  Partial<Pick<Record, 'description' | 'course' | 'ports'>> &
```
humphd, line 17: You've become our resident TS master!  I love it.

app/queues/dns/workers/poll-dns-status-worker.server.ts
```diff
@@ -0,0 +1,30 @@
+import { Worker } from 'bullmq';
+import { redis } from '~/lib/redis.server';
+import logger from '~/lib/logger.server';
+import { getChangeStatus } from '~/lib/dns.server';
+import { dnsUpdateQueueName } from './dns-update-worker.server';
+import { getChildrenValuesOfQueueName } from '~/utils';
+
+export const checkDnsStatusQueueName = 'check-dns-status';
+
+export const pollDnsStatusWorker = new Worker(
```
humphd, line 10: How come you don't give this `Worker` a type like all the others?

humphd (commented):
Two small things.  This is looking so much better, great work.

app/queues/dns/update-record-flow.server.ts
Genne23v: I feel I'm lucky to go through all this!

humphd (approved):

humphd:
Only took 2 days, not bad!

Issue #238, opened 2023-02-17
Test for CRUD operations in models


cychu42:
As part of #97, we need to have some tests for CRUD functions in PR #228

humphd:
@cychu42 does this still need to be open?

PR #350, opened 2023-03-15
Setup test database for running tests for database

cychu42:
Part of #238
This sets up a test database `starchart_test` in the container `test_database`, and it can be used to run tests without interfering with the other database container or any other database.

For convenience, I added a `.env.test` file, for devs to run `cp .env.test .env` to change the database connection string to the test database.

## Changes
* Add test database in in `docker-compose.yml`
* Add  `.env.test`
* Ignore the volume at `/mysql-data-test` for git, Prettier, ESlint, and Docker.

## Testing
1. Run `cp .env.test .env`
2. Run `docker compose up -d` to make containers like usual
3. Run `npm run steup` to connect, push schema, and seed the database
4. Run `npm run db:studio` to check that the database works


cychu42:
I wonder if there's some issue with the what GitHub action is using to run hadolint. Saw the same test error in Won's PR that has nothing to do with Dockerfile.

humphd:
I filed https://github.com/hadolint/hadolint-action/issues/79 on the hadolint-action failures.

cychu42:
Looks like it's fixed.

docker-compose.yml
```diff
@@ -17,6 +17,21 @@ services:
...
       MYSQL_ROOT_PASSWORD: root_password
 
+  test_database:
+    image: mysql:8
+    container_name: test_database
+    ports:
+      # Only for dev to run tests in a database
+      - 3307:3306
+    volumes:
+      # Persist data to ./mysql-data-test. Remove if you want to clear the db.
```
humphd, line 27: Since this is for testing, why don't you remove this `volumes` block altogether, and not have it persist.  After the container is brought down, the data would be gone.  This seems like what you want for testing.

docker-compose.yml
```diff
@@ -17,6 +17,21 @@ services:
       MYSQL_PASSWORD: starchart_password
       MYSQL_ROOT_PASSWORD: root_password
 
+  test_database:
```
humphd, line 20: I don't love that we're going to have to run 2 instances of MySQL in development.  Why not run 2 databases in the same instance?

See https://stackoverflow.com/questions/39204142/docker-compose-with-multiple-databases.

humphd (changes requested):

docker-compose.yml
cychu42: Make sense. Let me look into it.

.env.example
```diff
@@ -2,7 +2,7 @@
 # NOTE: this value is also mirrored in dev-secrets/DATABASE_URL.
 # We set it here for Prisma scripts to work, which expect it
 # in the env, but at runtime we read it as a secret in docker swarm.
-DATABASE_URL="mysql://starchart:starchart_password@127.0.0.1:3306/starchart"
+DATABASE_URL="mysql://root:root_password@127.0.0.1:3306/starchart"
```
humphd, line 5: You also need to update `dev-secrets/DATABASE_URL`

mysql/init/01-databases.sql
```diff
@@ -0,0 +1,7 @@
+-- create regular and test databases
```
humphd, line 1: Move this file to `config/` so we don't have to put `mysql` into git

docker-compose.yml
```diff
@@ -9,12 +9,11 @@ services:
       # Only for dev
       - 3306:3306
     volumes:
-      # Persist data to ./mysql-data. Remove if you want to clear the db.
-      - ./mysql-data:/var/lib/mysql
+      # Persist data to ./mysql/data. Remove if you want to clear the db.
+      - ./mysql/data:/var/lib/mysql
+      # Locate SQL file to setup database
+      - ./mysql/init:/docker-entrypoint-initdb.d
```
humphd, line 15: `./config/01-databases.sql:/docker-entrypoint-initdb.d/init/01-databases.sql`

docker-compose.yml
```diff
@@ -9,12 +9,11 @@ services:
       # Only for dev
       - 3306:3306
     volumes:
-      # Persist data to ./mysql-data. Remove if you want to clear the db.
-      - ./mysql-data:/var/lib/mysql
+      # Persist data to ./mysql/data. Remove if you want to clear the db.
+      - ./mysql/data:/var/lib/mysql
```
humphd, line 13: You can leave this how it was, if we mount the `.sql` from config`

humphd (changes requested):
Let's not rename `mysql-data`.  We can move the `.sql` you created into `config/` and then mount it where it needs to go in Docker.

For tests vs. running dev, the app uses `dev-secrets/DATABASE_URL`.  How will we deal with differences here?

Ask @dadolhay if there is anything to know about with respect to using the root user like this.  I think it's fine, but I'm not an expert.

Also, do you want to blow-away the test-db every time you init in your `.sql`?  It might make it easier to run tests.

mysql/init/01-databases.sql
cychu42: I can do that and put it under its own folder in `/config`, just because I don't want the database to load any other config files.

cychu42:
> For tests vs. running dev, the app uses dev-secrets/DATABASE_URL. How will we deal with differences here?
I'm not sure I understand. Is updating `dev-secrets/DATABASE_URL` not enough?

cychu42:
Looking at the test and `dev` scripts, we use `SECRETS_OVERRIDE=1` to use what's inside `/dev-secrets` for testing and development.

cychu42:
How about this? I applied changes according to your feedbacks, but I keep the original setup for `starchart` database intact in `docker-compose.yml` so we don't have to worry about changing much (including the secrets). I just add code relevant to the SQL file to make that test database.
If someone wants to switch to test database, they just run `cp .env.test .env`, and then run `npm run setup` or `npm prisma db push` if the schema isn't there already.

In my mind, I'm not sure it's necessary to drop the test database for testing, since tests should probably clean the test database before running tests, to avoid interference between each tests or any manual testing.

I haven't heard from Denes. Given both `startchart` user and `root` user have superuser access, I don't expect there to be a difference, and `root` user is now only used for the test database anyway.

humphd (commented):
I like this.  The only thing I'd like to see improved is not having a duplicate `.env` in here.  If we have a lot of these files, people have to remember to update them.  Can we eliminate the need for that?

Obviously people running the tests (e.g., CI) would need to set this value to override it.

I don't want to make this impossible, but I also don't want more maintenance hassle.

cychu42:
The `.env.test` is just an alternative to manually replacing the connection string.
I guess we can tell people to do that in CONTRIBUTION.md with copy/paste, unless we want to start setting the connection string env variable in every test script.

cychu42:
I guess this is cleaner than asking people to changing connection string back and forth by hand. Would definitely need a mention in `CONTRIBUTION.md` to mention this.

humphd (approved):

SerpentBytes (approved):

CONTRIBUTING.md
```diff
@@ -59,6 +59,8 @@ $ npm run db:studio
 
 > **Note** `npm run build` needs to be executed the first time running the project. As it generates a `build/server.js` script that `npm run dev` depends on. Subsequent times, only `npm run dev` is needed to run the app in development mode.
 
+> **Note** If you are running test scripts for the first time, change `DATABASE_URL` in `.env` to `DATABASE_URL="mysql://starchart:starchart_password@127.0.0.1:3306/starchart"` and run `npm run setup` to setup the test database. You can change this back afterward.
```
Eakam1007, line 62: This doesn't need to be done in this PR, but would it be worth to add a new `setup:test` script to do this instead?

cychu42: Perhaps. If someone can write a command to parse the file to change the variable, run `npm run setup`, then change it back.

Eakam1007: Ah I think we can simply copy the `setup` command to `setup:test`, and override the value for `DATABASE_URL`

Eakam1007: I think I might just create a PR for this today. Seems simple enough (unless I am missing something)

Issue #342, opened 2023-03-14
Clean-up old BullMQ jobs automatically


humphd:
BullMQ won't automatically remove old jobs when they complete or fail.  We have to [configure it to do this automatically](https://docs.bullmq.io/guide/queues/auto-removal-of-jobs).

We should probably [remove all jobs after 7 days](https://docs.bullmq.io/guide/queues/auto-removal-of-jobs#keep-jobs-based-on-their-age) or something

PR #357, opened 2023-03-16
Add Worker/Queue to process expired DNS records

### LINKED ISSUE #112: Expire DNS records after a configurable amount of time

SerpentBytes:
## Description
Resolves #112 by adding Worker/Queue to do the following:
1. Query the database for expired DNS records
2. If Step 1. yields any results, delete DNS Records from Route53 and delete the record from DB (via `deleteDnsRequest`)
3. Send Notification 

Step 2 is not working as intended. Invoking `deleteDnsRequest` in the front-end/back-end fails with the following warning: 
```bash
warn: Invalid request: Expected exactly one of [AliasTarget, all of [TTL, and ResourceRecords], or TrafficPolicyInstanceId], but found none in Change with [Action=DELETE, Name=osd700-a1.user1.starchart.com, Type=A, SetIdentifier=null]
```
Possibly linked to the missing [TTL](https://github.com/boto/boto3/issues/3161) value.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,3 +30,9 @@ SMTP_PORT=1025
 APP_URL=http://host.docker.internal:8080
 # The SimpleSAML IDP's XML metadata
 SAML_IDP_METADATA_PATH=config/idp-metadata-dev.xml
+
+# Background jobs
+# 24 * 60 * 60 * 1000 = 604800000 (24 hours)
+EXPIRATION_REPEAT_FREQUENCY_MS=86400000
+# 7 * 24 * 60 * 60 * 1000 = 604800000 (7 days)
+JOB_REMOVAL_FREQUENCY_MS=604800000
--- a/app/queues/common/expiration-request.server.ts
+++ b/app/queues/common/expiration-request.server.ts
@@ -1,0 +1,60 @@
+import { Worker, Queue, UnrecoverableError } from 'bullmq';
+
+import { redis } from '~/lib/redis.server';
+import logger from '~/lib/logger.server';
+
+import { getExpiredRecords } from '~/models/record.server';
+import { addNotification } from '../notifications/notifications.server';
+import { deleteDnsRequest } from '../dns/delete-record-flow.server';
+
+const { EXPIRATION_REPEAT_FREQUENCY_MS, JOB_REMOVAL_FREQUENCY_MS } = process.env;
+
+// constant  for removing job on completion/failure (in milliseconds)
+const JOB_REMOVAL_INTERVAL_MS = 7 * 24 * 60 * 60 * 1000; // 7 days
+declare global {
+  var __expiration_request_init__: boolean;
+}
+// queue name
+const expirationRequestQueueName = 'expiration-request';
+
+// queue initialization
+const expirationRequestQueue = new Queue(expirationRequestQueueName, {
+  connection: redis,
+});
+
+export function addExpirationRequest() {
+  return expirationRequestQueue.add(expirationRequestQueueName, {
+    repeat: { every: Number(EXPIRATION_REPEAT_FREQUENCY_MS) || 24 * 60 * 60 * 1000 },
+    removeOnComplete: { age: Number(JOB_REMOVAL_FREQUENCY_MS) || JOB_REMOVAL_INTERVAL_MS },
+    removeOnFail: { age: Number(JOB_REMOVAL_FREQUENCY_MS) || JOB_REMOVAL_INTERVAL_MS },
+  });
+}
+
+// worker definition
+const expirationRequestWorker = new Worker(
+  expirationRequestQueueName,
+  async (job) => {
+    try {
+      logger.info('process DNS record expiration');
+      let dnsRecords = await getExpiredRecords();
+      Promise.all(
+        dnsRecords.map(async ({ id, username, type, subdomain, value, user }) => {
+          // delete records from Route53 and DB
+          await deleteDnsRequest({ id, username, type, subdomain, value });
+          // add notification jobs (assuming deletion went successfully)
+          await addNotification({
+            emailAddress: user.email,
+            subject: 'DNS record expiration subject',
+            message: 'DNS record expiration message',
+          });
+        })
+      );
+    } catch (err) {
+      throw new UnrecoverableError(`Unable to process DNS record expiration: ${err}`);
+    }
+    logger.info('TODO: process certificate expiration');
+  },
+  { connection: redis }
+);
+
+process.on('SIGINT', () => expirationRequestWorker.close());
--- a/app/models/record.server.ts
+++ b/app/models/record.server.ts
@@ -112,3 +112,16 @@ export function renewDnsRecordById(id: Record['id']) {
     },
   });
 }
+
+export function getExpiredRecords() {
+  return prisma.record.findMany({
+    where: {
+      expiresAt: {
+        lt: new Date(),
+      },
+    },
+    include: {
+      user: true,
+    },
+  });
+}
--- a/app/queues/notifications/expiration-notification.server.ts
+++ b/app/queues/notifications/expiration-notification.server.ts
@@ -14,12 +14,11 @@ import type { NotificationData } from './notifications.server';
 declare global {
   var __expiration_init__: boolean;
 }
-
-enum RecordType {
+export enum RecordType {
   Certificate = 'certificate',
   DnsRecord = 'record',
 }
-interface ExpirationStatusPayload {
+export interface ExpirationStatusPayload {
   type: RecordType;
 }
 // constant for notification frequency in days
--- a/app/queues/notifications/notifications.server.ts
+++ b/app/queues/notifications/notifications.server.ts
@@ -3,12 +3,31 @@ import { Queue, Worker } from 'bullmq';
 import { redis } from '~/lib/redis.server';
 import logger from '~/lib/logger.server';
 import sendNotification from '~/lib/notifications.server';
+import { addExpirationRequest } from '../common/expiration-request.server';
 
 export type NotificationData = {
   emailAddress: string;
   subject: string;
   message: string;
 };
+
+async function init() {
+  try {
+    logger.debug('Expiration Requests init: adding jobs for certificate/record expiration');
+    await addExpirationRequest();
+  } catch (err) {
+    logger.error(`Unable to start expiration notification workers: ${err}`);
+  }
+}
+if (process.env.NODE_ENV === 'production') {
+  init();
+} else {
+  // Only do this setup once in dev
+  if (!global.__expiration_request_init__) {
+    init();
+    global.__expiration_request_init__ = true;
+  }
+}
 
 /**
  * This is the main way callers interact with the notifications


app/queues/common/expiration-request.server.ts
```diff
@@ -0,0 +1,105 @@
...
+const expirationRequestWorker = new Worker<ExpirationRequestPayload>(
+  expirationRequestQueueName,
+  async (job) => {
+    const { type } = job.data;
+    switch (type) {
+      case RecType.DnsRecord:
+        try {
+          logger.info(`process DNS record expiration`);
+          let dnsRecords = await getExpiredRecords(type);
+          if (dnsRecords?.length) {
```
humphd, line 75: This is unnecessary, you can `await Promise.all([])` and it will be ok.

app/queues/common/expiration-request.server.ts
```diff
@@ -0,0 +1,105 @@
...
+            await Promise.all(
+              dnsRecords.map(async (dnsRecord) => {
+                const { id, username, type, name, value, user } = dnsRecord;
```
humphd, line 78: Destructure in the args and skip a step

```ts
dnsRecords.map(async ({ id, username, type, name, value, user }) => {
```

app/queues/common/expiration-request.server.ts
```diff
@@ -0,0 +1,105 @@
...
+import logger from '~/lib/logger.server';
+import { prisma } from '~/db.server';
+
+import { addNotification } from '../notifications/notifications.server';
+import { deleteDnsRequest } from '../dns/dns-flow.server';
+import { RecordType as RecType } from '../notifications/expiration-notification.server';
+
+import type { ExpirationStatusPayload as ExpirationRequestPayload } from '../notifications/expiration-notification.server';
+
+const { EXPIRATION_JOB_REPEAT_FREQUENCY, JOB_REMOVAL_FREQUENCY } = process.env;
```
humphd, line 13: Set defaults on these so we don't have to provide in env if we don't want to override.

app/queues/common/expiration-request.server.ts
```diff
@@ -0,0 +1,105 @@
+import { Worker, Queue, UnrecoverableError } from 'bullmq';
+
+import { redis } from '~/lib/redis.server';
+import logger from '~/lib/logger.server';
+import { prisma } from '~/db.server';
+
+import { addNotification } from '../notifications/notifications.server';
+import { deleteDnsRequest } from '../dns/dns-flow.server';
+import { RecordType as RecType } from '../notifications/expiration-notification.server';
```
humphd, line 9: Why change the name here?

app/queues/common/expiration-request.server.ts
```diff
@@ -0,0 +1,105 @@
...
+const getExpiredRecords = (type: RecType) => {
+  switch (type) {
+    case RecType.DnsRecord:
+      return prisma.record.findMany({
+        where: {
+          expiresAt: {
+            lt: new Date(),
+          },
+        },
+        select: {
```
humphd, line 30: This probably isn't needed (i.e., you can get all the data vs trimming)

humphd (changes requested):
This is looking good.  A few things I noticed on first reading.

SerpentBytes:
@humphd, I have updated the code based on your initial feedback.

app/queues/common/expiration-request.server.ts
SerpentBytes: When I tried importing RecordType, I saw that the variable name was not unique; it's used in some Prisma code files and other places. To avoid any future naming conflicts, I thought to rename it.

app/queues/common/expiration-request.server.ts
SerpentBytes: Agreed.

app/queues/common/expiration-request.server.ts
humphd: got it.

app/queues/common/expiration-request.server.ts
```diff
@@ -0,0 +1,100 @@
...
+  connection: redis,
+});
+
+export function addExpirationRequest(type: RecordType) {
+  const jobName = `expirationRequestQueueName-${type}`;
+  return expirationRequestQueue.add(
+    jobName,
+    { type },
+    {
+      repeat: { every: Number(EXPIRATION_REPEAT_FREQUENCY) || 24 * 60 * 60 * 1000 },
```
humphd, line 54: I don't think you can use `Number()` here, since this is a cron string, right?

app/queues/common/expiration-request.server.ts
SerpentBytes: For now, it's not causing a naming conflict, so I have removed the alias.

app/queues/common/expiration-request.server.ts
```diff
@@ -0,0 +1,100 @@
...
+      removeOnComplete: { age: Number(JOB_REMOVAL_FREQUENCY) || JOB_REMOVAL_INTERVAL },
+      removeOnFail: { age: Number(JOB_REMOVAL_FREQUENCY) || JOB_REMOVAL_INTERVAL },
```
humphd, line 56: Same question here

app/queues/common/expiration-request.server.ts
humphd: I'm confused why this is still here?

humphd (changes requested):

app/queues/common/expiration-request.server.ts
SerpentBytes: We are not using the cron pattern anymore since not everyone is familiar with it. Therefore, we are casting the value of the environment variable into a Number, as "every" accepts a value in milliseconds.

app/queues/common/expiration-request.server.ts
SerpentBytes: see: https://github.com/DevelopingSpace/starchart/pull/357#discussion_r1140502435

app/queues/common/expiration-request.server.ts
SerpentBytes: > I'm confused why this is still here?

I don't know what you are referring to here.

app/queues/common/expiration-request.server.ts
humphd: Above you have this:

```ini
# Background jobs
EXPIRATION_REPEAT_FREQUENCY=24 * 60 * 60 * 1000
JOB_REMOVAL_FREQUENCY=7 * 24 * 60  * 60 * 1000
```

Should you change that if you're not using CRON any more?  Doing `Number('7 * 24 * 60  * 60 * 1000')` will give `NaN`.

app/queues/common/expiration-request.server.ts
humphd:
```ts
if (dnsRecords?.length) {
```

humphd (changes requested):

app/queues/common/expiration-request.server.ts
SerpentBytes: Only try deleting if we get results back from a query, or is that just unnecessary since the job will fail?

app/queues/common/expiration-request.server.ts
SerpentBytes: Right, the quotes.

SerpentBytes:
@humphd I have removed the `if` check and updated the notifications job-related `.env` variables values so casting to Number from String does not result in `NaN`.

.env.example
```diff
@@ -30,3 +30,7 @@ SMTP_PORT=1025
 HOSTNAME=http://host.docker.internal:8080
 # The SimpleSAML IDP's XML metadata
 SAML_IDP_METADATA_PATH=config/idp-metadata-dev.xml
+
+# Background jobs
+EXPIRATION_REPEAT_FREQUENCY=86400000
```
humphd, line 35: What you can do here is leave a comment that explains what the humber means:

```
# 24 * 7 * ... = 86400000
EXPIRATION_REPEAT_FREQUENCY=86400000
```

Best of both worlds.

humphd: Also, when you have a number like this, it's good to name it with units: `EXPIRATION_REPEAT_FREQUENCY_SECONDS` or `_MS` etc

app/queues/common/expiration-request.server.ts
```diff
@@ -0,0 +1,98 @@
...
+// worker definition
+const expirationRequestWorker = new Worker<ExpirationRequestPayload>(
+  expirationRequestQueueName,
+  async (job) => {
+    const { type } = job.data;
+    switch (type) {
+      case RecordType.DnsRecord:
+        try {
+          logger.info('process DNS record expiration');
+          let dnsRecords = await getExpiredRecords(type);
```
humphd, line 70: So this can return a non-array?  That makes sense why you were wanting to check the `dnsRecords?.length`:

```ts
const dnsRecords = await getExpiredRecords(type);
if(!dnsRecords) {
  return;
}
return Promise.all(dnsRecords.map(...))
```

humphd (changes requested):
Looks good.  A few things I notice and I think this is ready.

app/queues/common/expiration-request.server.ts
SerpentBytes: Yes, it can be potentially null or undefined.

.env.example
SerpentBytes: Agreed. It's helpful to know what the numbers mean.

SerpentBytes:
Updated the code with helpful comments and added if condition `return` if we get `null/undefined` back from the query that fetches expired DNS records.

app/queues/common/expiration-request.server.ts
```diff
@@ -0,0 +1,101 @@
...
+const { EXPIRATION_REPEAT_FREQUENCY_MS, JOB_REMOVAL_FREQUENCY_MS } = process.env;
+const JOB_REMOVAL_INTERVAL = 7 * 24 * 60 * 60 * 1000;
+declare global {
+  var __expiration_request_init__: boolean;
+}
+// queue name
+const expirationRequestQueueName = 'expiration-request';
+
+// query to fetch expired records from database based on type
+const getExpiredRecords = (type: RecordType) => {
```
dadolhay, line 22: I'd suggest to create two separate functions, one for records, one for certificates. I'd probably move this functionality to the records and certificates model files too.

app/queues/common/expiration-request.server.ts
```diff
@@ -0,0 +1,101 @@
...
+  connection: redis,
+});
+
+export function addExpirationRequest(type: RecordType) {
+  const jobName = `expirationRequestQueueName-${type}`;
+  return expirationRequestQueue.add(
+    jobName,
+    { type },
+    {
+      repeat: { every: Number(EXPIRATION_REPEAT_FREQUENCY_MS) || 24 * 60 * 60 * 1000 },
```
dadolhay, line 54: I'd make these mandatory and throw an exception if they are missing

app/queues/common/expiration-request.server.ts
```diff
@@ -0,0 +1,101 @@
...
+      removeOnComplete: { age: Number(JOB_REMOVAL_FREQUENCY_MS) || JOB_REMOVAL_INTERVAL },
```
dadolhay, line 55: @humphd Should we do this in other queues too? seems like a good idea to clean stuff out of redis after a while

app/queues/common/expiration-request.server.ts
```diff
@@ -0,0 +1,101 @@
...
+      removeOnFail: { age: Number(JOB_REMOVAL_FREQUENCY_MS) || JOB_REMOVAL_INTERVAL },
+    }
+  );
+}
+
+// worker definition
+const expirationRequestWorker = new Worker<ExpirationRequestPayload>(
```
dadolhay, line 62: What is the rationale in combining record and cert expiration? Wouldn't it be cleaner to make them separate?
Or do them both in one queue job (eliminating type)?

Either way would simplify the code quite a bit (especially the second)

dadolhay (changes requested):

app/queues/common/expiration-request.server.ts
SerpentBytes: There's an issue filed already. See #342.

app/queues/common/expiration-request.server.ts
SerpentBytes: Agreed. It would be cleaner to do them in a single queue job.

app/queues/common/expiration-request.server.ts
SerpentBytes: Do we still need to throw an exception if we provide a hard-coded default value?

.env.example
```diff
@@ -30,3 +30,9 @@ SMTP_PORT=1025
 HOSTNAME=http://host.docker.internal:8080
 # The SimpleSAML IDP's XML metadata
 SAML_IDP_METADATA_PATH=config/idp-metadata-dev.xml
+
+# Background jobs
+# 24 * 60 * 60 * 1000 = 604800000 (24 hours)
```
humphd, line 35: Awesome!

app/queues/common/expiration-request.server.ts
```diff
@@ -0,0 +1,101 @@
...
+import { prisma } from '~/db.server';
+
+import { addNotification } from '../notifications/notifications.server';
+import { deleteDnsRequest } from '../dns/delete-record-flow.server';
+import { RecordType } from '../notifications/expiration-notification.server';
+
+import type { ExpirationStatusPayload as ExpirationRequestPayload } from '../notifications/expiration-notification.server';
+
+const { EXPIRATION_REPEAT_FREQUENCY_MS, JOB_REMOVAL_FREQUENCY_MS } = process.env;
+const JOB_REMOVAL_INTERVAL = 7 * 24 * 60 * 60 * 1000;
```
humphd, line 14: Add a `_MS` to this as well, and a comment indicating what this magic number equals in hours/days or whatever.

app/queues/common/expiration-request.server.ts
humphd: Agree, this can move up a level in our API.

app/queues/common/expiration-request.server.ts
humphd: We already have too many config options to maintain.  If we can provide a sane default and override when we want to change, I'd prefer that.

humphd (changes requested):

app/queues/common/expiration-request.server.ts
SerpentBytes: Let me know if [063424f](https://github.com/DevelopingSpace/starchart/pull/357/commits/063424f13bffda2bd529b0e1087f62a2f9d3b110) is the change you were requesting.

humphd (approved):

app/queues/common/expiration-request.server.ts
dadolhay: What I suggest is to do away with `type`, i.e., this `case RecordType.DnsRecord:` would go away, and this worker would first process the DNS records, then would delete expired certificates from the DB, it would not take the "type" attribute at all

dadolhay (dismissed):
Regardless, I'm going to approve this, if we decide to, we can make that change later, or you can incorporate it int this pr as well. Whatever makes more sense to you.

app/queues/common/expiration-request.server.ts
SerpentBytes: Thanks for clarifying.

SerpentBytes:
@dadolhay I misunderstood you earlier. I have updated the code to process DNS records without using `type`.

humphd (approved):

dadolhay (approved):
Looks awesome!

PR #363, opened 2023-03-16
Remove old deps, README not being used

humphd:
I'm pretty sure we don't need any of this.  I'll see if CI passes with out it.

If you think I'm wrong, please let me know.


package.json
```diff
@@ -93,9 +92,7 @@
     "eslint-plugin-playwright": "^0.12.0",
     "happy-dom": "^8.1.5",
     "husky": "^8.0.3",
-    "jest": "^29.4.1",
     "lint-staged": "^13.1.2",
-    "msw": "^0.49.2",
```
Ririio, line 98: If you don't mind me asking what's msw?

Eakam1007 (approved):
Looks good to me, I am pretty sure we don't use any of these

package.json
Eakam1007: Looks like it is a mocking library ([msw](https://www.npmjs.com/package/msw))

humphd:
If something weird breaks in the next week, help me remember that we removed all these, and something might have been needed that we missed.

However, I'm going to try it.

Issue #251, opened 2023-02-21
E2E test for certificate components


Ririio:
We need to create a test for the certificate UI to prevent future commits from breaking it

Issue #173, opened 2023-02-10
Create a test in regards to header page


Ririio:
- User should be redirected to the login page when they press the `sign out` button
  - Ensure user is set to null
- User should be redirected to either the domains or certificate page when they click the link

PR #365, opened 2023-03-17
Test: Certificate, Header

Ririio:
Fixes: #251, Fixes: #173 

This week, I'm creating a playwright test for the front end which includes, header, landing page and certificate

<details>

<summary>The certificate will cycle through 3 pages, and check for values that are specific for those pages</summary>

- Request Page
  - Check if user is currently on the `/certificate` route
  - Check for user domain name through text
  - Expect that the title heading is `Certificate`
- Loading Page
  - Check for the loading page to contain text 
- Key Page
  - Click download/clipboard buttons for private and public key
  - Catch toast pop up through heading

</details>

<details>
<summary>The header contains a hamburger state that is only visible in mobile, thus creating a new file `.mobile` to account for such instances</summary>

- Signout Button
  - Expect to redirect to `/login` route when clicked
- Card Links
  - Mobile
    - Click Hamburger Menu
    - Receive links to certificate and domain pages
    - Expect URL to either be `/domain` or `/certificate`
  - Desktop
    - Click either domain and certificate links
    - Expect URL to either be `/domain` or `/certificate`

</details>

<details>
<summary>On the landing page test we first check before starting if heading "DNS Records" and "Certificate" can be located in the `/` route</summary>

- Card
  - Button 
    - Find a button named "Manage DNS Records" or "Certificate"
    - Expect `/domains` or `/certificate` URL
  - Instruction Link
    - Look for a paragraph, and check if the paragraph contains a text: **to learn more, see these instructions...**
    - Expected URL when clicked `/domains/instructions` `/certificate/instructions`

</details>

We are also adding a `testIgnore` under `playwright.config.ts` to accommodate for instances where values are showing/hiding under different devices

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -34,12 +34,12 @@ const config: PlaywrightTestConfig = {
     /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
     actionTimeout: 0,
     /* Base URL to use in actions like `await page.goto('/')`. */
-    baseURL: `http://localhost:${process.env.PORT || 8080}`,
+    baseURL: 'http://localhost:8080',
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-    trace: 'retain-on-failure',
+    trace: 'on-first-retry',
     /* Record video on first retry */
-    video: 'retain-on-failure',
+    video: 'on-first-retry',
   },
 
   /* Configure projects for major browsers */
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -34,12 +34,12 @@ const config: PlaywrightTestConfig = {
     /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
     actionTimeout: 0,
     /* Base URL to use in actions like `await page.goto('/')`. */
-    baseURL: `http://localhost:${process.env.PORT}`,
+    baseURL: `http://localhost:${process.env.PORT || 8080}`,
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-    trace: 'on-first-retry',
+    trace: 'retain-on-failure',
     /* Record video on first retry */
-    video: 'on-first-retry',
+    video: 'retain-on-failure',
   },
 
   /* Configure projects for major browsers */
@@ -58,6 +58,7 @@ const config: PlaywrightTestConfig = {
         ...devices['Desktop Chrome'],
       },
       dependencies: ['setup'],
+      testIgnore: /.*\.mobile\.spec\.ts/,
     },
 
     {
@@ -66,6 +67,7 @@ const config: PlaywrightTestConfig = {
         ...devices['Desktop Firefox'],
       },
       dependencies: ['setup'],
+      testIgnore: /.*\.mobile\.spec\.ts/,
     },
 
     {
@@ -74,6 +76,7 @@ const config: PlaywrightTestConfig = {
         ...devices['Desktop Safari'],
       },
       dependencies: ['setup'],
+      testIgnore: /.*\.mobile\.spec\.ts/,
     },
 
     /* Test against mobile viewports. */
@@ -83,6 +86,7 @@ const config: PlaywrightTestConfig = {
         ...devices['Pixel 5'],
       },
       dependencies: ['setup'],
+      testIgnore: /.*\.desktop\.spec\.ts/,
     },
     /**
      * We override the isMobile for Mobile Safari to false to make it work in CI
@@ -100,6 +104,7 @@ const config: PlaywrightTestConfig = {
         defaultBrowserType: devices['iPhone 12'].defaultBrowserType,
       },
       dependencies: ['setup'],
+      testIgnore: /.*\.desktop\.spec\.ts/,
     },
 
     /* Test against branded browsers. */
@@ -109,6 +114,7 @@ const config: PlaywrightTestConfig = {
         channel: 'msedge',
       },
       dependencies: ['setup'],
+      testIgnore: /.*\.mobile\.spec\.ts/,
     },
     {
       name: 'Google Chrome',
@@ -116,6 +122,7 @@ const config: PlaywrightTestConfig = {
         channel: 'chrome',
       },
       dependencies: ['setup'],
+      testIgnore: /.*\.mobile\.spec\.ts/,
     },
   ],
 
--- a/test/e2e/certificate.spec.ts
+++ b/test/e2e/certificate.spec.ts
@@ -24,9 +24,7 @@ test.describe('Certificate Page', () => {
       })
       .nth(1);
 
-    await expect(loadingPageText).toContainText(
-      'We have received your request, and will notify you when your certificate is read'
-    );
+    await loadingPageText.waitFor();
 
     //Copy Key Toast Text
     await page.getByRole('button', { name: 'Copy Public Key' }).click();
@@ -35,8 +33,8 @@ test.describe('Certificate Page', () => {
     const publicCopyToastText = page.getByText('Public Key was copied to the clipboard');
     const privateCopyToastText = page.getByText('Private Key was copied to the clipboard');
 
-    await expect(publicCopyToastText).toContainText('Public Key was copied to the clipboard');
-    await expect(privateCopyToastText).toContainText('Private Key was copied to the clipboard');
+    await publicCopyToastText.waitFor();
+    await privateCopyToastText.waitFor();
 
     //Download Key Toast Text
     await page.getByRole('button', { name: 'Download Public Key' }).click();
@@ -45,8 +43,8 @@ test.describe('Certificate Page', () => {
     const publicDownloadToastText = page.getByText('Public Key is Downloaded');
     const privateDownloadToastText = page.getByText('Private Key is Downloaded');
 
-    await expect(publicDownloadToastText).toContainText('Public Key is Downloaded');
-    await expect(privateDownloadToastText).toContainText('Private Key is Downloaded');
+    await publicDownloadToastText.waitFor();
+    await privateDownloadToastText.waitFor();
 
     //Text Titles
     const publicKey = page.getByRole('heading', { name: 'Public Key' });
--- a/test/e2e/certificate.spec.ts
+++ b/test/e2e/certificate.spec.ts
@@ -14,21 +14,41 @@ test.describe('Certificate Page', () => {
 
     await expect(domainName).toContainText('user1.starchart.com');
     await expect(titleHeader).toContainText('Certificate');
+
     await page.getByRole('button', { name: 'Request a Certificate' }).click();
 
-    page
+    const loadingPageText = page
       .locator('div')
       .filter({
         hasText: 'We have received your request, and will notify you when your certificate is read',
       })
       .nth(1);
 
+    await expect(loadingPageText).toContainText(
+      'We have received your request, and will notify you when your certificate is read'
+    );
+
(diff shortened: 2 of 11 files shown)


playwright.config.ts
```diff
@@ -34,12 +34,12 @@ const config: PlaywrightTestConfig = {
     /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
     actionTimeout: 0,
     /* Base URL to use in actions like `await page.goto('/')`. */
-    baseURL: `http://localhost:${process.env.PORT}`,
+    baseURL: `http://localhost:${process.env.PORT || 8080}`,
```
Eakam1007, line 37: Since we are using port 8080 everywhere, I think it would be better to just change this to use a hard coded port
```suggestion
    baseURL: 'http://localhost:8080',
```

playwright.config.ts
```diff
@@ -34,12 +34,12 @@ const config: PlaywrightTestConfig = {
...
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-    trace: 'on-first-retry',
+    trace: 'retain-on-failure',
     /* Record video on first retry */
-    video: 'on-first-retry',
+    video: 'retain-on-failure',
```
Eakam1007, line 42: I don't think we need to change this, since we will get the same end result of a video for every failed test.  
This change will also make it record a video for every test and then discard videos for passed tests, which seems unnecessary. 
Is there a reason you changed this?

test/e2e/login.spec.ts
```diff
@@ -1,29 +1,33 @@
...
-  await page.getByLabel('Username').fill('user1');
-  await page.getByLabel('Password').fill('user1pass');
-  await page.getByRole('button', { name: 'Login' }).click();
-  await page.waitForURL('http://localhost:8080');
-  const locator = page.locator('#header-user');
-  const starchartHeading = page.getByRole('heading', { name: 'My.Custom.Domain' });
-  await expect(locator).toContainText('user1');
-  await expect(starchartHeading).toContainText('My.Custom.Domain');
-});
+test.describe('Login with user', () => {
```
Eakam1007, line 3: This seems unnecessary since the test file is named `login.spec`

Eakam1007 (changes requested):

playwright.config.ts
Ririio: I recall changing this back during our meeting, not sure why it reverted, I'll fix that now

test/e2e/certificate.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    await expect(domainName).toContainText('user1.starchart.com');
+    await expect(titleHeader).toContainText('Certificate');
+    await page.getByRole('button', { name: 'Request a Certificate' }).click();
+
+    page
+      .locator('div')
+      .filter({
+        hasText: 'We have received your request, and will notify you when your certificate is read',
+      })
+      .nth(1);
```
Eakam1007, line 24: Could you please explain what this does? From what I understand, this simply gets the div, and doesn't do anything with it?

test/e2e/landing-page.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+  loggedInAsUser();
+
+  test.beforeEach(async ({ page }) => {
+    await page.goto('/');
+
+    const dnsCard = page.getByRole('heading', { name: 'DNS Records' });
+    const certCard = page.getByRole('heading', { name: 'Certificate' });
+
+    await expect(dnsCard).toContainText('DNS Records');
+    await expect(certCard).toContainText('Certificate');
```
Eakam1007, line 14: This should be its own `test()` block, maybe something like 
```ts
test('contains DNS Record and Certificate cards', ...)
```

test/e2e/landing-page.spec.ts
```diff
@@ -0,0 +1,50 @@
+import { test, expect } from '@playwright/test';
+import { loggedInAsUser } from './utils';
+
+test.describe('landing page link', () => {
```
Eakam1007, line 4:
```suggestion
test.describe('Landing page', () => {
```
I think this is more descriptive

test/e2e/landing-page.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    await page.goto('/');
+
+    const dnsCard = page.getByRole('heading', { name: 'DNS Records' });
+    const certCard = page.getByRole('heading', { name: 'Certificate' });
+
+    await expect(dnsCard).toContainText('DNS Records');
+    await expect(certCard).toContainText('Certificate');
+  });
+
+  test('Manage DNS Records Link Button', async ({ page }) => {
```
Eakam1007, line 17:
```suggestion
  test('Manage DNS Records Button', async ({ page }) => {
```

test/e2e/landing-page.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    await page.getByRole('link', { name: 'Manage DNS Records' }).click();
+    await expect(page).toHaveURL('/domains');
+  });
+
+  test('Manage DNS Records Instruction Link', async ({ page }) => {
```
Eakam1007, line 22:
```suggestion
  test('DNS Records Instructions Link', async ({ page }) => {
```

test/e2e/landing-page.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+      .filter({
+        hasText: 'DNS Record: Lorem Ipsum is simply dummy text of the printing and typesetting ind',
+      })
+      .getByRole('link', { name: 'to learn more, see these instructions...' })
+      .click();
+
+    await expect(page).toHaveURL('/domains/instructions');
+  });
+
+  test('Manage Certificate Link Button', async ({ page }) => {
```
Eakam1007, line 34:
```suggestion
  test('Manage Certificate Button', async ({ page }) => {
```

test/e2e/landing-page.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    await page.getByRole('link', { name: 'Manage Certificate' }).click();
+    await expect(page).toHaveURL('/certificate');
+  });
+
+  test('Manage Certificate Instruction Link', async ({ page }) => {
```
Eakam1007, line 39:
```suggestion
  test('Certificate Instructions Link', async ({ page }) => {
```

test/e2e/landing-page.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    await page
+      .getByRole('paragraph')
+      .filter({
+        hasText:
+          'Certificate: Lorem Ipsum is simply dummy text of the printing and typesetting ind',
+      })
+      .getByRole('link', { name: 'to learn more, see these instructions...' })
+      .click();
+    await expect(page).toHaveURL('/certificate/instructions');
```
Eakam1007, line 48: Can you add an empty line after line 47 to make it consistent with lines 37-39?

test/e2e/login.spec.ts
```diff
@@ -19,7 +21,7 @@ test('Login with User 1 dev redirect', async ({ page }) => {
   await page.getByLabel('Username').fill('user1');
   await page.getByLabel('Password').fill('user1pass');
   await page.getByRole('button', { name: 'Login' }).click();
-  await page.waitForURL('http://localhost:8080/domains');
+  await page.waitForURL('/domains');
   const locator1 = page.locator('#header-user');
   const userInNav = page.getByRole('heading', { name: 'My.Custom.Domain' });
   const domainHeading = page.getByRole('heading', { name: 'Domains' });
```
Eakam1007, line 27:
```suggestion
  await page.waitForURL('/domains');
  
  const locator1 = page.locator('#header-user');
  const userInNav = page.getByRole('heading', { name: 'My.Custom.Domain' });
  const domainHeading = page.getByRole('heading', { name: 'Domains' });
  
```
Empty lines to make it consistent with lines 9-14

Eakam1007 (changes requested):
A few suggestions, but looks good

test/e2e/certificate.spec.ts
Ririio: oh that's actually a good catch. I was planning on finding the div to check if the request has rendered the loading page. I didn't do an expect to check that fact so it looks weird having it there

test/e2e/landing-page.spec.ts
Ririio: I don't find that necessary. The purpose of that test.before is to check if it renders the landing page's cards before doing any of the test.

Eakam1007: Usually, you use before to do something that is required to test in the `test` blocks. This is kinda breaking away from that. Putting this in a `test` block would do the same thing

Ririio: Oh I guess that's true... Alright that make sense

test/e2e/certificate.spec.ts
```diff
@@ -14,21 +14,41 @@ test.describe('Certificate Page', () => {
...
+    const loadingPageText = page
       .locator('div')
       .filter({
         hasText: 'We have received your request, and will notify you when your certificate is read',
       })
       .nth(1);
 
+    await expect(loadingPageText).toContainText(
+      'We have received your request, and will notify you when your certificate is read'
+    );
```
Eakam1007: Would it be better to add a `class` or `id` to this div, and use that to select the div? Here, we are selecting by text, and then checking the same text, kinda redundant.

OR

Use `await loadingPageText.waitFor()` (https://playwright.dev/docs/api/class-locator#locator-wait-for) instead of line 27

test/e2e/certificate.spec.ts
```diff
@@ -14,21 +14,41 @@ test.describe('Certificate Page', () => {
...
+
+    //Copy Key Toast Text
     await page.getByRole('button', { name: 'Copy Public Key' }).click();
     await page.getByRole('button', { name: 'Copy Private Key' }).click();
 
+    const publicCopyToastText = page.getByText('Public Key was copied to the clipboard');
+    const privateCopyToastText = page.getByText('Private Key was copied to the clipboard');
+
+    await expect(publicCopyToastText).toContainText('Public Key was copied to the clipboard');
+    await expect(privateCopyToastText).toContainText('Private Key was copied to the clipboard');
```
Eakam1007, line 39: Same here, you can use `waitFor()` (https://playwright.dev/docs/api/class-locator#locator-wait-for)

test/e2e/certificate.spec.ts
```diff
@@ -14,21 +14,41 @@ test.describe('Certificate Page', () => {
...
+
+    //Download Key Toast Text
     await page.getByRole('button', { name: 'Download Public Key' }).click();
     await page.getByRole('button', { name: 'Download Private Key' }).click();
 
+    const publicDownloadToastText = page.getByText('Public Key is Downloaded');
+    const privateDownloadToastText = page.getByText('Private Key is Downloaded');
+
+    await expect(publicDownloadToastText).toContainText('Public Key is Downloaded');
+    await expect(privateDownloadToastText).toContainText('Private Key is Downloaded');
```
Eakam1007, line 49: Same here, `waitFor()` could be used instead of checking the text again (https://playwright.dev/docs/api/class-locator#locator-wait-for)

Eakam1007 (changes requested):

test/e2e/certificate.spec.ts
Ririio: I think the `waitFor` seems like a good idea to have considering it returns if the entire text is visible

Eakam1007 (approved):
Looks good!

Myrfion (approved):
LGTM

PR #366, opened 2023-03-17
Giving a choice between SLO and Starchart logout

### LINKED ISSUE #362: Adding SLO prompt
Single logout prompt and option to SLO.

sfrunza13:
Closes #362 

There should be some discussion around this in the coming days, I would like to maybe make this an alert the way we have an alert for delete record but I don't know how to do this across all components.

For now I have made logout.tsx a page with some UI, a couple of buttons to select which type of logout to perform and placed it under the index folder to get the header on it and have it look more a part of the app.

I added back the bits of SLO logic that I had before, this creates a request to the IDP to logout and then receives a response in the query string on the callback.

image


app/routes/__index/logout.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <AbsoluteCenter>
+      <Container>
+        <Stack spacing={6}>
+          <Form method="post">
+            <input type="hidden" name="intent" value="slo" />
+            <Button type="submit">Single Logout from all Seneca Services</Button>
+          </Form>
+          <Form method="post">
+            <input type="hidden" name="intent" value="logoutStarchart" />
+            <Button type="submit">Logout of Starchart and Starchart alone</Button>
```
Ririio, line 46: Question: I noticed this with another PR, but what is the used of an input if it's hidden?

humphd: Think of a `form` element as a blob of data.  If you want to include values in that blob of data that aren't entered by the user, you need a place to put them.  That's the point of the `hidden` type.

humphd (changes requested):
Most people aren't going to care about the distinction you're drawing here, and I want to de-emphasize it.

When I sign-out, it should sign me out of Starchart automatically. You can give a message saying, "You've been signed out of My.Custom.Domain. If you'd also like to sign out of Seneca, click here" or something.

This way, people can safely ignore what you're doing here vs. having 2 big call to action buttons that are going to cause confusion.

app/routes/__index/logout.tsx
Ririio: ohh that make sense thank you

sfrunza13:
> Most people aren't going to care about the distinction you're drawing here, and I want to de-emphasize it.
> 
> When I sign-out, it should sign me out of Starchart automatically. You can give a message saying, "You've been signed out of My.Custom.Domain. If you'd also like to sign out of Seneca, click here" or something.
> 
> This way, people can safely ignore what you're doing here vs. having 2 big call to action buttons that are going to cause confusion.

I have been wondering how to keep the user after destroying the session, how do I know who to SLO?

sfrunza13:
This also means that if this SLO prompt will be with the rest of the index stuff I will have to change the header to useOptionalUser cause the signout will happen before the prompt for SLO. Which is still a problem since I don't know what user to SLO.

humphd:
I see your issue.  Let's give the user a choice:

1. Sign out of My.Custom.Domain
2. Sign out of My.Custom.Doman and all Seneca services

Make 1. the red call-to-action button, and have 2. be gray/lighter, so you don't naturally use it.

sfrunza13:
> I see your issue. Let's give the user a choice:
> 
> 1. Sign out of My.Custom.Domain
> 2. Sign out of My.Custom.Doman and all Seneca services
> 
> Make 1. the red call-to-action button, and have 2. be gray/lighter, so you don't naturally use it.

I have an idea, it might be awful I can't really tell but I am going to commit it so you can see.

humphd:
> I have an idea, it might be awful I can't really tell but I am going to commit it so you can see.

Test it on some Seneca friends, see if they can make sense of it.  That's who we are building this for.

sfrunza13:
Is this an ok use of global var? do I need to use global here?

app/components/header.tsx
```diff
@@ -93,7 +92,7 @@ export default function Header() {
 
       <Flex justifyContent="flex-end" alignItems="center" color="white" gap="5" width="100%">
         <Hide below="lg">
-          <Text id="header-user">{user.username}</Text>
+          <Text id="header-user">{user ? user.username : ''}</Text>
```
humphd, line 95:
```ts
{ user && <Text id="header-user">{user.username}</Text> }
```

app/components/header.tsx
```diff
@@ -16,12 +16,11 @@ import {
 import { TriangleUpIcon, LockIcon, HamburgerIcon } from '@chakra-ui/icons';
 import { Link, useFetcher } from '@remix-run/react';
 
-import { useUser } from '~/utils';
+import { useOptionalUser } from '~/utils';
 
 export default function Header() {
-  const user = useUser();
+  const user = useOptionalUser();
```
humphd, line 22: @Ririio does this change cause any issues elsewhere in our code?

app/routes/__index/logout.tsx
```diff
@@ -0,0 +1,64 @@
+import { redirect } from '@remix-run/node';
+import { Form, useLoaderData } from '@remix-run/react';
+import { AbsoluteCenter, Button, Container, Stack, Text } from '@chakra-ui/react';
+import { createLogoutRequest } from '~/lib/saml.server';
+import { getUsername, logout } from '~/session.server';
+
+import type { ActionArgs, LoaderArgs } from '@remix-run/node';
+
+declare global {
+  var __tempUser__: string | undefined;
```
humphd, line 10: This isn't going to work.  If you have 2 users doing this at the same time, only 1 will be able to store their state here.  You need to store this info in the browser or cookie or in some DOM state or something.

app/routes/__index/logout.tsx
```diff
@@ -0,0 +1,64 @@
...
+};
+
+export default function Index() {
+  const user = useLoaderData<string>();
+
+  return (
+    <AbsoluteCenter>
+      <Container>
+        <Stack spacing={6} align="left">
+          <Text fontSize={'40px'}>{user ? user : ''}</Text>
```
humphd, line 50: Why would we ever end up here without a `user`?

humphd (changes requested):

humphd (commented):
I'd love to go through this with you on a call this week to understand some of your choices.  I'm not sure I understand all of this.

sfrunza13:
> 

I would be happy to impart what I understand about my own choices : ' )

app/components/header.tsx
Ririio: I don't think it will, the purpose was merely to display the username, it might affect the testing, because we require a rendering of the username in that specific area, but I don't think it will cause any problems

app/components/header.tsx
Ririio: I am a bit confused as to why we make the user optional? What is the purpose of this if you don't mind me asking?

sfrunza13: The session and the user info the header relies on will be deleted when the SLO logout prompt is displayed, but the header is still rendered.

app/routes/__index/logout.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <Container>
+        <Stack spacing={6} align="left">
+          <Text fontSize={'40px'}>{user}</Text>
+          <Text fontSize={'30px'}>You have been logged out of My.Custom.Domain.</Text>
+          <Text fontSize={'25px'}>
+            If you would also like to be logged out of all other Seneca services click the "Seneca
+            Logout" Button
+          </Text>
+          <Form method="post">
+            <input type="hidden" name="tempUser" value={user} />
```
SerpentBytes, line 57: Should we use the Chakra UI's `Input` type here? Or does it not matter since the input field would alway remain hidden?

SerpentBytes (dismissed):
Nice. Works on my machine.

app/components/header.tsx
```diff
@@ -93,9 +92,7 @@ export default function Header() {
       </Link>
 
       <Flex justifyContent="flex-end" alignItems="center" color="white" gap="5" width="100%">
-        <Hide below="lg">
-          <Text id="header-user">{user.username}</Text>
-        </Hide>
+        <Hide below="lg">{user && <Text id="header-user">{user.username}</Text>}</Hide>
```
humphd, line 95: Could this be `{user?.username}` instead, so you don't have to do the `&&`?

app/routes/logout.tsx
```diff
@@ -1,18 +1,84 @@
...
+import type { ActionArgs, LoaderArgs } from '@remix-run/node';
 
 export const action = async ({ request }: ActionArgs) => {
   const user = await getUsername(request);
+  const formData = await request.formData();
 
-  // Invalidate the Starchart session but do not log out from Seneca IDP.
-  if (user) {
-    return logout(request);
+  const value = formData.get('sloUsername');
```
humphd, line 22: Instead of `value` can you name this what it is, `sloUsername`?

app/routes/logout.tsx
```diff
@@ -1,18 +1,84 @@
...
+  if (user !== undefined) {
+    // Invalidate the Starchart session but do not log out from Seneca IDP.
+    return logout(request, '/logout');
+  }
+
+  if (value) {
```
humphd, line 28: `if(typeof sloUsername === 'string')` and you can skip the `as string` below.

app/routes/logout.tsx
```diff
@@ -1,18 +1,84 @@
...
-export const loader = async () => {
-  return redirect('/');
+export const loader = async ({ request }: LoaderArgs) => {
+  const cookies = request.headers.get('Cookie');
+
+  const sloUsername = await sloUsernameCookie.parse(cookies);
+  if (!sloUsername) {
+    return redirect('/');
+  } else {
+    // remove the cookie
```
humphd, line 44: remove the SLO cookie (be clear which one, since we have session cookies too)

app/session.server.ts
```diff
@@ -93,11 +97,26 @@ export async function createUserSession({
...
+  headers.append('Set-Cookie', await sessionStorage.destroySession(session));
+
+  return redirect(redirectTo ? redirectTo : '/', {
+    headers,
   });
 }
+
+export const sloUsernameCookie = createCookie('sloUsername', {
+  domain: process.env.NODE_ENV === 'production' ? new URL(process.env.APP_URL).hostname : undefined,
+  // domain: 'localhost',
```
humphd, line 114: Remove this please

humphd (changes requested):

app/routes/logout.tsx
```diff
@@ -1,18 +1,84 @@
...
+} from '@chakra-ui/react';
+import { createLogoutRequest } from '~/lib/saml.server';
+import { getUsername, sloUsernameCookie, logout } from '~/session.server';
 
-import type { ActionArgs } from '@remix-run/node';
+import type { ActionArgs, LoaderArgs } from '@remix-run/node';
 
 export const action = async ({ request }: ActionArgs) => {
   const user = await getUsername(request);
+  const formData = await request.formData();
```
humphd, line 20: Move this and line 22 down to 27, where we use it:

```ts
const formData = await request.formData();
const sloUsername = formData.get('sloUsername');
if (typeof sloUsername === 'string') {
  ...
}
```

app/routes/logout.tsx
```diff
@@ -1,18 +1,84 @@
...
 
-  // Invalidate the Starchart session but do not log out from Seneca IDP.
-  if (user) {
-    return logout(request);
+  const sloUsername = formData.get('sloUsername');
+  if (user !== undefined) {
```
humphd, line 23: Is `undefined` important here (vs `null`)?  If you just mean "not set," it's rare to do this and I'd use `if (!user) {...}`.

app/routes/logout.tsx
```diff
@@ -1,18 +1,84 @@
...
 
-export const loader = async () => {
-  return redirect('/');
+export const loader = async ({ request }: LoaderArgs) => {
+  const cookies = request.headers.get('Cookie');
+
+  const sloUsername = await sloUsernameCookie.parse(cookies);
+  if (!sloUsername) {
+    return redirect('/');
+  } else {
```
humphd, line 43: No else after a return.

humphd (changes requested):

app/routes/logout.tsx
sfrunza13: I keep on doing this 😓

humphd (dismissed):

humphd (changes requested):

humphd (approved):

PR #370, opened 2023-03-17
Improve DNS logging on failures

humphd:
I need some better logging when DNS fails for some reason, to debug on staging.

### DIFF
--- a/app/lib/dns.server.ts
+++ b/app/lib/dns.server.ts
@@ -109,7 +109,7 @@ export async function createHostedZone(domain: string) {
     }
     return response.HostedZone.Id.replace('/hostedzone/', '');
   } catch (error) {
-    logger.warn(error);
+    logger.warn('DNS Error in createHostedZone', { domain, error });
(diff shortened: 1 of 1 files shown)


app/lib/dns.server.ts
```diff
@@ -190,11 +190,22 @@ export const deleteRecord = async (
 ) => {
   try {
     if (!isNameValid(name, username)) {
-      throw new Error('Invalid name provided');
+      logger.error('DNS Error in deleteRecord - invalid record name provided', {
+        name,
+        username,
+        baseDomain: buildUserBaseDomain(username),
+      });
+      throw new Error('DNS Error in upsertRecord - invalid name provided');
```
Eakam1007, line 198: Shouldn't this be `deleteRecord` instead of `upsertRecord`?

humphd: Yes, fixing...

Eakam1007 (approved):

SerpentBytes (approved):

PR #369, opened 2023-03-17
Add integration for editing dns records

Myrfion:
## What it changes
- creates a new page by editing the DNS record (I changed the file name to match the new naming)
- updates the form component for DNS records so it supports editing 
- makes server-side integration with DNS records workflow and its API 

closes #331

PR #375, opened 2023-03-17
Include all . characters in domain name

### LINKED ISSUE #359: Missing `.`s in domain display
In the Domains table, we are showing a user's domains, but they are missing all the `.`s except for the first one:

Screenshot 2023-03-16 at 4 34 31 PM

I have confirmed that this is not the case in the DB or Route53, it only happens in the UI.  I think it's probably due to how we're doing the UI to show the subdomain with a different colour.

cc @Myrfion

humphd:
Fixes #359.

Screenshot 2023-03-17 at 3 36 19 PM

### DIFF
--- a/app/components/dns-record/dns-record-name.tsx
+++ b/app/components/dns-record/dns-record-name.tsx
@@ -1,24 +1,21 @@
 import { Flex, Text } from '@chakra-ui/react';
 
 interface DnsRecordNameProps {
-  name: string;
+  subdomain: string;
+  baseDomain: string;
 }
 
-const DnsRecordName = ({ name }: DnsRecordNameProps) => {
-  const [nameBase, ...restOfName] = name.split('.');
-
-  return (
-    <Flex alignItems="flex-end" flexDirection="row">
-      <Text>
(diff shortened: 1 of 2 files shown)


Eakam1007 (dismissed):

humphd:
The approach to take with this will be changed by #377.  I might wait.

Ririio (approved):

humphd:
I've re-done this using the new `subdomain` field in the db, with a slightly different approach.

Myrfion:
Sorry, I just realized that I did the same thing as your PR does in #369 😢 I just didn't use `useUser`, but did it by doing something else in loader (your approach is way better, I just didn't think about this one. What would be a good strategy to resolve the issue that appeared?

humphd:
@Myrfion why don't you take what I did in this PR, and do it in yours.  Then you can close this PR when it's done.

Myrfion:
> @Myrfion why don't you take what I did in this PR, and do it in yours. Then you can close this PR when it's done.

Okay, updated my PR, closing this since #369 includes this PR's functionality

PR #376, opened 2023-03-17
Rename HOSTURL to APP_URL

Eakam1007:
`dotenv` does not override environment variables if they are already set. On Windows 11 (and possibly other versions of Windows), `HOSTNAME` is set to the device's name, which causes the app to crash when running `npm run dev`:

crash message

In the above image, you can see the input is `Eakam-Alien` which is the PC's Device name. 

The solution is to use a different name other than `HOSTNAME`


humphd (approved):

sfrunza13 (approved):
Interesting, thanks for catching

PR #380, opened 2023-03-18
refactor: record 'name' to 'subdomain'

### LINKED ISSUE #377: Should rename the `record` table `name` column to `subdomain`
We had some issues regarding the vague `name` attribute in some cases. To make the db and subsequent code structures more specific, we should use the `subdomain` column name

SerpentBytes:
## Description
Closes #377 

### Steps to Test
1. `npm run setup`
2. Run the project locally

### DIFF
--- a/app/components/domains-table.tsx
+++ b/app/components/domains-table.tsx
@@ -143,14 +143,14 @@ export default function DomainsTable(props: DomainsTableProps) {
                         <Td>{renderDomainStatus(domain.status)}</Td>
                         <Td>
                           <Flex justifyContent="space-between" alignItems="center">
-                            <DnsRecordName name={domain.name} />
+                            <DnsRecordName name={domain.subdomain} />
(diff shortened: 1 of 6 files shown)


humphd (approved):
This is a great fix, and simplifies/clarifies a bunch of the API for DNS stuff.

@cychu42, @dadolhay we'll have to figure out how I handle the DB update on staging, since I don't think `prisma db push` will work here with existing data in the table?

dadolhay (approved):

dadolhay:
This change is awesome, thank you for implementing it!

@humphd It would probably just delete that column and recreate a new one instead.

Locally, we can nuke the complete db
For staging and prod, we can also do the same, or we can write a manual migration doing:
```sql
alter table `record` rename column `name` to `subdomain`;
```

humphd:
OK, staging is updated with `subdomain` in db, no problems!

PR #383, opened 2023-03-19
Disable actions for inactive records

### LINKED ISSUE #361: Disable UI buttons for DNS Record in pending state
When we are creating a DNS Record, it shows immediately in the UI table.  However, it's not appropriate to allow the user to click Delete, Renew, Edit, etc. until the process of creation has completed (e.g., `status=active` or `status=error`).

We should disable these buttons until the `pending` status has been cleared. 

cc @Myrfion

Myrfion:
- disables action buttons for the record which is not active
- throw error if the action is performed on an inactive record 
- redirect back to the records list if the user tries to access the edit page on the inactive record

image

closes #361

### DIFF
--- a/app/routes/__index/domains/index.tsx
+++ b/app/routes/__index/domains/index.tsx
@@ -51,6 +51,12 @@ export const action = async ({ request }: ActionArgs) => {
     });
   }
 
+  if (record.status !== 'active') {
+    throw new Response('Record is not active, action forbiden', {
+      status: 403,
+    });
+  }
+
   switch (intent) {
     case 'renew-record':
       await renewDnsRecordById(record.id);
--- a/app/components/domains-table.tsx
+++ b/app/components/domains-table.tsx
@@ -132,6 +132,7 @@ export default function DnsRecordsTable(props: DnsRecordsTableProps) {
                 const isLoading =
                   transition.state === 'submitting' &&
                   Number(transition.submission.formData.get('id')) === dnsRecord.id;
+                const isRecordActive = dnsRecord.status === 'active';
 
                 return (
                   <Tr key={dnsRecord.id}>
@@ -177,6 +178,7 @@ export default function DnsRecordsTable(props: DnsRecordsTableProps) {
                                   aria-label="Refresh domain"
                                   variant="ghost"
                                   type="submit"
+                                  isDisabled={!isRecordActive}
                                 />
                               </Tooltip>
                             </Form>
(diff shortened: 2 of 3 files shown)


app/routes/__index/domains/index.tsx
```diff
@@ -51,6 +51,12 @@ export const action = async ({ request }: ActionArgs) => {
     });
   }
 
+  if (record.status !== 'active') {
+    throw new Response('Record is not active, action forbiden', {
+      status: 403,
+    });
+  }
```
Ririio, line 58: I was wondering... if we are going to redirect the user back into domain when they are trying to navigate the page, where does the `throw` happense?

humphd (approved):

app/routes/__index/domains/index.tsx
Myrfion: We are redirecting users back only from the domain edit page, while if they try to do something with the inactive record on domains list page we throw an error

app/routes/__index/domains/index.tsx
```diff
@@ -51,6 +51,12 @@ export const action = async ({ request }: ActionArgs) => {
     });
   }
 
+  if (record.status !== 'active') {
+    throw new Response('Record is not active, action forbiden', {
```
SerpentBytes, line 55: spelling error in "forbiden"

humphd:
@Myrfion I use https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker to catch all my spelling mistakes in the editor.  It's amazing.  It makes it look like I can spell (I can't!).

SerpentBytes (approved):

app/routes/__index/domains/index.tsx
```diff
@@ -51,6 +51,12 @@ export const action = async ({ request }: ActionArgs) => {
     });
   }
 
+  if (record.status !== 'active') {
+    throw new Response('Record is not active, action forbidden', {
+      status: 403,
```
humphd, line 56: I wonder about using a `403` here.  A `403` implies authorization failure vs. an invalid resource state.  I think this is a `409`: https://http.dev/409

humphd (changes requested):

app/routes/__index/domains/index.tsx
Myrfion: I used 403 because the chatGPT suggested so 😭 

image

humphd (approved):

Ririio (approved):

PR #384, opened 2023-03-19
Limit the number of records per user

### LINKED ISSUE #110: Enforce a configurable limit on the number of DNS records per user

Genne23v:
Close #110 

- This PR will limit the user to have more than the configurable number of records. 
- This configurable number is set as MAX_USER_RECORD_LIMIT in `.env`
- If MAX_USER_RECORD_LIMIT is not set, 5 will be stored in MAX_USER_RECORD_LIMIT

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -36,3 +36,6 @@ SAML_IDP_METADATA_PATH=config/idp-metadata-dev.xml
 EXPIRATION_REPEAT_FREQUENCY_MS=86400000
 # 7 * 24 * 60 * 60 * 1000 = 604800000 (7 days)
 JOB_REMOVAL_FREQUENCY_MS=604800000
+
+# Limit the maximum number of records per user
+USER_RECORD_LIMIT=20
--- a/app/models/record.server.ts
+++ b/app/models/record.server.ts
@@ -7,6 +7,12 @@ import type { Record } from '@prisma/client';
 export const createUserRecord = async (
   data: Pick<Record, 'username' | 'type' | 'subdomain' | 'value'>
 ) => {
+  if (process.env.USER_RECORD_LIMIT) {
+    if ((await getUserRecordCount(data.username)) >= Number(process.env.USER_RECORD_LIMIT)) {
+      throw new Error('User has reached the maximum number of records');
+    }
+  }
+
   if (await doesRecordExist(data)) {
     throw new Error('Record already exists');
   }


.env.example
```diff
@@ -31,8 +31,14 @@ APP_URL=http://host.docker.internal:8080
...
+SAML_ENTITY_ID=http://host.docker.internal:8080/sp
+
 # Background jobs
 # 24 * 60 * 60 * 1000 = 604800000 (24 hours)
 EXPIRATION_REPEAT_FREQUENCY_MS=86400000
 # 7 * 24 * 60 * 60 * 1000 = 604800000 (7 days)
 JOB_REMOVAL_FREQUENCY_MS=604800000
+
+# Limit the maximum number of record per user
+MAX_USER_RECORD_LIMIT=5
```
humphd, line 44: `USER_RECORD_LIMIT` I think.  Let's use 20 as our default.

app/models/record.server.ts
```diff
@@ -7,6 +7,13 @@ import type { Record } from '@prisma/client';
 export const createUserRecord = async (
   data: Pick<Record, 'username' | 'type' | 'subdomain' | 'value'>
 ) => {
+  if (!process.env.MAX_USER_RECORD_LIMIT) {
```
humphd, line 10: Let's not hard-code a default limit here.  If the env is set, then we do 13-15, otherwise we skip it (i.e., we only enforce a limit if requested).

.env.example
```diff
@@ -31,8 +31,14 @@ APP_URL=http://host.docker.internal:8080
 # The SimpleSAML IDP's XML metadata
 SAML_IDP_METADATA_PATH=config/idp-metadata-dev.xml
 
+# Our apps's SP Entity ID, which is also the URL to our metadata.
+SAML_ENTITY_ID=http://host.docker.internal:8080/sp
```
humphd, line 35: Remove 34, 35 please.

humphd (changes requested):
Nice simple fix!

.env.example
Genne23v: Sorry for the mistake

.env.example
Genne23v: Will be updated!

humphd (approved):

SerpentBytes (dismissed):

humphd (approved):

PR #400, opened 2023-03-21
Add webhook/ with webhook JSON config and deploy.sh with docs

### LINKED ISSUE #50: Create deployment webhook and script

### LINKED ISSUE #74: Implement escaping of webhook params
When serving webhooks with user provided data, such as the deploy webhook, we want the received webhook parameters to be properly escaped.

This can be done by providing each argument to the `printf` utility with a '%q' specifier. 
```sh
printf %q "$ARG1"
```
This will print a properly escaped value of `$ARG1` by escaping all special characters such as quotes, slashes, etc in the `$ARG1` variable.

The webhook image we are running is a based on Alpine and does not contain most GNU tools, therefore the `printf` utility in the image does not support `%q` specifier.

https://www.shellcheck.net/wiki/SC3050

The `webhook/hooks` directory contains scripts for all the webhooks, it also houses utility scripts, prefixed with a `.` such as `.authenticate.sh`, consider creating a `.escape.sh` script to achieve the functionality of `printf %q` to be used in the webhook scripts.

### LINKED ISSUE #76: Include HMAC-256 in webhook x-hub-signature-signature header
We're using a webhook to update deployments on staging and production.  Our GitHub Actions workflows can trigger an action in our servers by `POST`ing some JSON, which includes details about the new version that should be used.

There are a few ways to secure such a request.  One way is to limit the IPs that are allowed to send the message.  This is hard with GitHub, since there are so many CI IPs in use.  Another way is to include a secret token in the headers.  An even better way is to use a secret to create an HMAC signature of the message body (i.e., take the secret + JSON and calculate a digest), then send that in a header.

This method is preferred because the secret is never transferred, and only the sender/receiver can create/verify such signatures.

In our case, doing this involves two things:

1. Update our hooks definition to [include a rule](https://github.com/adnanh/webhook/blob/master/docs/Hook-Rules.md#match-payload-hmac-sha256) for matching an HMAC-SHA256 signature in the `x-signature` header
2. Add a Docker secret in deployment with our the HMAC signing secret.  Must be 32 characters or more in length.
3. Add the same secret to our GitHub Actions Encrypted Secrets
4. Use a GitHub Action to trigger the webhook and send the HMAC signature in the `x-hub-signature` header

This will require us to create a CD workflow as well, which we don't have yet.

### LINKED ISSUE #77: Figure out how to run webhook container rootless
By default, most Docker containers are run using `root`.  It would be nice if we could find a way to allow our webhook container to use the Docker socket (i.e., to communicate with the host and trigger updates via docker-cli calls) while not being the root user.

humphd:
Closes #77
Closes #76
Closes #74
Closes #50

This finishes the work begun in #57 to get our deployment webhook in place.  ITS wants it to run outside of Docker, and has already installed `webhook` on the manager nodes.

This is the first of at least 2 parts to get this completed.  I need to land the major bits here, then test it and adjust and document things in a follow-up.

The main points here are as follows:

1. On staging (and later production), we'll run a [webhook server](https://github.com/adnanh/webhook).  It will listen on :8000 and ITS will route anything that comes to `/hooks/*` to this server.
2. The webhook will listen for `deploy` messages from GitHub, which will include 2 things.  First a new `tag` to use with Docker.  When we update to a new version on `main`, this will send that image tag so it can be deployed; second, we'll send a digest created with a secret.  Our deployment servers will only respond to webhooks that can sign the body (i.e., no one else can hit this endpoint)
3. Assuming the webhook can be verified by the server (calculating the secret + body to see if it matches the digest), the `deploy.sh` script will be run.  The new `tag` will be sent along as an ENV var.  Docker will try to update the starchart image to the new tag.

As we discussed in #291, I'm not doing any database migration/syncing in here.  Any changes to the database schema would require a manual shut-down and update of the DB first.  It would be good to get to an automated version of this some day, but not in this PR.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -314,12 +314,14 @@ jobs:
             # cache node modules using the same key as restore.
           key: ${{ steps.node_modules-cache-restore.outputs.cache-primary-key }}
 
-  Starchart-Docker:
+  Docker-Build-Deploy:
     # We'll only build and push when landing on main, not for PRs
     if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
     runs-on: ubuntu-latest
     # Don't bother unless everything else passes
     needs: [Build, Prettier, Unit-Tests, E2E-Tests, Dockerfile-Lint, ESLint, Type-Check]
+    env:
+      IMAGE: ghcr.io/DevelopingSpace/starchart
     permissions:
       contents: read
       packages: write
@@ -327,18 +329,11 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
 
+      - name: Define GITHUB_SHORT_SHA
+        run: echo "GITHUB_SHA_SHORT=$(echo $GITHUB_SHA | cut -c 1-6)" >> $GITHUB_ENV
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
-
-      - name: Extract Docker metadata from git info
-        id: meta
-        uses: docker/metadata-action@v4
-        with:
-          images: ghcr.io/DevelopingSpace/starchart
-          flavor: latest=true
-          tags: |
-            type=ref,event=branch
-            type=sha
 
       - name: Log in to the GitHub Container Registry
         uses: docker/login-action@v2
@@ -354,5 +349,16 @@ jobs:
           build-args: |
             SAML_IDP_METADATA_PATH=./config/idp-metadata-staging.xml
           push: true
-          tags: ${{ steps.meta.outputs.tags }}
-          labels: ${{ steps.meta.outputs.labels }}
+          tags: |
+            ${{ env.IMAGE }}:sha-${{ env.GITHUB_SHA_SHORT }}, \
+            ${{ env.IMAGE }}:main, \
+            ${{ env.IMAGE }}:latest
+
+      - name: Deploy to Staging via Webhook
+        uses: navied/secure-actions-webhook@0.2.0
+        env:
+          HMAC_SECRET: ${{ secrets.WEBHOOK_STAGING_SECRET }}
+        with:
+          url: https://mycustomdomain-dev.senecacollege.ca/hooks/deploy
+          hmacSecret: ${{ env.HMAC_SECRET }}
+          data: '{"tag": "sha-${{ env.GITHUB_SHA_SHORT }}"}'
--- a/webhook/deploy.sh
+++ b/webhook/deploy.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+# https://github.com/DevelopingSpace/starchart/pkgs/container/starchart
+IMAGE=ghcr.io/developingspace/starchart
+SERVICE=starchart_mycustomdomain
+
+# We expect to receive the new image tag via the TAG environment variable
+if [[ -z "$TAG" ]]; then
+    logger -s "webhook received, no tag set, skipping."
+    exit 1
+fi
+
+logger -s "webhook received, starting update of ${SERVICE} to ${IMAGE}:${TAG}"
+
+if docker service update --image "$IMAGE":"$TAG" "$SERVICE"; then
+    logger -s "webhook update of ${SERVICE} to ${IMAGE}:${TAG} succeeded."
+    exit 0
+else
+    logger -s "webhook update of ${SERVICE} to ${IMAGE}:${TAG} failed."
+    exit 1
+fi
--- a/webhook/hooks.json
+++ b/webhook/hooks.json
@@ -1,0 +1,24 @@
+[
+  {
+    "id": "deploy",
+    "execute-command": "/home/customdomain/deploy.sh",
+    "command-working-directory": "/home/customdomain",
+    "pass-environment-to-command": [
+      {
+        "source": "payload",
+        "name": "tag",
+        "envname": "TAG"
+      }
+    ],
+    "trigger-rule": {
+      "match": {
+        "type": "payload-hmac-sha256",
+        "secret": "TODO: update me to match secret stored in GitHub Actions...",
+        "parameter": {
+          "source": "header",
+          "name": "X-Hub-Signature"
+        }
+      }
+    }
+  }
+]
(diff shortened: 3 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -354,5 +349,16 @@ jobs:
...
+            ${{ env.IMAGE }}:latest
+
+      - name: Deploy to Staging via Webhook
+        uses: navied/secure-actions-webhook@0.2.0
+        env:
+          HMAC_SECRET: ${{ secrets.WEBHOOK_STAGING_SECRET }}
+        with:
+          url: https://mycustomdomain-dev.senecacollege.ca/hooks/deploy
+          hmacSecret: ${{ env.HMAC_SECRET }}
+          data: '{"tag": "${{ env.IMAGE }}:sha-${{ env.GITHUB_SHA_SHORT }}"}'
```
humphd, line 364: Don't send image here, only the new tag.

.github/workflows/ci.yaml
```diff
@@ -354,5 +349,16 @@ jobs:
...
           push: true
-          tags: ${{ steps.meta.outputs.tags }}
-          labels: ${{ steps.meta.outputs.labels }}
+          tags: |
+            ${{ env.IMAGE }}:sha-${{ env.GITHUB_SHA_SHORT }}, \
+            ${{ env.IMAGE }}:main, \
+            ${{ env.IMAGE }}:latest
+
+      - name: Deploy to Staging via Webhook
+        uses: navied/secure-actions-webhook@0.2.0
```
sfrunza13, line 358: If I am correct this was written by senior staff engineer @ shopify, last updated a few years ago, one star on the repo. How did you find this? 👀

webhook/deploy.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+# https://github.com/DevelopingSpace/starchart/pkgs/container/starchart
+IMAGE=ghcr.io/developingspace/starchart
+SERVICE=starchart_mycustomdomain
+
+# We expect to receive the new image tag via the TAG environment variable
+if [[ -z "$TAG" ]]; then
```
sfrunza13, line 9: Idk how to read this, what is -z?

.github/workflows/ci.yaml
```diff
@@ -314,32 +314,27 @@ jobs:
...
+      IMAGE: ghcr.io/DevelopingSpace/starchart
     permissions:
       contents: read
       packages: write
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
 
+      - name: Define GITHUB_SHORT_SHA
+        run: echo "GITHUB_SHA_SHORT=$(echo $GITHUB_SHA | cut -c 1-6)" >> $GITHUB_ENV
```
sfrunza13, line 333: I don't think this really makes a difference but aren't the characters on the tail end of a sha the most meaningful?

sfrunza13 (approved):
I don't think I am going to be able to find a fault with this. After everything in the needs occurs add IMAGE to env, then build and push an image with new tags, use navied/secure-actions-webhook@0.2.0 to post to hooks safely, then deploy.sh logic which should use the tag passed to it through data as you already mentioned yourself so that will be changed for the better. I think it's good.

webhook/deploy.sh
Eakam1007: Seems like this is to check it $TAG is null

webhook/hooks.json
```diff
@@ -0,0 +1,24 @@
...
+      {
+        "source": "payload",
+        "name": "tag",
+        "envname": "TAG"
+      }
+    ],
+    "trigger-rule": {
+      "match": {
+        "type": "payload-hmac-sha256",
+        "secret": "TODO: update me to match secret stored in GitHub Actions...",
```
Eakam1007, line 16: Wanted to confirm, Is this secret supposed to be a path to a secret file or something like that? Or is this going to be manually updated later (and not comnitted to git)? Doesn't seems like a secret if we commit this here

.github/workflows/ci.yaml
Eakam1007: Apart from this, and one small question looks good

Eakam1007 (changes requested):

webhook/hooks.json
sfrunza13: good point

.github/workflows/ci.yaml
humphd: https://github.com/marketplace?type=actions. I needed an HMAC SHA256 webhook action, and there are precious few for some reason.

.github/workflows/ci.yaml
humphd: You can take any hash and truncate it to use only the first part of it as a short-hand.  This is what GitHub does when they show my commit as `93a5088` vs. `93a508896294bdd9f74b1d4ca92a57c49999f34e`

sfrunza13 (approved):

webhook/hooks.json
Eakam1007: So, will you ssh into the server to change this manually?

Eakam1007 (dismissed):
Just a small question out of curiosity, but looks good

humphd:
> Just a small question out of curiosity, but looks good

Correct, these files won't be used as is. They are a template for manually setting this up on staging.

humphd:
I've updated this again.  I've added the `systemd` startup unit file, so the webhook runs as a service (it's already installed on staging).  I think this is ready to test!

sfrunza13 (approved):

Eakam1007 (approved):

humphd:
Thanks for the review!  I'm going to try this now...

humphd:
Crap, I got my order wrong, I have to do another PR to make this happen.

PR #402, opened 2023-03-21
Don't fail on delete with missing record; update record status early

### LINKED ISSUE #360: How should a Route53 Delete deal with missing domain?
In testing (and theoretically in production), we can get into a state where the MySQL database has a DNS Record, but it doesn't exist in Route53 for some reason (manually removed, error during creation, etc).

When the DNS Flow happens, we currently fail and retry the delete a few times, then fail the whole job.  However, if we are trying to delete a DNS Record, and it isn't there, the job could actually pass.

We need to decide on how to deal with this case.  I'm wanting to not have orphaned records in the DB that we can't delete via automation because they are only partially created.

cc @Genne23v, @dadolhay, @SerpentBytes

### LINKED ISSUE #390: Delete record not working
In addition to #360, we need to fix how the record delete UX works.

I've updated staging to the latest code from 0.6.  When I try to delete a record, I see this in the server logs:

```
starchart_mycustomdomain.1.s128rohy5ltb@cudm-mgmt01dv.dcm.senecacollege.ca    | {"error":{"$fault":"client","$metadata":{"attempts":1,"httpStatusCode":400,"requestId":"8981f7e1-9b16-4e65-ad94-49672c3c60fe","totalRetryDelay":0},"Code":"InvalidChangeBatch","Type":"Sender","name":"InvalidChangeBatch"},"level":"warn","message":"DNS Error in deleteRecord","name":"test2.davidhumphrey._stage_.mystudentproject.ca","timestamp":"2023-03-20T12:56:37.140Z","type":"CNAME","username":"david.humphrey","value":"humphd.github.io"}

...repeats 5 times
```

There are a few issues here.

1. We should immediately change the record's status to `pending` so the UI can poll for changes, and remove the record when deleted, or show the `error` status.
2. We should make sure that we're sending the correct `name` to Route53.  It looks like we're using the `fqdn` vs. `subdomain` above, and I wonder if that's the reason for the `400`?
3. If we get back a `4xx` which isn't due to rate limiting, there is no point doing 4 more retries: the client request is in error, and needs to be altered to work. 
4. If it fails, we should go into the `error` status in the db/UI, but that never happens.

Some of the above issues will also be present and need to be improved in create and update as well.

humphd:
Fixes #360
Fixes #390

This makes 2 changes to how the DNS logic works:

1. When we try to delete a record from Route53, we first check if the record is there.  If it is not, we are already done, and don't fail.  Doing this requires some changes to other parts of the code that expect a Change ID, which is now going to be `null`.  We simply skip doing the `insync` checks with Route53 and report back that the change is ready.
2. For all DNS flows, the database is updated first, so that the UI can poll for changes.

To test this:

1. Create a new domain, make sure the pending/active UI works
2. Update the domain, make sure the pending/active UI works
3. Delete the domain, make sure the pending/active UI works

### DIFF
--- a/app/lib/dns.server.ts
+++ b/app/lib/dns.server.ts
@@ -2,6 +2,7 @@ import {
   Route53Client,
   CreateHostedZoneCommand,
   ChangeResourceRecordSetsCommand,
+  ListResourceRecordSetsCommand,
   GetChangeCommand,
 } from '@aws-sdk/client-route-53';
 import isFQDN from 'validator/lib/isFQDN';
@@ -15,6 +16,7 @@ import type {
   CreateHostedZoneResponse,
   ChangeResourceRecordSetsResponse,
   GetChangeResponse,
+  ListResourceRecordSetsResponse,
 } from '@aws-sdk/client-route-53';
 import type { DnsRecordType } from '@prisma/client';
 
@@ -45,6 +47,10 @@ if (NODE_ENV === 'production') {
   // In dev, we only need a root domain, and can fake the rest
   process.env.ROOT_DOMAIN = process.env.ROOT_DOMAIN || 'starchart.com';
 }
+
+// Ensure a trailing `.` on a domain to make it fully qualified:
+// a.b.c -> a.b.c. | a.b.c. -> a.b.c.
+const toFQDN = (domain: string) => domain.replace(/\.?$/, '.');
 
 /**
  * In production, we have to have a zone id to do anything, but in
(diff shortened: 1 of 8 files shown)


humphd:
Rebased onto `main` with all the `DNS*` changes.  This is working locally for me.

app/queues/dns/update-dns-record-flow.server.ts
```diff
@@ -34,6 +36,8 @@ export const updateDnsRequest = async (data: UpdateDnsRequestData) => {
     course: data.course,
     ports: data.ports,
   });
+  // And set its status to pending
+  await updateDnsRecordStatus(id, DnsRecordStatus.pending);
```
Genne23v, line 40: I think we should also update the DB after Route53 gets updated. I will write a PR for this one.

Genne23v (changes requested):
I was working on this issue as well. But this PR has a better solution. Thanks for this PR!

app/lib/dns.server.ts
```diff
@@ -128,6 +134,36 @@ export const createDnsRecord = (
   }
 };
 
+async function checkRecordExists(type: DnsRecordType, name: string, value: string) {
```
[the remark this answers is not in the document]
humphd: Good catch, fixing in new commit.

app/models/dns-record.server.ts
```diff
@@ -65,6 +65,13 @@ export function updateDnsRecordById(
   });
 }
 
+export function updateDnsRecordStatus(id: DnsRecord['id'], status: DnsRecordStatus) {
```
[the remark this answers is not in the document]
humphd: Good idea, I made some improvements.

humphd:
I've made some updates, can you guys re-review?

app/lib/dns.server.ts
```diff
@@ -128,6 +134,36 @@ export const createDnsRecord = (
...
+    // Grab the one and only one and compare values
+    const { Name, Type, ResourceRecords } = ResourceRecordSets[0];
+    return (
+      Name === toFQDN(name) &&
+      Type === type &&
+      ResourceRecords?.length === 1 &&
+      ResourceRecords[0].Value === value
+    );
+  } catch (error) {
+    logger.warn('DNS Error in checkRecordExists', { name, type, value, error });
```
SerpentBytes, line 162: nit: `checkDnsRecordExists`

app/lib/dns.server.ts
```diff
@@ -213,6 +249,12 @@ export const deleteDnsRecord = async (
       throw new Error('DNS Error in deleteDnsRecord - invalid value provided');
     }
 
+    // If no such record exists in Route53, we're done ("delete" state already achieved)
+    if (!(await checkDnsRecordExists(type, name, value))) {
+      logger.debug('DNS deleteRecord - no such record in Route53, skipping', { type, name, value });
```
SerpentBytes, line 254: typo: `deleteDnsRecord`

SerpentBytes: Shouldn't this be `checkDnsRecordExists`?

PR #410, opened 2023-03-22
Update vitest config

### LINKED ISSUE #134: Confirm our Vitest config and unit test setup
We ran into some bugs today with version issues that triggered odd failures in our Vitest unit test run.  I think we should look at how this is configured and tweak it if necessary.  I've not used Vitest before, so I don't have a lot of advice.  Whoever does this should go through their docs and understand how ours is setup, and make changes if necessary.

Especially since we're going to be using Playwright for most UI tests, we can use this for unit testing in node mostly I think.

Eakam1007:
Resolves #134 

Make a few changes to our vitest config:
- Added a `test:coverage` script to run tests with coverage. This shows the coverage report in the terminal and generated a html report in `coverage` folder
- Removed c8 and replaced it with istanbul which will be used to generate the coverage. This was mainly done because there were issues getting the default `c8` to properly generate coverage as no files were displayed in the coverage. Even with the `all` files setting, none of the files showed anything being covered by tests
- Added `coverage` folder to `.gitignore`
- Updated `watchExclude` to exclude our `mysql-data` folder
- Removed redundant `include` in vitest.config

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ node_modules
 /playwright-report/
 /playwright/.cache/
 /test/e2e/.auth
+/coverage
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,7 +60,7 @@
         "@types/react-dom": "^18.0.11",
         "@types/validator": "^13.7.12",
         "@vitejs/plugin-react": "^3.0.1",
-        "@vitest/coverage-c8": "^0.28.5",
+        "@vitest/coverage-istanbul": "^0.29.7",
         "c8": "^7.13.0",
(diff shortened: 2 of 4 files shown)


cychu42:
As discussed in the meeting, I think it would be beneficial to see what files are not covered in `app/models` by unit tests.

sfrunza13 (approved):
Let's get this in :)

Eakam1007:
> As discussed in the meeting, I think it would be beneficial to see what files are not covered in `app/models` by unit tests.

I have filed #420 to add coverage for playwright tests, and I am planning to update this coverage based on that

PR #412, opened 2023-03-22
Notify user when DNS record status change from `pending` to `active`/`error`

### LINKED ISSUE #387: Notify users when a record switches from pending to active/error
Since it can take a while, we should let users know when operations on their new/updated/deleted records are completed by sending an email notification.

SerpentBytes:
## Description
Resolves #387

### Steps to Test
1. Run the project
2. Navigate to the `/dev` route and interact with the "Add DNS Record" button
3. Navigate to Mailhog UI at http://localhost:8025 to check if a notification email was sent

### DIFF
--- a/app/queues/dns/dns-queue.server.ts
+++ b/app/queues/dns/dns-queue.server.ts
@@ -3,10 +3,12 @@ import { DnsRecordStatus } from '@prisma/client';
 
 import { redis } from '~/lib/redis.server';
 import logger from '~/lib/logger.server';
-import { updateDnsRecordById } from '~/models/dns-record.server';
+import { updateDnsRecordById, getDnsRecordById } from '~/models/dns-record.server';
 import { dnsWorker } from './dns-worker.server';
+import { addNotification } from '../notifications/notifications.server';
 
 import type { DnsJobData } from './dns-worker.server';
+import type { DnsRecord } from '@prisma/client';
 
 export const dnsQueue = new Queue<DnsJobData>('dns-queue', { connection: redis });
 const dnsQueueEvents = new QueueEvents('dns-queue', { connection: redis });
@@ -18,8 +20,39 @@ declare global {
 function init() {
   logger.debug('DNS Queue and DNS Worker init');
 
+  const sendDnsRecordStatusUpdateNotification = async (
+    id: DnsRecord['id'],
+    status: DnsRecord['status']
+  ) => {
+    const dnsRecord = await getDnsRecordById(id);
+    if (!dnsRecord) {
+      throw new Error(`Could not get DNS Record with id=${id}`);
+    }
+    const { username, subdomain, type, value, user } = dnsRecord;
+    switch (status) {
+      case DnsRecordStatus.error:
+        logger.debug(
+          `adding job to send error email notification for DNS record: username=${username}, subdomain=${subdomain}, type=${type}, value=${value}`
+        );
+        return addNotification({
+          emailAddress: user.email,
+          subject: `DNS Record: ${subdomain} encountered error`,
+          message: `${username}, your DNS record with subdomain ${subdomain} encountered an error.`,
+        });
+      case DnsRecordStatus.active:
+        logger.debug(
+          `adding job to send active email notification for DNS record: username=${username}, subdomain=${subdomain}, type=${type}, value=${value}`
+        );
+        return addNotification({
+          emailAddress: user.email,
+          subject: `DNS Record: ${subdomain} is active`,
+          message: `${username}, your DNS record with subdomain ${subdomain} is ready.`,
+        });
+    }
+  };
+
   /**
-   * When a DNS job fails, clean up the status in the db
+   * When a DNS job fails, clean up the status in the db, and add email notification job
    */
   dnsQueueEvents.on('failed', async ({ jobId }) => {
     const job = await dnsQueue.getJob(jobId);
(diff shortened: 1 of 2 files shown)


app/queues/dns/workers/sync-db-status-worker.server.ts
```diff
@@ -5,21 +5,47 @@ import logger from '~/lib/logger.server';
...
   id: DnsRecord['id'];
+  username: DnsRecord['username'];
 }
-
+const notifyUserBasedOnStatus = async (
+  username: DnsRecord['username'],
+  status: DnsRecordStatus
+) => {
+  // get user information from DB
+  let user = await getUserByUsername(username);
```
humphd, line 26: Because you have to talk to the DB anyway, let's not bother adding the `username` to all the worker code/types above, and instead use the DNS Record `id` to get the user and other info we need to send the message here.

humphd (changes requested):

app/queues/dns/workers/sync-db-status-worker.server.ts
```diff
@@ -5,21 +5,47 @@ import logger from '~/lib/logger.server';
...
+  if (!user) {
+    throw new Error(`Could not get user by username: ${username}`);
+  }
+  switch (status) {
+    case DnsRecordStatus.active:
+      return addNotification({
+        emailAddress: user.email,
+        subject: 'Sample DNS Record active subject',
```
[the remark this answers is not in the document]
SerpentBytes: Since I am getting all record data (including `status`) from the DB to send notifications, do I still need the `switch` statement? I remember it was mentioned that Route53 and DB could be out of sync.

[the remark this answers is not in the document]
SerpentBytes: Actually, let me just test it out.

dadolhay:
Are we sure we want to notify users about success? creating/updating a record should only take a few seconds. I think the success verification is more annoying than useful

SerpentBytes:
> Are we sure we want to notify users about success? creating/updating a record should only take a few seconds. I think the success verification is more annoying than useful

I think it would be helpful for debugging purposes, as there's a bug associated with email notifications not working on staging, according to @humphd. We can remove it later.

humphd:
@dadolhay right, we discussed this and decided to do it on staging for testing, but later we can just do it for errors.

humphd (commented):
My changes in #419 are going to affect this, so I wonder if we should wait to land it?

SerpentBytes:
> My changes in #419 are going to affect this, so I wonder if we should wait to land it?

Go ahead. This is a small PR, so would not be difficult to rebase. 🙂

humphd:
@SerpentBytes I'm hoping to land #419 today, and then we can fix this to work with it.

humphd:
@SerpentBytes OK, #419 is in!

Here is how you should approach this.  We need to send a notification when the DNS job is done.  There are 2 cases: `completed` (job worked) or `failed` (job failed).  We have code for `failed` already here: https://github.com/DevelopingSpace/starchart/blob/main/app/queues/dns/dns-queue.server.ts#L24-L41. You can add similar code for `completed` in the same area.

In both cases, we need to get the `job` based on the `jobId` and then use that to get the `username` + `dnsRecordId` from the `job.data`. We can use those to lookup the info we need in order to enqueue an email notification job to let the user know what happened.

Does that make sense?

SerpentBytes:
> @SerpentBytes OK, #419 is in!
> 
> Here is how you should approach this. We need to send a notification when the DNS job is done. There are 2 cases: `completed` (job worked) or `failed` (job failed). We have code for `failed` already here: https://github.com/DevelopingSpace/starchart/blob/main/app/queues/dns/dns-queue.server.ts#L24-L41. You can add similar code for `completed` in the same area.
> 
> In both cases, we need to get the `job` based on the `jobId` and then use that to get the `username` + `dnsRecordId` from the `job.data`. We can use those to lookup the info we need in order to enqueue an email notification job to let the user know what happened.
> 
> Does that make sense?

Makes sense. 👍 
I will work on updating this PR with the requested changes later today.

SerpentBytes:
@humphd I have updated this PR with the requested changes. Reviews will be appreciated.

app/queues/dns/dns-queue.server.ts
```diff
@@ -18,8 +19,34 @@ declare global {
...
+    id: DnsRecord['id'],
+    status: DnsRecord['status']
+  ) => {
+    const dnsRecord = await getDnsRecordById(id);
+    if (!dnsRecord) {
+      throw new Error(`Could not get DNS Record with id=${id}`);
+    }
+    switch (status) {
+      case DnsRecordStatus.error:
+        logger.debug('adding job to send DNS record error notification');
```
humphd, line 32: I would include info in the debug message from the `dnsRecord` (who owns it?  What is name, value, type?)

humphd (changes requested):
One lint issue, the rest looks good!

SerpentBytes:
@humphd, I have fixed the ESLint issue and added more context to debug log messages. I would appreciate another round of reviews.

app/queues/dns/dns-queue.server.ts
```diff
@@ -30,9 +63,36 @@ function init() {
...
+
+  /**
+   * When a DNS job completes, update the status in the db, and add email notification job
+   */
+  dnsQueueEvents.on('completed', async ({ jobId }) => {
+    const job = await dnsQueue.getJob(jobId);
+    if (job) {
+      const { dnsRecordId } = job.data;
+      try {
+        await updateDnsRecordById(dnsRecordId, {
```
humphd, line 85: I already do this in the worker, see https://github.com/DevelopingSpace/starchart/blob/main/app/queues/dns/dns-worker.server.ts#L128-L147

Let's only send the notification here.

I the `fail` case I have to update the db, because I can't deal with failure in the worker itself due to retries.

humphd (changes requested):

app/queues/dns/dns-queue.server.ts
SerpentBytes: My bad; let me fix this.

SerpentBytes:
@humphd I've updated the code based on your feedback.

humphd (approved):

Eakam1007 (approved):

PR #414, opened 2023-03-22
feat: Finish certificate order

dadolhay:
- Alters system to create an order entry in the flow dispatcher. We need to send the ID to every job, as workers can only get the direct children's retval
- Fixes DNS system integration issue
- Adds additional error handling
- Completes the certificate order and stores the cert in the DB

!! changes `Certificate` db signature !!


app/queues/certificate/challenge-completer-worker.server.ts
```diff
@@ -3,47 +3,35 @@ import { redis } from '~/lib/redis.server';
...
 import * as certificateModel from '~/models/certificate.server';
-import { getChildrenValuesOfQueueName } from '~/utils';
-import { orderCreatorQueueName } from './order-creator-worker.server';
-
-import type { OrderCreatorOutputData } from './order-creator-worker.server';
 
 export interface ChallengeCompleterData {
   rootDomain: string;
   username: string;
+  certificateId: number;
```
humphd, line 10: Use the type from Prisma so it matches.

app/queues/certificate/dns-cleaner-worker.server.ts
```diff
@@ -5,6 +5,7 @@ import logger from '~/lib/logger.server';
 export interface DnsCleanerData {
   rootDomain: string;
   username: string;
+  certificateId: number;
```
humphd, line 8: Same here, use Prisma type.

Question: you are defining the same type over and over again.  Can we abstract it to a separate file and re-use across all these jobs which share the same signature?

app/queues/certificate/order-completer-worker.server.ts
```diff
@@ -1,19 +1,91 @@
...
+      newError.stack = (e as Error).stack;
+      throw newError;
+    }
+
+    /**
+     * Initialize Let's Encrypt
+     */
+    let letsEncrypt;
+    try {
+      logger.debug("Initializing Let's encrypt");
```
humphd, line 49: Why does this file change to using " vs. '?  I'm surprised Prettier isn't mad.

humphd (changes requested):
This is great, and only a few things stuck out to me.

app/queues/certificate/order-completer-worker.server.ts
dadolhay: It's because there is an apostrophe in the string. If that wasn't there, it would switch it to `'` This way there is no need to escape that char

app/queues/certificate/dns-cleaner-worker.server.ts
dadolhay: yes, this was my plan for whan I do the refactor

dadolhay: Any preference, how to name that file?

app/queues/certificate/challenge-completer-worker.server.ts
```diff
@@ -4,10 +4,12 @@ import logger from '~/lib/logger.server';
 import LetsEncrypt from '~/lib/lets-encrypt.server';
 import * as certificateModel from '~/models/certificate.server';
 
+import type { DnsRecord } from '@prisma/client';
+
 export interface ChallengeCompleterData {
   rootDomain: string;
   username: string;
```
humphd, line 11: `DnsRecord['username']`

humphd (changes requested):
All of these are the same type, why not put it in `app/queues/certificate/*` in its own file and name it something like `CertificateJobData` or something and reuse it across all of these files?

prisma/schema.prisma
```diff
@@ -40,7 +40,7 @@ model Certificate {
   id           Int               @id @default(autoincrement())
   username     String
   domain       String
-  orderUrl     String            @unique @db.VarChar(255)
+  orderUrl     String?           @unique @db.VarChar(255)
```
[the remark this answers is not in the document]
sfrunza13: I believe it's because in this flow the certificates are created before the orders with only the username and domain and the orders are then added in `order-creator-worker.server.ts`, there's a few steps there like initializing lets encrypt and then actually creating the order that can throw which I imagine might benefit from the job retrying on failure (not sure)? But in any case this is how I see it working and why I think orderUrl is optional, cause we start without it.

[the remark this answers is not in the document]
sfrunza13: This is also why in `certificate.server.ts` you can see orderUrl moved from create to update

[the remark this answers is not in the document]
dadolhay: @sfrunza13 you are spot on!
Also since parent queues can only access **direct children's return values** I needed a way to pass all jobs the certificate id, and I can only do that if I create it in the flow, where I don't yet have the order URL

sfrunza13 (changes requested):

app/queues/certificate/certificateJobTypes.server.ts
```diff
@@ -0,0 +1,7 @@
+import type { DnsRecord } from '@prisma/client';
+
+export interface CertificateJobData {
```
humphd, line 3: Wonderful improvement!

app/utils.ts
```diff
@@ -74,6 +74,19 @@ export function buildDomain(username: PrismaUser['username'], name?: string) {
   return buildUserBaseDomain(username);
 }
 
+export function getSubdomainFromFqdn(username: PrismaUser['username'], fqdn: string): string {
+  const baseDomain = buildUserBaseDomain(username);
+
+  if (!fqdn.endsWith(`.${buildUserBaseDomain(username)}`)) {
```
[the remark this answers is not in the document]
humphd: He's on the server side, we don't have a `user` object, only their `username`

humphd (approved):
I'm excited to get this up on staging soon so we can test it live!

app/utils.ts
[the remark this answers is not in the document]
sfrunza13: I mean there is baseDomain declared on 78, why not use the same on line 80?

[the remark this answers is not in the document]
dadolhay: @sfrunza13 Nice catch. It seems that I was not diligent enough when refactoring. Updated.

humphd:
@dadolhay can you fix the merge conflict you have here?  I landed the DNS refactor, which altered the import you need to use.

dadolhay:
@humphd Sure, updated!

sfrunza13 (approved):

humphd (approved):
Ship it! 🚢

humphd:
I'm not sure which conflicts this has?  The web UI is claiming it can't get merged yet...

dadolhay:
It merged for me without issue

PR #415, opened 2023-03-22
Adding tests for functions relating to User model

cychu42:
This is part of issue #238 

## Changes:
In `test\unit\user.server.test.ts`, add tests for:
* isStudent()
* isAdmin()
* isFaculty()
* checkUsernameExists()
* getUserByUsername()
* createUser()

### DIFF
--- a/test/unit/user.server.test.ts
+++ b/test/unit/user.server.test.ts
@@ -62,7 +62,7 @@ describe('deactivateUserByUsername()', () => {
   });
 
   afterAll(async () => {
-    await deleteUserByUsername(user.username);
+    await prisma.user.deleteMany().catch(() => {});
   });
 
   test('sets deactivated flag to true', async () => {
@@ -96,8 +96,7 @@ describe('isDeactivated()', () => {
   });
 
   afterAll(async () => {
-    await deleteUserByUsername(activeUser.username);
-    await deleteUserByUsername(deactivatedUser.username);
+    await prisma.user.deleteMany().catch(() => {});
   });
 
   test('returns true for deactivated user', async () => {
--- a/test/unit/user.server.test.ts
+++ b/test/unit/user.server.test.ts
@@ -4,9 +4,55 @@ import {
   deactivateUserByUsername,
   deleteUserByUsername,
   isDeactivated,
+  getUserByUsername,
+  isStudent,
+  isAdmin,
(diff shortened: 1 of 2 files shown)


test/unit/user.server.test.ts
```diff
@@ -4,9 +4,55 @@ import {
...
+    expect(user.deactivated).toBe(false);
+  });
+
+  test('creating an Ueser with same username should throw error', async () => {
+    await expect(
+      createUser('jsmith', 'Joanna Smith', 'j-smith@myseneca.ca', 'mycustomdomain-admins')
+    ).rejects.toThrow();
+  });
+
+  test('creating an Ueser with same email should throw error', async () => {
```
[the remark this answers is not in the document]
cychu42: I don't know who this Ueser is. My brain seems to like to bring them up.

SerpentBytes (approved):

test/unit/user.server.test.ts
```diff
@@ -4,9 +4,55 @@ import {
...
+    user = await createUser(
+      'jsmith',
+      'John Smith',
+      'jsmith@myseneca.ca',
+      'mycustomdomain-students'
+    );
+  });
+
+  afterAll(async () => {
+    await prisma.user.deleteMany().catch(() => {});
```
Eakam1007, line 30: Could use `deleteUserByUsername` instead similar to the rest of the file, or is there a reason you went with this approach?

cychu42: I prefer this as a catch-all way of cleaning all users. If anyone ever create more than one user as part of a test suite, like in the last suite starting on line 166, it can handle the clean-up without having to be modified.
Using `deleteUserByUsername()` means you have to specifically delete that one user, and it requires test writers to keep track of things.

cychu42: Plus, I'm leaning toward not even seeding the test database, to keep the state of the test database "clean". Another reason I want to clean up all users.
But that's probably another discussion/issue.

SerpentBytes (approved):

test/unit/user.server.test.ts
Eakam1007: Thank you!

Eakam1007 (approved):

PR #419, opened 2023-03-22
Refactor DNS away from flows, deal with errors

### LINKED ISSUE #368: Fail earlier with DNS issues
I tried creating a DNS record on staging, and it's failing due to what appears to be a config/permissions error:

```
starchart_mycustomdomain-dev.1.spbo7h7vjgwx@cudm-mgmt01dv.dcm.senecacollege.ca    | {"$fault":"client","$metadata":{"attempts":1,"httpStatusCode":403,"requestId":"a250980e-b074-41da-85e4-08cc5a7d9e66","totalRetryDelay":0},"Code":"AccessDenied","Type":"Sender","level":"warn","message":"User: arn:aws:iam::REDACTED is not authorized to perform: route53:ChangeResourceRecordSets on resource: arn:aws:route53:::REDACTED because no identity-based policy allows the route53:ChangeResourceRecordSets action","name":"AccessDenied","timestamp":"2023-03-17T15:28:22.328Z"}
starchart_mycustomdomain-dev.1.spbo7h7vjgwx@cudm-mgmt01dv.dcm.senecacollege.ca    | {"$fault":"client","$metadata":{"attempts":1,"httpStatusCode":403,"requestId":"cc2ff934-ea00-4035-aa38-55c86e1b43bb","totalRetryDelay":0},"Code":"AccessDenied","Type":"Sender","level":"warn","message":"User: arn:aws:iam::REDACTED is not authorized to perform: route53:ChangeResourceRecordSets on resource: arn:aws:route53:::REDACTED because no identity-based policy allows the route53:ChangeResourceRecordSets action","name":"AccessDenied","timestamp":"2023-03-17T15:28:37.589Z"}
starchart_mycustomdomain-dev.1.spbo7h7vjgwx@cudm-mgmt01dv.dcm.senecacollege.ca    | {"$fault":"client","$metadata":{"attempts":1,"httpStatusCode":403,"requestId":"99b035ba-1c17-4518-89db-54b69a2c63a3","totalRetryDelay":0},"Code":"AccessDenied","Type":"Sender","level":"warn","message":"User: arn:aws:iam::REDACTED is not authorized to perform: route53:ChangeResourceRecordSets on resource: arn:aws:route53:::REDACTED because no identity-based policy allows the route53:ChangeResourceRecordSets action","name":"AccessDenied","timestamp":"2023-03-17T15:29:07.734Z"}
starchart_mycustomdomain-dev.1.spbo7h7vjgwx@cudm-mgmt01dv.dcm.senecacollege.ca    | {"$fault":"client","$metadata":{"attempts":1,"httpStatusCode":403,"requestId":"aa4a0070-fa9f-4b43-8d99-8c73d7e87fe6","totalRetryDelay":0},"Code":"AccessDenied","Type":"Sender","level":"warn","message":"User: arn:aws:iam::REDACTED is not authorized to perform: route53:ChangeResourceRecordSets on resource: arn:aws:route53:::REDACTED because no identity-based policy allows the route53:ChangeResourceRecordSets action","name":"AccessDenied","timestamp":"2023-03-17T15:30:08.063Z"}
starchart_mycustomdomain-dev.1.spbo7h7vjgwx@cudm-mgmt01dv.dcm.senecacollege.ca    | {"$fault":"client","$metadata":{"attempts":1,"httpStatusCode":403,"requestId":"d1b53823-f8b8-4765-9fe0-b05c7c2c83b0","totalRetryDelay":0},"Code":"AccessDenied","Type":"Sender","level":"warn","message":"User: arn:aws:iam::REDACTED is not authorized to perform: route53:ChangeResourceRecordSets on resource: arn:aws:route53:::REDACTED because no identity-based policy allows the route53:ChangeResourceRecordSets action","name":"AccessDenied","timestamp":"2023-03-17T15:32:08.322Z"}
```

However, it never gets to the `error` state on the server or in UI:

Screenshot 2023-03-17 at 11 49 37 AM

I'm discussing the config/permissions with ITS, but I wanted to talk about the failure case in our DNS flow.  These 5 errors happen pretty quickly, and then nothing happens.

We should fail sooner I think.  In this case, no amount of waiting is going to fix the problem.

In the case of being Throttled by AWS, we'll get this:

> Five requests per second per AWS account per Region. If you submit more than five requests per second in a Region, Resolver returns an HTTP 400 error (Bad request). The response header also includes a Code element with a value of Throttling and a Message element with a value of Rate exceeded.

If we get throttled, retrying later is good.  If we are failing some call to AWS, retrying forever isn't likely to fix things.

cc @Genne23v, @dadolhay

humphd:
Fixes #368 

This refactors our DNS queue/workers to properly deal with the different stages of our DNS "pipeline."  Previously we used a Flow in BullMQ, but the way that errors work between child and parent jobs meant that we couldn't properly process error cases.

This switches the architecture to use a [Step Job pattern](https://docs.bullmq.io/patterns/process-step-jobs).  The main reason I'm going this route is that I want to group all of the Route53 calls into a single queue, which I can limit globally (i.e., 5 per second max).  By having everything in one queue, it makes it easier.

Here is what I've done:

1. Created `app/queues/dns/queue.server.ts` to manage the new `dns-queue`.  It uses `QueueEvents` to allow listening for `failed` jobs.  When a job in the queue fails, we have a chance to update the database.
2. Created `app/queues/dns/worker.server.ts` to define the worker.   It uses a series of steps instead of our Flow, and the code progresses through these steps.  Data is stored on the Job's `data` vs. being passed between workers.  The step logic is basically the same thing @Genne23v did before, just moved to different functions.
3. Created `app/queues/dns/index.server.ts` as the main API people use to work with this.  It exposes methods for adding jobs to the queue.  These are also using the code @Genne23v wrote before, just moving things around.

Elsewhere, I've updated all the call sites to use the new API names.  I've also made some other changes that were necessary for testing:

1. I've updated `app/routes/__index/dns-records/index.tsx` so that you can Delete a record when it's in the Error state.  Currently you can't
2. I've updated `deleteDnsRecord` to return `undefined` vs. `null` when no Change ID is returned (i.e., record already deleted in Route53).  This makes the types easier to write.

To test this, try creating, update, deleting records in the UI.


app/models/dns-record.server.ts
```diff
@@ -1,6 +1,7 @@
 import { DnsRecordStatus } from '@prisma/client';
 import dayjs from 'dayjs';
 import { prisma } from '~/db.server';
+import logger from '~/lib/logger.server';
```
Genne23v, line 4: This line needs to be removed

app/queues/dns/dns-worker.server.ts
```diff
@@ -0,0 +1,144 @@
...
+        default: {
+          throw new UnrecoverableError(`invalid step: ${step}`);
+        }
+      }
+    }
+  },
+  {
+    connection: redis,
+    limiter: {
+      max: 5,
```
Genne23v, line 73: I think we use the max number request as we have only one worker for Route53 handling, right?

app/queues/dns/dns-worker.server.ts
```diff
@@ -0,0 +1,144 @@
...
+      duration: 1_000,
+    },
+  }
+);
+
+async function executeChange({ dnsOperation, username, type, fqdn, value }: DnsChangeData) {
+  fqdn = fqdn.toLowerCase();
```
Genne23v, line 80: I initially lowercased the domain because of database, but removed this line in last commit because Route53 always converts domain to lowercase. I tested and confirmed.

app/queues/dns/index.server.ts
```diff
@@ -0,0 +1,94 @@
...
+import type { DnsJobData } from './dns-worker.server';
+
+type Subdomain = { subdomain: string };
+
+function addJob(jobName: string, jobData: DnsJobData) {
+  return dnsQueue.add(jobName, jobData, {
+    attempts: 3,
+    backoff: {
+      type: 'exponential',
+      delay: 60_000,
```
Genne23v, line 17: I know DNS record update will take maximum 60 seconds. So I assumed most cases it will be less than 60 seconds and used much less number to finish the job as early as possible. I guess this 60 second is intended to avoid any too early retry.

app/routes/dev.tsx
```diff
@@ -12,14 +12,13 @@ import logger from '~/lib/logger.server';
 import { requireUser, requireUsername } from '~/session.server';
```
Genne23v, line 12: I'm not sure about others, but I'm ok to remove this dev page.  @SerpentBytes @dadolhay

Genne23v (changes requested):
Thank you for the big change! I left a few comments on my review.

app/queues/dns/index.server.ts
humphd: I went with 10s, what do you think?

app/routes/dev.tsx
humphd: I'll leave it in here for now, but we can rip it out later this week or next.

humphd:
New commit up with fixes.

app/queues/dns/dns-worker.server.ts
```diff
@@ -0,0 +1,142 @@
...
+  try {
+    switch (dnsOperation) {
+      case 'delete':
+        if (dnsStatus === 'INSYNC') {
+          return deleteDnsRecordById(id);
+        }
+        break;
+      default:
+        return updateDnsRecordById(id, {
+          status: dnsStatus === 'INSYNC' ? DnsRecordStatus.active : DnsRecordStatus.error,
```
Genne23v, line 135: I'm not sure if this is expected. I returned `undefined` when creating a record, it updates the DB as `active`. I know we are going to have reconciler, but the user will see my record is available right away. Probably we should add `switch` by job type in `waitOnChange` to handle `create` differently?

Genne23v (changes requested):

app/routes/__index/dns-records/index.tsx
```diff
@@ -55,21 +55,20 @@ export const action = async ({ request }: ActionArgs) => {
...
-    throw new Response('DNS Record is not active, action forbidden', {
-      status: 409,
-    });
-  }
-
   switch (intent) {
     case 'renew-dns-record':
+      if (dnsRecord.status !== 'active') {
+        throw new Response('DNS Record is not active, action forbidden', {
+          status: 409,
```
SerpentBytes, line 62: Should this error code be `403` (forbidden) or `409` (conflict)? A user could mistakenly press the renew button in the front end, thinking it would refresh the status.

app/queues/dns/index.server.ts
humphd: NOTE: the way exponential backoff works, it's `2^retry * delay`, so that means 10s, 20s, 80s, 160s, etc.  If we did 10s and 6, it would get to 640s = 10.6 hours.

app/queues/dns/dns-worker.server.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    while (step !== Step.Finished) {
+      switch (step) {
+        case Step.Route53Change: {
+          const changeId = await executeChange({
+            ...job.data,
+          });
+          await job.update({ ...job.data, changeId, step: Step.WaitOnRoute53Change });
+        }
+
```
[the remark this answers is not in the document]
humphd: Indeed!  I also need to update `step`.  **Great** catch!

app/queues/dns/dns-worker.server.ts
humphd: This is exactly what we want.  When we `create` the DB record will be available right away, but `pending`.  When it finishes, we update to be either `active` or `error` and the UI changes to match.

app/queues/dns/index.server.ts
Genne23v: I think 10s with 3 retry reasonable

humphd:
I've made the review fixes, thank you both for helping me improve this!

Myrfion (approved):

humphd:
@SerpentBytes @Genne23v this is ready for another look

Genne23v (approved):
It looks good to me!

SerpentBytes (approved):

PR #421, opened 2023-03-23
Update playwright and add --ui script via e2e

humphd:
Today I was showing @Eakam1007 the new release of Playwright, and the amazing `--ui` flag they added, see https://www.youtube.com/watch?v=jF0yA-JLQW0

It lets you run your tests in this cool visual timeline view.  I've never seen anything quite like it.

I update our code to use it.  To test:

```
npm install
npm run e2e
```

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "start:e2e": "cross-env NODE_ENV=test node --require dotenv/config ./build/server.js",
     "test": "cross-env SECRETS_OVERRIDE=1 DATABASE_URL='mysql://root:root_password@127.0.0.1:3306/starchart_test' NODE_OPTIONS=--require=dotenv/config vitest",
     "test:coverage": "cross-env SECRETS_OVERRIDE=1 DATABASE_URL='mysql://root:root_password@127.0.0.1:3306/starchart_test' NODE_OPTIONS=--require=dotenv/config vitest run --coverage",
+    "e2e": "cross-env PORT=8080 DATABASE_URL='mysql://root:root_password@127.0.0.1:3306/starchart_test' start-server-and-test dev http://localhost:8080 \"playwright test --ui\"",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -25,6 +25,7 @@
     "start": "cross-env NODE_ENV=production node ./build/server.js",
     "start:e2e": "cross-env NODE_ENV=test node --require dotenv/config ./build/server.js",
     "test": "cross-env SECRETS_OVERRIDE=1 DATABASE_URL='mysql://root:root_password@127.0.0.1:3306/starchart_test' NODE_OPTIONS=--require=dotenv/config vitest",
+    "e2e": "cross-env PORT=8080 DATABASE_URL='mysql://root:root_password@127.0.0.1:3306/starchart_test' start-server-and-test dev http://localhost:8080 \"playwright test --ui\"",
```
Eakam1007, line 28: Should we name this something else, similar to `test:e2e:dev`? Or would the name get too long?

Eakam1007 (approved):
The UI thing is awesome, solves the issues with using the VS extension and `baseURL`. Thanks for this!

package.json
humphd: I hate these names, I can never remember what they are.  I want something that's easy to use and means "I'm a dev who wants to run the e2e tests..."  The long names are fine for scripts/CI.  What do you think?

sfrunza13 (approved):
In reference to the above conversation I like the shorter names myself (particularly for those scripts most frequently used).

package.json
Eakam1007: Yea, that makes sense. The names are quite long

humphd:
I've rebased and also updated the Wiki, see https://github.com/DevelopingSpace/starchart/wiki/Playwright#running-using-scripts

Eakam1007 (approved):
Thanks for updating the wiki

PR #424, opened 2023-03-23
Exclude acme challenge in user account

### LINKED ISSUE #413: Don't count _acme-challenge records against user's record limit
The Certificate creation process creates `TXT` DNS Records named `_acme-challenge`. We shouldn't be counting these when we calculate the user's DNS Record count (i.e., don't count them in the 20 that a user gets).

Genne23v:
Close #413 

Exclude `_acme-challenge` record and count only active records.

### DIFF
--- a/app/models/dns-record.server.ts
+++ b/app/models/dns-record.server.ts
@@ -1,4 +1,4 @@
-import { DnsRecordStatus } from '@prisma/client';
+import { DnsRecordStatus, DnsRecordType } from '@prisma/client';
 import dayjs from 'dayjs';
 import { prisma } from '~/db.server';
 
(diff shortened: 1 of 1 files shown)


app/models/dns-record.server.ts
```diff
@@ -33,12 +33,25 @@ export function getDnsRecordById(id: DnsRecord['id']) {
...
       username,
+      type: DnsRecordType.TXT,
+      subdomain: '_acme-challenge',
     },
   });
+};
+
+export async function getUserDnsRecordCount(username: DnsRecord['username']) {
+  let userRecordCount = await prisma.dnsRecord.count({
+    where: {
```
humphd, line 48: Instead of 2 functions for count, we add a boolean that is default `true` to the function, maybe `excludeAcmeRecords` and use it to set `NOT` on the `where`

https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#not-1

humphd (changes requested):

app/models/dns-record.server.ts
Genne23v: Thank you for the recommendation. I updated my PR!

Genne23v:
> 

Thanks for the mention. I just updated `getDnsRecordsByUsername`.

humphd (approved):
I think this is right. Nice work.

humphd (approved):

cychu42 (approved):
The code looks fine. Nice job!

> Do we ever want to get back the ACME TXT records for a user?

 It's probably fine. If we ever need to, fetching `_acme-challenge` records would be a different use case from fetching "normal" records of users and can still be queried by `prisma.dnsRecord.findMany()`.

PR #427, opened 2023-03-23
changed session cookie to lax

### LINKED ISSUE #426: Login on staging requires 2 attempts
Since we updated our SAML login this week, staging is now requiring you to login twice.  The first time you are redirected back to the login page, and the second time you login it works.

@sfrunza13 wonders on Slack if it's related to changes we made to cookies/session.

sfrunza13:
fixes #426 I think

### DIFF
--- a/app/session.server.ts
+++ b/app/session.server.ts
@@ -17,7 +17,7 @@ export const sessionStorage = createCookieSessionStorage({
     name: '__session',
     httpOnly: true,
     path: '/',
-    sameSite: 'strict',
+    sameSite: 'lax',
     secrets: [secrets.SESSION_SECRET],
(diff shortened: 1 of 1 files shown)


humphd (approved):
To confirm: below where we still do `strict`, that's OK?

```ts
export const sloUsernameCookie = createCookie('sloUsername', {
  domain: process.env.NODE_ENV === 'production' ? new URL(process.env.APP_URL).hostname : undefined,
  path: '/logout',
  sameSite: 'strict',
```

sfrunza13:
> 

I think so since this is only used on /logout and all the requests to logout come from our site.

Eakam1007 (approved):

PR #428, opened 2023-03-23
Add tooltip for optional DNS record info in table

### LINKED ISSUE #304: Display DNS record description in a tooltip
@humphd on [Slack](https://seneca-open-source.slack.com/archives/C04HR4WV5P0/p1678194333718579):

Screenshot 2023-03-08 at 11 19 22 AM

## Details
We should place all the optional info and display it in a tooltip. Doing this would make the view more compact, and the relevant information would be hard to miss.

### LINKED ISSUE #425: Unable to update optional DNS Record values
While working on #419 I noticed a bug:

1. Create a new DNS Record, don't fill in the ports, course, or description (e.g., leave out all optional fields)
2. Wait until its `status` has changed to `active`
3. Click the Edit button
4. Add the optional fields so they all have a value
5. Click the Update button
6. Wait until its `status` has changed to `active`
7. Click the Edit button
8. Notice the optional fields are all empty

cc @Genne23v, @Myrfion

humphd:
Fixes #304 

If we have optional data for a DNS Record, this shows it in a tooltip when you hover over the name.  If we don't, we don't show anything.  I had thought to use an icon like CloudFlare, but since most records won't have all this data, it seemed like a waste to show it and do nothing with it.  This way the data reveals itself if we have it, but doesn't clutter the UI.

Screenshot 2023-03-23 at 5 24 19 PM

I think we need to fix #425 to test this properly, but you can use `npm run db:studio` to manually enter the values on an existing record.

### DIFF
--- a/app/components/dns-record/dns-record-name.tsx
+++ b/app/components/dns-record/dns-record-name.tsx
@@ -1,16 +1,39 @@
-import { Flex, Text } from '@chakra-ui/react';
+import { Flex, Text, Tooltip } from '@chakra-ui/react';
+
+import type { DnsRecord } from '@prisma/client';
 
 interface DnsRecordNameProps {
-  subdomain: string;
+  dnsRecord: DnsRecord;
   baseDomain: string;
 }
 
-const DnsRecordName = ({ subdomain, baseDomain }: DnsRecordNameProps) => {
-  return (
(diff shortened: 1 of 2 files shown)


Eakam1007 (approved):

Ririio (approved):
Oh so this is how it worked, I don't see any issues with it

PR #429, opened 2023-03-23
Hide table if no DNS Records exist

### LINKED ISSUE #403: First run experience for DNS Records page
The first time you use the site and go to the DNS Records, you're met with this empty table:

Screenshot 2023-03-21 at 4 56 21 PM

Let's not show the table when the user has no records, and instead show some text in the centre with the "Create" button that says something like, "Create your first DNS Record!"

cc @Ririio, @Myrfion

Eakam1007:
Resolves #403 

This changes the DNS Records tables so that:
- The text container takes up the same width as the DNS Records table. This was done because the DNS Records table takes up the full width
- If there are no DNS Records, a create button will be displayed in the center of the page

#### Without DNS Records
image

#### With DNS Records
image

### DIFF
--- a/app/routes/__index/dns-records/index.tsx
+++ b/app/routes/__index/dns-records/index.tsx
@@ -1,6 +1,6 @@
 import { useMemo } from 'react';
 import { AddIcon } from '@chakra-ui/icons';
-import { Button, Container, Flex, Heading, Text } from '@chakra-ui/react';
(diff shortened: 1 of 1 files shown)


humphd (approved):

Ririio (approved):
Looks good, didn't see any issue

PR #430, opened 2023-03-24
Improve landing page link accessibility

### LINKED ISSUE #341: Improve anchor text accessibility on landing page
Screenshot 2023-03-14 at 5 39 57 PM

The light blue text colour for links is not using enough contrast.  I think we need to pick something else.  Does blue even go with your colour theme?

cc @Myrfion

Eakam1007:
Resolves #341 

Changed the color of the link text to use one of your brand colors to increase the contrast and improve accessibility. 
Also modified the link text slightly (by capitalizing the `T`, and making it its own sentence)

image

### DIFF
--- a/app/components/landing-page/landing-page-card.tsx
+++ b/app/components/landing-page/landing-page-card.tsx
@@ -28,8 +28,8 @@ export default function LandingPageCard({
       <Flex width={{ md: 'xs' }} height={{ sm: 'xl' }} flexDirection="column">
         <Text fontSize={{ base: 'xs', md: 'md' }}>
           {cardDescription}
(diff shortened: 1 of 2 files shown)


humphd (approved):
This is excellent.  Question: should we make this the default in the theme, so we don't have to manually do this elsewhere when we use `<a>`?

Eakam1007:
> This is excellent. Question: should we make this the default in the theme, so we don't have to manually do this elsewhere when we use `<a>`?

I initially went with this, but it also affects the links in the header: `DNS Records` and `Certificates`

humphd:
Ah, makes sense.  This is a great change as is.

Myrfion (approved):

PR #431, opened 2023-03-24
Improve DNS record renew button usability

### LINKED ISSUE #389: Improve usability of record renew button
Let's move the renew button closer to the expiry date, so it's clear what it goes with.

Screenshot 2023-03-20 at 8 41 12 AM

Also, let's disable the button if the expiry is already 6 months out from the current date (i.e., if there is no need to renew it yet).

@Myrfion or @Ririio, can one of you do this?

Eakam1007:
Resolves #389 

### Changes
- Moved renew button closer to expiry date
- Added logic to disable renew button if DNS Record's expiry date is 6 months out or more

image

### DIFF
--- a/app/components/dns-records-table.tsx
+++ b/app/components/dns-records-table.tsx
@@ -30,6 +30,7 @@ import DnsRecordDeleteAlertDialog from './dns-record-delete-alert-dialog';
 import { Form, useNavigate, useTransition } from '@remix-run/react';
 import DnsRecordName from './dns-record/dns-record-name';
 import { useUser } from '~/utils';
+import dayjs from 'dayjs';
 
 interface DnsRecordsTableProps {
   dnsRecords: DnsRecord[];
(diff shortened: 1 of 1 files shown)


humphd (approved):

app/components/dns-records-table.tsx
```diff
@@ -132,6 +133,7 @@ export default function DnsRecordsTable(props: DnsRecordsTableProps) {
                 const isLoading =
                   transition.state === 'submitting' &&
                   Number(transition.submission.formData.get('id')) === dnsRecord.id;
+                const renewalAllowed = dayjs(dnsRecord.expiresAt).isBefore(dayjs().add(6, 'month'));
```
Myrfion, line 136: Since all boolean variables there has the suffix `is`, maybe it would be better to add it to yours too to keep consistency in naming. So `isRenewalAllowed`

Myrfion (changes requested):

app/components/dns-records-table.tsx
Eakam1007: Hmmm, to me it reads better without the `is`. Do you think it would be better to remove the `is` from the other variables too?

humphd: `isRenewable` would be good.  I'd say that the `is*` style is already in place, let's match it.

Eakam1007: Oh, `isRenewable` is good. I will change it

Myrfion (approved):

humphd (approved):

PR #432, opened 2023-03-24
Fix html, body styling, update head

humphd:
I was testing the site in Lighthouse, and it was complaining that we didn't have a `<!DOCTYPE html>`.  We do have it, but the way that the styles for `html` and `body` were being done by Chakra UI, it was putting a `<style>` tag outside the document.  This fixes it.

I also updated the `<head>` to fix the title and description.

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -1,4 +1,4 @@
-import { ChakraProvider, chakra } from '@chakra-ui/react';
+import { ChakraProvider } from '@chakra-ui/react';
 import { json } from '@remix-run/node';
 import {
   Links,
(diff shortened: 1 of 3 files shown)


Myrfion (approved):

Eakam1007 (dismissed):
Interesting that it was putting a style tag outside before, thanks for the fix

SerpentBytes (dismissed):

humphd:
I had to modify this a bit to work on the login page.

SerpentBytes (approved):

Eakam1007 (approved):

Issue #393, opened 2023-03-20
Connecting back-end for the certificate UI


Ririio:
Implementing the back-end so that the certificate UI receives all the necessary components to run on real-time instead of the mock values

PR #322, opened 2023-03-11
fix: problems in notifications related code

SerpentBytes:
## Description
This PR  fixes problematic code in `sendNotification`, as identified in #262 

### Steps to Test
- Run the project

PR #434, opened 2023-03-24
Back-End Connection with Certificate UI

Ririio:
Fixes: #393, Fixes: #322 

- Replaced `getIssuedCertificateByUsername` with status to make the transition between 3 states more seemles
  - The loader is stored in a `certificate` to check for the current state
  - `useRevalidator` ensures that it will display components depending on the status of the certificate
- Should solve issue where dns record is populated with `_acme` records

Demo:
https://user-images.githubusercontent.com/32577022/227579673-cbcefb3c-aa38-4b15-9102-38e01f989998.mp4


app/components/certificate/certificate-request.tsx
```diff
@@ -38,9 +41,16 @@ export default function CertificateRequestView({ domain, onRequest }: Certificat
...
       </Flex>
-      <Button width={{ base: 'full', md: '3xs' }} shadow="xl" onClick={onRequest}>
-        Request a Certificate
-      </Button>
+      <Form method="post" onSubmit={() => setIsDisabled(true)}>
+        <Button
+          type="submit"
+          width={{ base: 'full', md: '3xs' }}
+          shadow="xl"
+          isDisabled={isDisabled}
```
humphd, line 49: This is good for now, but I wonder if in a follow-up we should update the page with some text that says, "OK, we're doing it, hang tight..."

app/models/certificate.server.ts
```diff
@@ -4,18 +4,19 @@ import { prisma } from '~/db.server';
...
-   * the most recent one that has been successfully issued
-   */
+export async function getCertificateStatusByUsername(username: Certificate['username']) {
   return prisma.certificate
     .findMany({
-      where: { username, status: 'issued' },
-      orderBy: { validFrom: 'desc' },
+      where: {
+        username,
+        OR: [{ status: 'issued' }, { status: 'pending' }, { status: 'failed' }],
```
humphd, line 12: This defines all 3 possible states, which implies that you don't actually need to include this.  Is there a reason you do?

app/models/certificate.server.ts
```diff
@@ -4,18 +4,19 @@ import { prisma } from '~/db.server';
...
+      },
+      orderBy: { id: 'desc' },
       take: 1,
     })
-    .then(([certificate]) => certificate);
+    .then(([certificate]) => {
+      return certificate || {};
```
humphd, line 18: I don't think we should return an empty object here.  Let's solve this problem at the call site (e.g., synthesize an object if you need one in React).

app/routes/__index/certificate/index.tsx
```diff
@@ -1,57 +1,80 @@
...
 
-export default function CertificateIndexRoute() {
-  const user = useUser();
-  const [certificateRequested, setCertificateRequested] = useState(false);
-  const [loading, setLoading] = useState(false);
+export const loader = async ({ request }: LoaderArgs) => {
+  const user = await requireUser(request);
 
-  const currentDate: Date = new Date(2023, 2);
+  return await getCertificateStatusByUsername(user.username);
```
humphd, line 21: Get rid of `await` here.  Also, since a certificate will include dates, should we use @Myrfion's trick with returning `typedjson`?  See for example https://github.com/DevelopingSpace/starchart/blob/d4b1db4e5eb873d1da4e34f29d71bdaabb8b2400/app/routes/__index/dns-records/%24dnsRecordId.tsx#L32

app/routes/__index/certificate/index.tsx
```diff
@@ -1,57 +1,80 @@
...
+      : null
+  );
 
   function formatDate(val: Date): string {
-    let date = val.toLocaleDateString('en-US', {
-      month: 'short',
-      day: '2-digit',
-      year: 'numeric',
-    });
+    // let date = val.toLocaleDateString('en-US', {
```
humphd, line 56: Remove code you don't need vs. commenting out

app/routes/__index/certificate/index.tsx
```diff
@@ -1,57 +1,80 @@
...
-  const currentDate: Date = new Date(2023, 2);
+  return await getCertificateStatusByUsername(user.username);
+};
 
-  const certificate: Certificate = {
-    id: 1,
+export const action = async ({ request }: ActionArgs) => {
+  const user = await requireUser(request);
+
+  await addCertRequest({
```
humphd, line 27: I would confirm the request verb here (e.g., `POST`?)

humphd: Also, you should probably confirm that there isn't a request already pending.  I know you disable the UI to do this, but someone could `POST` manually to do this again.

app/routes/__index/certificate/index.tsx
```diff
@@ -1,57 +1,80 @@
...
+export default function CertificateIndexRoute() {
+  const user = useUser();
+  const revalidator = useRevalidator();
+  const certificate = useTypedLoaderData<typeof loader>();
+
+  useInterval(
+    () => {
+      revalidator.revalidate();
+    },
+    certificate.status === 'pending' ||
```
humphd, line 47: Similar to the above, can you explain what you're trying to do here?  I'm confused by the scope of this check.

humphd (changes requested):
Love the video demo, thank you!  And thank you for pushing on this so we can get it in for 0.7.  Great stuff.

I've done a first pass through the code.

app/models/certificate.server.ts
Ririio: I was wondering how we handle certificate that has failed? I thought that we delete them if that happens, but I'm not quite sure...

app/models/certificate.server.ts
Ririio: Synthesize?

app/routes/__index/certificate/index.tsx
Ririio: Oh yeah I see the problem thank you I will change it. Also, by "confirming" what do you mean exactly?

app/routes/__index/certificate/index.tsx
Ririio: I want the page to refresh when the current status of certificate is either `pending`, `failed` or `issued`. This to display their respective components

app/models/certificate.server.ts
humphd: @dadolhay can confirm, but when we finish an order currently only have code to set `status=issued` - https://github.com/DevelopingSpace/starchart/blob/main/app/queues/certificate/order-completer-worker.server.ts#L76-L82.  It doesn't look to me like we have an error handler to set it equal to `failed` yet.  We were discussing this yesterday, and it hasn't been done, but it will be `status=failed`.

app/routes/__index/certificate/index.tsx
humphd: This condition is saying "whenever anything" which is going to be hard to do.

I suspect that you want to think about this like so:

1. is `certificate` undefined?  (i.e., nothing in DB)
2. if `certificate` is defined, is the status `pending`?  If so, run the interval
3. if `certificate` is defined`, and status is `!pending`, stop the interval

humphd (changes requested):

app/models/certificate.server.ts
```diff
@@ -41,6 +42,10 @@ export function updateCertificateById(
   });
 }
 
+export function deleteAllByUsername(username: Certificate['username']) {
```
SerpentBytes, line 45: The function `deleteAllByUsername` name is ambiguous. I would change the id to `deleteAllCertificatesByUsername` or something more explicit.

Ririio: Sure that's a good idea

app/models/certificate.server.ts
Ririio: I can create a try and catch, and when it fails, it will instead return an empty object with `status=failed` to send the user to the certificate "request" page

app/routes/__index/certificate/index.tsx
Ririio: oh you're right... there's no need to refresh the page if there's no loading. Okay I'll change it

app/models/certificate.server.ts
Ririio: What do you mean exactly? Don't you need an `OR` for conditional?

app/routes/__index/certificate/index.tsx
Ririio: I decided to simply go with `certificate.status === 'pending' ? 5_000 : nulll` due to the fact that there is no need for the page to refresh unless they are in the loading page.

app/models/certificate.server.ts
humphd: At the level of the DB model, just returned undefined or a `certificate`, and do this other stuff in the API layer above.

app/routes/__index/certificate/index.tsx
humphd: Check what the request method is

app/routes/__index/certificate/index.tsx
humphd: perfect

app/models/certificate.server.ts
Ririio: I wrote the try and catch on the loader itself, not sure if that's what you meant...

app/models/certificate.server.ts
humphd: If you're saying, `it's any of these` and those are the only values we allow, there's no point in doing the condition.  Every record in our DB is one of `issued`, `pending`, or `failed`, so this will return the whole set.

app/models/certificate.server.ts
humphd: OK, you can undo this change then.

app/routes/__index/certificate/index.tsx
```diff
@@ -1,43 +1,75 @@
...
-  const [certificateRequested, setCertificateRequested] = useState(false);
-  const [loading, setLoading] = useState(false);
+export const loader = async ({ request }: LoaderArgs) => {
+  const user = await requireUser(request);
+  let certificate;
+
+  try {
+    certificate = await getCertificateStatusByUsername(user.username);
+  } catch {
+    certificate = {
```
humphd, line 24: I would just let certificate stay undefined.  Below, you can use `certificate?.status` to check if there is an Object before you try to access `.status`

humphd (changes requested):

app/models/certificate.server.ts
Ririio: oh wait you're right

test/e2e/certificate.spec.ts
```diff
@@ -1,21 +1,29 @@
...
-  });
+    test('Request a Certificate Page', async ({ page }) => {
+      await page.goto('/certificate');
+      await deleteAllCertificateByUsername('user1');
+    });
 
-  test('Request a Certificate', async ({ page }) => {
     const titleHeader = page.getByRole('heading', { name: 'Certificate' });
-    const domainName = page.getByText('user1.starchart.com');
+    await page.waitForSelector('text=user1.starchart.com');
```
humphd, line 21: Same question here.

test/e2e/certificate.spec.ts
```diff
@@ -1,21 +1,29 @@
...
 
-    await expect(domainName).toContainText('user1.starchart.com');
-    await expect(titleHeader).toContainText('Certificate');
+    await titleHeader.waitFor();
 
     await page.getByRole('button', { name: 'Request a Certificate' }).click();
+    createCertificate({ username: 'user1', domain: 'user1.starchart.com' });
```
humphd, line 26: And here

test/e2e/certificate.spec.ts
```diff
@@ -24,7 +32,14 @@ test.describe('Certificate Page', () => {
       })
       .nth(1);
 
-    await loadingPageText.waitFor();
+    loadingPageText.waitFor();
+
+    const certificate = await getCertificateByUsername('user1');
```
humphd, line 37: And here.

app/routes/__index/certificate/index.tsx
```diff
@@ -1,43 +1,68 @@
...
 
-export default function CertificateIndexRoute() {
-  const user = useUser();
-  const [certificateRequested, setCertificateRequested] = useState(false);
-  const [loading, setLoading] = useState(false);
+export const loader = async ({ request }: LoaderArgs) => {
+  const user = await requireUser(request);
 
-  const currentDate: Date = new Date(2023, 2);
+  const certificate = (await getCertificateByUsername(user.username)) || {};
```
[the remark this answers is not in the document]
Ririio: `typedjson` prevents me from moving forward if it receives an empty array... I can  pass an empty object `return typedjson(certificate || {});` or throw an error, but if I do the latter nothing will ever be executed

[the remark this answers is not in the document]
humphd: Probably if you do this, it will be OK:

```ts
return typedjson({ certificate });
```

So you always return an Object, but it may or may not have a `certificate` inside.

app/routes/__index/certificate/index.tsx
```diff
@@ -1,43 +1,68 @@
...
+
+  useInterval(
+    () => {
+      revalidator.revalidate();
+    },
+    certificate?.status === 'pending' ? 5_000 : null
+  );
+
   function formatDate(val: Date): string {
     let date = val.toLocaleDateString('en-US', {
```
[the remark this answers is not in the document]
Ririio: I recall receiving an error with the client and server mismatching a while back, that's why I decided to do `en-US` has that been fixed recently?
image

app/routes/__index/certificate/index.tsx
```diff
@@ -65,15 +90,15 @@ export default function CertificateIndexRoute() {
         width={{ base: 'md', sm: 'lg', md: '2xl', lg: '4xl' }}
         marginTop={{ base: '16', md: '5' }}
       >
-        {certificateRequested ? (
+        {certificate?.status === 'issued' ? (
```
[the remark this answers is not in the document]
Ririio: failed would simply send them to the "Request View" page. I had it initially display a page similar to "Loading", but I wasn't quite sure how you guys handle a failed certificate, so I decided to remove that for now...

app/routes/__index/dns-records/instructions.tsx
```diff
@@ -0,0 +1,21 @@
+import { Heading, Text, Link } from '@chakra-ui/react';
```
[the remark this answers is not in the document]
Ririio: It was causing an issue with the E2E Test for the CI, I wasn't quite sure how everything was working up until now when the page never existed

app/routes/__index/certificate/index.tsx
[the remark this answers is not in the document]
Ririio: It's causing issue with `certificate?.status` not existing for `certificate`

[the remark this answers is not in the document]
Ririio: oh nvm it's merely due to the change in the structure of the object

app/routes/__index/certificate/index.tsx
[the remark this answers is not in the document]
humphd: Odd, OK, leave it.

app/routes/__index/dns-records/instructions.tsx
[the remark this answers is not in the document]
humphd: OK, leave it then.

app/models/certificate.server.ts
```diff
@@ -4,15 +4,18 @@ import { prisma } from '~/db.server';
 
 export type { Certificate } from '@prisma/client';
 
-export function getIssuedCertificateByUsername(username: Certificate['username']) {
+export async function getCertificateByUsername(username: Certificate['username']) {
```
humphd, line 7: Because you don't `await` in this function, and return the `Promise` directly, you don't need `async` here.

app/routes/__index/certificate/index.tsx
```diff
@@ -1,43 +1,80 @@
...
-  const [certificateRequested, setCertificateRequested] = useState(false);
-  const [loading, setLoading] = useState(false);
+export const loader = async ({ request }: LoaderArgs) => {
+  const username = await requireUsername(request);
+
+  try {
+    const certificate = await getCertificateByUsername(username);
+    return typedjson({ certificate });
+  } catch (error: any) {
+    throw new Error('Error retrieving certificate: ' + error.message);
```
humphd, line 24: @Myrfion is this the right way to do the error?

app/routes/__index/certificate/index.tsx
```diff
@@ -1,43 +1,80 @@
...
+  return json({
+    result: 'ok',
+    message: 'Certificate Requested',
+  });
+};
+
+export default function CertificateIndexRoute() {
+  const user = useUser();
+  const revalidator = useRevalidator();
+  const certificate = useTypedLoaderData<typeof loader>().certificate;
```
humphd, line 69:
```ts
const { certificate } = useTypedLoaderData<typeof loader>();
```

humphd (changes requested):
Looks good, a few small things and some questions.

app/routes/__index/certificate/index.tsx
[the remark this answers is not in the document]
Ririio: I have a component for displaying, but I'm not quite sure how to handle it, how I did it at the start war

***When `status===failed` display a page showing that it has failed***

However, as to what to do afterwards, I wasn't quite sure thus I removed it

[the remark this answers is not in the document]
humphd: That's a good question.  I think we'd want to provide a way to try again.  Let's ask @dadolhay on Wed what he thinks.

app/models/certificate.server.ts
Ririio: I did that before, but I did receive a warning that it might turn into an async
image

app/routes/__index/certificate/index.tsx
Ririio: oh I never knew this was a thing good to know

Ririio: oh I can't believe I've been using it without thoroughly understanding what it does

app/routes/__index/certificate/index.tsx
Myrfion: @Ririio I do errors kinda differently, here is the example:

```typescript
 throw new Response("Not Found", {
      status: 404
    });
```

Ririio: Sure I can change that. I was just wondering are we changing all error handling with `response` ?I thought they more for user interaction error like looking for a dnsRecord that doesn't exists, or routes that we don't have?

test/e2e/certificate.spec.ts
```diff
@@ -1,21 +1,29 @@
...
 
 test.describe('Certificate Page', () => {
   loggedInAsUser();
 
   test.beforeEach(async ({ page }) => {
-    await page.goto('/certificate');
-  });
+    test('Request a Certificate Page', async ({ page }) => {
+      await page.goto('/certificate');
+      await deleteAllCertificateByUsername('user1');
```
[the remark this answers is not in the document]
Eakam1007: This got buried in my emails; we could update the util function and the auth setup to use a specific username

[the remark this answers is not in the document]
humphd: Maybe we should do that in a follow-up issue.  I think it would be useful.

app/routes/__index/certificate/index.tsx
humphd: @Ririio: Do you want to file a follow-up and fix this there?  I'd like to avoid blocking this further

app/routes/__index/certificate/index.tsx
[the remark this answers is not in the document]
humphd: @Ririio: File a follow up on this too

humphd (approved):
Some follow-ups to file, but let's get this in after that and fix bugs on top vs. blocking more.

test/e2e/certificate.spec.ts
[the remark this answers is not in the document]
Eakam1007: Yea, it would also help with the mocking admins

app/routes/__index/certificate/index.tsx
Ririio: I created a follow up

app/models/certificate.server.ts
```diff
@@ -41,6 +44,10 @@ export function updateCertificateById(
   });
 }
 
+export function deleteAllCertificatesByUsername(username: Certificate['username']) {
```
dadolhay, line 47: Why do we need this? A cert will not become invalid just because we delete it from here, on the other hand we lose the ability to revoke it. `deleteCertificateById` was designed to delete outdated certificates that no longer work

app/routes/__index/certificate/index.tsx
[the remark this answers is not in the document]
dadolhay: Failed certificate requests are useless, so you can show that it failed if you want to, but you should allow the user to request a new one.

dadolhay (changes requested):

app/models/certificate.server.ts
Ririio: It was there previously, I simply used it for the test to work. Should I delete it?

dadolhay: Interesting, it did show up as new code. Regardless, it's gone now

dadolhay (approved):
Okay This is good enough, later we can re-visit

Eakam1007 (approved):

humphd:
@Ririio in future, please don't land [16 separate commits](https://github.com/DevelopingSpace/starchart/pull/434/commits), consolidate them into a single commit first by squashing.  It makes it harder to back things out if they go wrong, or to revert a bug later on.

PR #439, opened 2023-03-24
Update staging hosted zone and root domain

humphd:
This updates the Route53 Hosted Zone and Root Domain we use on staging.  I've also removed the unnecessary `restart`s, which aren't needed in swarm.


Genne23v (approved):
LGTM!

SerpentBytes (approved):

Issue #438, opened 2023-03-24
Fix database schema for multiple DNS Record values on same name+type


humphd:
We've been testing the Let's Encrypt Order flow on staging today, and it's been great for tracking down bugs.  We have lots!

First bug we found was that Let's Encrypt does not support `_` in a domain, only letters, numbers and dashes.  Our staging hosted zone was `_stage_.mystudentproject.ca`, so everything failed.  ITS has now changed it to `stage.mystudentproject.ca`.

Second bug is that Let's Encrypt wants to be able to write multiple `TXT` values to `_acme-challenge.{username}.{rootdomain}`. Our DNS/database logic assumes that a name+value is unique.  Not so!

Let's Encrypt tells us to create two `TXT` records:

```
{"challengeKey":"8cV4hs2A8VmH3a2f2QYkvANYtXZWm9I93kUXYZtiGgE","domain":"_acme-challenge.davidhumphrey.stage.mystudentproject.ca","level":"info","message":"Adding challenge to DNS","timestamp":"2023-03-24T21:08:11.407Z"}
{"challengeKey":"VWl3fxuuU1ay9BB3vbmejMfvT1PkMialo4G4hwrhp68","domain":"_acme-challenge.davidhumphrey.stage.mystudentproject.ca","level":"info","message":"Adding challenge to DNS","timestamp":"2023-03-24T21:08:11.408Z"}
```

We issue two calls to do a `create`, which is really an `upsert`:

```
{"level":"debug","message":"Route53 DNS change: create _acme-challenge.davidhumphrey.stage.mystudentproject.ca=8cV4hs2A8VmH3a2f2QYkvANYtXZWm9I93kUXYZtiGgE with type=TXT for user=david.humphrey","timestamp":"2023-03-24T21:08:11.441Z"}
{"level":"debug","message":"Route53 DNS change: create _acme-challenge.davidhumphrey.stage.mystudentproject.ca=VWl3fxuuU1ay9BB3vbmejMfvT1PkMialo4G4hwrhp68 with type=TXT for user=david.humphrey","timestamp":"2023-03-24T21:08:11.749Z"}
```

These then fail, I think because we have an in-flight upsert that collides with another:

```
[...]

Issue #396, opened 2023-03-20
DNS reconciler


dadolhay:
Our own internal DB should be the "primary source of truth".

We should
- Load all records from DB
- Load all records from route53
- Pair each db record with route53 record **on exact match only** (a record has to be identified by type, subdomain and content, as there may be multiple TXT records with the same subdomain but with different content, just to bring an example)
- loop through all the pairs:
-   If route53 part is missing => create it
-   if db part missing => delete record in route53
-   if both parts of the pair is present => nothing to do
- Collect all reauired changes take the first 1000 elements from the list
- send the changes to route53

---
We would also need to implement some kind of throttling for this that mimics lodash (but qith queues)
https://lodash.com/docs/#throttle
where `options = {leading: true, trailing: true}`

dadolhay:
This would also modify our state handling for records
We sould only have "pending" and "synced" statuses
When a record is created or altered => set it to pending
When the reconciler has finished with a specific record => set it to "synced"
When a record is deleted => delete it from the db directly, reconciler will take care of route53 the next time

This would also completely take care of the "cannot edit while being created/updated/deleted" problem

humphd:
According to the [AWS docs for ListResourceRecordSet](https://docs.aws.amazon.com/Route53/latest/APIReference/API_ListResourceRecordSets.html), we can get **up to 300 resource record sets at a time**.   The docs go on:

## Specifying where to start listing records

>You can use the name and type elements to specify the resource record set that the list begins with:
>
>If you do not specify Name or Type
> The results begin with the first resource record set that the hosted zone contains.
>
> If you specify Name but not Type
> The results begin with the first resource record set in the list whose name is greater than or equal to Name.
>
> If you specify Type but not Name
> Amazon Route 53 returns the InvalidInput error.
>
> If you specify both Name and Type
> The results begin with the first resource record set in the list whose name is greater than or equal to Name, and whose type is greater than or equal to Type.

So we "page" our way through > 300 records.  We might want to build some test code that creates 300+ in moto.

Also, the docs talk about `PENDING` records being returned too, so we'll have to be aware of this.

Next, we have to be careful not to issue `ChangeResourceRecordSets` commands while we are paging the results:

> To ensure that you get an accurate listing of the resource record sets for a hosted zone at a point in time, do not submit a ChangeResourceRecordSets request while you're paging through the results of a ListResourceRecordSets request. If you do, some pages may display results without the latest changes while other pages display results with the latest changes.

[...]

PR #441, opened 2023-03-24
feat: DNS reconciler engine

dadolhay:


dadolhay:
Part of #396

app/lib/dns.server.ts
```diff
@@ -165,6 +166,55 @@ async function checkDnsRecordExists(type: DnsRecordType, fqdn: string, value: st
...
 
+/**
+ * Get a full page of records from AWS Route53. If fqdn and recordType is specified,
+ * load the next page starting with that record
+ *
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-route-53/classes/listresourcerecordsetscommand.html
+ */
+export const getRecordPage = async (
+  fqdn?: string,
+  // Using string as it cna be any record type not just the ones we use, also AWS sdk refers to it as string
```
humphd, line 177: s/cna/can/

app/lib/dns.server.ts
```diff
@@ -165,6 +166,55 @@ async function checkDnsRecordExists(type: DnsRecordType, fqdn: string, value: st
...
+  try {
+    const command = new ListResourceRecordSetsCommand({
+      HostedZoneId: await getHostedZoneId(),
+      StartRecordName: fqdn,
+      StartRecordType: type,
+    });
+
+    return route53Client.send(command);
+  } catch (error) {
+    logger.warn('Error loading record page from Route53', { type, fqdn, error });
```
humphd, line 189: For consistency, prepend `DNS Error - `

app/lib/dns.server.ts
```diff
@@ -165,6 +166,55 @@ async function checkDnsRecordExists(type: DnsRecordType, fqdn: string, value: st
...
+    });
+    const response: ChangeResourceRecordSetsResponse = await route53Client.send(command);
+
+    if (!response.ChangeInfo?.Id) {
+      throw new Error(`DNS Error - missing ID in AWS ChangeInfo response`);
+    }
+
+    return response.ChangeInfo.Id;
+  } catch (error) {
+    logger.error('Failed to execute changeSet', { error });
```
humphd, line 213: Same here, `DNS Error - `

app/reconciler/ReconcilerTypes.ts
```diff
@@ -0,0 +1,8 @@
+import type { DnsRecordType } from '@prisma/client';
+
+export interface ReconcilerCompareStructure {
```
humphd, line 3: You don't want to mirror that same structure as AWS?  Not important that you do, but since we're creating a structure, we might consider using theirs.

app/reconciler/createChangeSetFromCompareStructures.ts
```diff
@@ -0,0 +1,92 @@
+import { isEqual } from 'lodash';
+import logger from '~/lib/logger.server';
+
+import { DnsRecordType } from '@prisma/client';
+
+import type { Change } from '@aws-sdk/client-route-53';
+import type { ReconcilerCompareStructure } from './ReconcilerTypes';
+
+interface InputType {
```
humphd, line 9: `CompareStructures`

app/reconciler/createChangeSetFromCompareStructures.ts
```diff
@@ -0,0 +1,92 @@
...
+  dbStructure: ReconcilerCompareStructure;
+  route53Structure: ReconcilerCompareStructure;
+}
+
+const createChangeSetFromCompareStructures = ({ dbStructure, route53Structure }: InputType) => {
+  const toChange: Change[] = [];
+
+  /**
```
humphd, line 17: Split each of these two steps out into their own functions so that it's easier to read.

app/reconciler/readRuote53IntoCompareStructure.server.ts
```diff
@@ -0,0 +1,70 @@
+import { set } from 'lodash';
```
humphd, line 1: Typo on filename, `Ruote`

app/reconciler/readRuote53IntoCompareStructure.server.ts
```diff
@@ -0,0 +1,70 @@
...
+    // and set that value on our compare structure
+    set(MUTATEDcompareStructure, [recordSet.Name, recordSet.Type], value);
+  });
+};
+
+/**
+ * This Fn reads the complete Route53 zone into a `ReconcilerCompareStructure` type object
+ * to be used for later comparison with our database data
+ */
+const readRuote53IntoCompareStructure = async (): Promise<ReconcilerCompareStructure> => {
```
humphd, line 46: Same typo here, `Ruote`

app/models/dns-record.server.ts
```diff
@@ -126,3 +126,13 @@ export function getExpiredDnsRecords() {
     },
   });
 }
+
+/**
+ * This Fn gets the base data that is needed for reconciliation
+ * but does that for all records in the db. Should be no more than 10k
+ */
```
humphd, line 133: It's worth adding a note here that this is our ground truth for all details about DNS Records.

app/models/dns-record.server.ts
```diff
@@ -126,3 +126,13 @@ export function getExpiredDnsRecords() {
...
+export function getReconciliationData() {
+  return prisma.dnsRecord.findMany({
+    select: { username: true, subdomain: true, type: true, value: true },
```
humphd, line 136: Any value in pre-sorting this at the db level?

app/reconciler/createChangeSetFromCompareStructures.ts
```diff
@@ -0,0 +1,92 @@
...
+      const route53Value = route53Structure[fqdn][type as DnsRecordType]!;
+      const dbValue = dbStructure[fqdn]?.[type as DnsRecordType];
+
+      if (dbValue) {
+        // Both of them have this
+        return;
+      }
+
+      toChange.push({
+        Action: 'DELETE',
```
humphd, line 32: To confirm, you don't need `MultiValueAnswer` on a `DELETE`?  I think not, but wanting to check.

app/reconciler/index.ts
```diff
@@ -0,0 +1,27 @@
+import { executeChangeSet } from '~/lib/dns.server';
+import logger from '~/lib/logger.server';
+import readDbIntoCompareStructure from './readDbIntoCompareStructure.server';
+import readRuote53IntoCompareStructure from './readRuote53IntoCompareStructure.server';
+import createChangeSetFromCompareStructures from './createChangeSetFromCompareStructures';
+
+const reconciler = async () => {
```
humphd, line 7: `reconcile`

app/reconciler/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  const dbStructure = await readDbIntoCompareStructure();
+  const route53Structure = await readRuote53IntoCompareStructure();
```
humphd, line 9: You might as well run these at the same time:

```ts
const [dbStructure, route53Structure] = Promise.all([readDbIntoCompareStructure(), readRuote53IntoCompareStructure()]);
```

app/reconciler/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  if (!changeSet.length) {
+    logger.debug('Reconciler found no changes to be pushed');
+    return;
+  }
+
+  logger.debug(`Reconciler intends to push the following ${changeSet.length} changes`, {
+    changeSet,
+  });
+
+  await executeChangeSet(changeSet.slice(0, 1000));
```
humphd, line 24: It might be nice to return some info about how many changes you just made, on how many total records, etc.  We could use this later to report back to the admin ui.

app/reconciler/readDbIntoCompareStructure.server.ts
```diff
@@ -0,0 +1,31 @@
...
+const readDbIntoCompareStructure = async (): Promise<ReconcilerCompareStructure> => {
+  const MUTATEDcompareStructure: ReconcilerCompareStructure = {};
+
+  const dnsData = await getReconciliationData();
+
+  // Populate compareStructure with data from the `Record` table
+  dnsData.forEach(({ username, subdomain, type, value }) => {
+    const fqdn = `${buildDomain(username, subdomain)}.`;
+
+    // Get the previous value for fqdn.recordType (or an empty array if missing)
```
humphd, line 20: This could use an example, since we don't store data as `fqdn.recordType` elsewhere, so readers will wonder what's happening if not reading this as a DIFF like I am now with full context.

app/reconciler/readRuote53IntoCompareStructure.server.ts
```diff
@@ -0,0 +1,70 @@
...
+
+// Using this in JS code later, cannot `import type`
+import { DnsRecordType } from '@prisma/client';
+import type { ReconcilerCompareStructure } from './ReconcilerTypes';
+import type { ResourceRecordSet, ListResourceRecordSetsResponse } from '@aws-sdk/client-route-53';
+
+/**
+ * !!! There is cross-function object mutation in this file.
+ * This is a huge anti-pattern, but in this case it made the code
+ * much more performant. So I marked the variable as such and
```
humphd, line 12: Your description for this issue should include details about this.  Do we care if it's slow, given that this is a background job?  What were your findings?

I think the way that you've marked it in both places is good.

app/reconciler/readRuote53IntoCompareStructure.server.ts
```diff
@@ -0,0 +1,70 @@
...
+  // !!! `MUTATEDcompareStructure` is passed by reference and is mutated !!!
+  MUTATEDcompareStructure: ReconcilerCompareStructure
+) => {
+  recordSets.forEach((recordSet) => {
+    // Unsure as to how those could be undefined, but according to AWS sdk, they could
+    if (!recordSet.Name || !recordSet.Type) {
+      return;
+    }
+
+    // We only care about record types that we handle. NS, CNAME, etc. should be ignored
```
humphd, line 27: We include `CNAME`, why have you listed it here with `NS`?

humphd (changes requested):
Thanks for getting to this so quickly.

This is going to need a bunch of tests in follow-ups.  cc @cychu42, @Eakam1007, @Genne23v, @sfrunza13, @SerpentBytes who can probably all lend a hand with reviews and writing tests in the future.

app/reconciler/ReconcilerTypes.ts
dadolhay: That was my original plan, but those structures are hard to compare. I'd have to do a lot of very expensive find operations, that would scale execution time esponentially O(n^2) based on the recordSet count. We would not see the issue right now, but years down the line, with many users, it sould probably cause problems.

For this structure, I'm basically using the keys as indexes, so this should scale in linear manner O(n)

app/models/dns-record.server.ts
dadolhay: I don't think so, it should be rare that we handle multiple values for a single domain-type pair, and I'd not 100% trust that the sort result is the exact same between js sort() and mysql order by.

dadolhay: in other cases sorting is not meaningful, we'll generate our internal structure

app/reconciler/createChangeSetFromCompareStructures.ts
dadolhay: No, during testing it worked perfectly without it

app/reconciler/index.ts
dadolhay: Good point

app/reconciler/readRuote53IntoCompareStructure.server.ts
dadolhay: Typo, I meant SOA

app/reconciler/createChangeSetFromCompareStructures.ts
dadolhay: Actually, you were right, it DID need it. The documentation does not mention it, upsert works without it, the returned error is complete nonsense (misses attributes that I do actually send), but sending MultiValue for the delete solved it. 

Nice catch, saved me a lot of debug time!

dadolhay:
> Thanks for getting to this so quickly.
> 
> This is going to need a bunch of tests in follow-ups. cc @cychu42, @Eakam1007, @Genne23v, @sfrunza13, @SerpentBytes who can probably all lend a hand with reviews and writing tests in the future.

For sure, I know how important this is for the project and the team.

Actually, if this passes reviews, we could completely drop the queue based DNS system, and tie this single call in after any `Record` table change for now.

I don't think that we have too much traffic on staging, so I'm not too afraid for race conditions until I'm able to finish the scheduler for this.

In the meantime, this seems to work, and insert / update / delete recordSets correctly in Route53 ==> the cert generation might just work after we migrate to this (would also immediately kill all drift between our db and Route53 on staging)

app/reconciler/readDbIntoCompareStructure.server.ts
```diff
@@ -0,0 +1,40 @@
...
+import { buildDomain } from '../utils';
+
+import type { ReconcilerCompareStructure } from './ReconcilerTypes';
+
+/**
+ * This Fn reads the database `Record` table into a `ReconcilerCompareStructure` type object
+ * to be used for later comparison with Route53
+ */
+const readDbIntoCompareStructure = async (): Promise<ReconcilerCompareStructure> => {
+  const MUTATEDcompareStructure: ReconcilerCompareStructure = {};
```
humphd, line 12: Thinking about this a bit more, and I wonder if we should create a `Reconciler` class, with these structures as properties, and the functions operating on the instance data.  The whole thing would be:

```ts
const reconciler = new Reconciler();
// Could also fold this init() call into reconcile()`
await reconciler.init();
await reconciler.reconcile();
```

Now we don't need to deal with mutations across functions, and can modify things in the instance, which isn't going to be odd to future maintainers.

humphd (commented):
I think this is looking good.  I have one question about whether using an instance vs. module functions would improve maintainability, but otherwise, I think we should get others to review this and land it so we an start testing and building the code around it.

humphd:
#438 deals with the database side of using multiple values for a given record name+type.

app/reconciler/readDbIntoCompareStructure.server.ts
dadolhay: This is an interesting idea. We would not eliminate the issue of mutating the object (I like to treat everything as immutable) but at least it would wrap it.

Making the complete Reconciler a class would create too much & complex code in a single class, but we could make the [readRoute53IntoCompareStructure.server.ts](https://github.com/DevelopingSpace/starchart/pull/441/files#diff-305e5602fea0950a79e73510dd6f343bfa48215e1004b1af0ed28a38df45fd8f) into one (that's the only one that's problemmatic). Maybe the DB one too for consistency

humphd: Agreed, keep the reconciler spread out across files, but combine the compare structures into a instance that we can pass around and modify as needed.

app/reconciler/createChangeSetFromCompareStructures.ts
```diff
@@ -0,0 +1,103 @@
...
+  dbStructure,
+  route53Structure,
+}: CompareStructures) => {
+  const toChange: Change[] = [];
+
+  /**
+   * If something is present in route53 but is completely missing in db,
+   * then we have to delete it
+   */
+  Object.keys(route53Structure).forEach((fqdn) => {
```
humphd, line 24: I tried asking ChatGPT to improve readability of this code, interesting to see what it did:

```ts
import { isEqual } from 'lodash';

import { DnsRecordType } from '@prisma/client';

import type { Change } from '@aws-sdk/client-route-53';
import type { ReconcilerCompareStructure } from './ReconcilerTypes';

interface CompareStructures {
  dbStructure: ReconcilerCompareStructure;
  route53Structure: ReconcilerCompareStructure;
}

export function createDeleteChangeSet({ dbStructure, route53Structure }: CompareStructures) {
  const toChange: Change[] = [];

  for (const [fqdn, route53Types] of Object.entries(route53Structure)) {
    for (const [type, route53Value] of Object.entries(route53Types)) {
      const dbValue = dbStructure[fqdn]?.[type as DnsRecordType];

      if (!dbValue) {
        toChange.push({
          Action: 'DELETE',
          ResourceRecordSet: {
            MultiValueAnswer: route53Value.length > 1,
            Name: fqdn,
            Type: type,
            ResourceRecords: route53Value.map((Value) => ({ Value })),
            TTL: 60 * 5,
          },
        });
      }
    }
  }

  return toChange;
}

export function createUpsertChangeSet({ dbStructure, route53Structure }: CompareStructures) {
  const toChange: Change[] = [];

  for (const [fqdn, dbTypes] of Object.entries(dbStructure)) {
    for (const [type, dbValue] of Object.entries(dbTypes)) {
      const route53Value = route53Structure[fqdn]?.[type as DnsRecordType];

      if (!route53Value || !isEqual(dbValue?.sort(), route53Value?.sort())) {
        const isMultivalue =
          dbValue.length > 1 &&
          [DnsRecordType.A, DnsRecordType.AAAA, DnsRecordType.TXT].includes(type as DnsRecordType);

        if (!isMultivalue) {
          logger.error(
            'Error creating DNS changeset: only A, AAAA, and TXT records can be multivalue. Ignoring records.',
            { fqdn, type }
          );
          continue;
        }

        toChange.push({
          Action: 'UPSERT',
          ResourceRecordSet: {
            MultiValueAnswer: isMultivalue,
            Name: fqdn,
            Type: type,
            ResourceRecords: dbValue.map((Value) => ({ Value })),
            TTL: 60 * 5,
          },
        });
      }
    }
  }

  return toChange;
}
```

app/lib/dns.server.ts
```diff
@@ -165,6 +166,55 @@ async function checkDnsRecordExists(type: DnsRecordType, fqdn: string, value: st
...
+
+    // Rethrow to keep stack
+    throw error;
+  }
+};
+
+/**
+ * Execute a complete changeset in Route53 generated by the reconciler
+ */
+export const executeChangeSet = async (Changes: Change[]) => {
```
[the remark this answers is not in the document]
humphd: The AWS SDK is generated from a spec automatically, and uses capitals where JS/TS normally uses lowercase in names.  This is due to the fact that the spec is language agnostic.  He's using it to push into the SDK `ChangeBatch`, which wants `Changes` vs. `changes` so it just makes the code slightly shorter to not have to do `Changes: changes`.  But your point is normally correct.  AWS is just wrong here :)

app/lib/dns.server.ts
```diff
@@ -165,6 +166,55 @@ async function checkDnsRecordExists(type: DnsRecordType, fqdn: string, value: st
   }
 }
 
+/**
+ * Get a full page of records from AWS Route53. If fqdn and recordType is specified,
+ * load the next page starting with that record
+ *
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-route-53/classes/listresourcerecordsetscommand.html
+ */
+export const getRecordPage = async (
```
SerpentBytes, line 175: Should this be `getDnsRecordPage`?

app/reconciler/createChangeSetFromCompareStructures.ts
dadolhay: Well, it wrote code, but it's not the same functionality, like this is completely incorrect:
```js
const isMultivalue =
          dbValue.length > 1 &&
          [DnsRecordType.A, DnsRecordType.AAAA, DnsRecordType.TXT].includes(type as DnsRecordType);

        if (!isMultivalue) {
          logger.error(
            'Error creating DNS changeset: only A, AAAA, and TXT records can be multivalue. Ignoring records.',
            { fqdn, type }
          );
          continue;
        }
```

humphd: Totally, I was just interested in its style.  Not suggesting you use it.  Fascinating to see what it can and can't do.

app/lib/dns.server.ts
dadolhay: actually `getDnsRecordSetPage` now that you mention it :)

humphd (approved):
Looks sane to me, but without tests I have no idea.  Let's try it and build around and on it.

Eakam1007 (approved):

SerpentBytes (approved):

dadolhay:
Thanks everyone for the reviews and suggestions. Your ideas helped a lot!

PR #447, opened 2023-03-26
feat: Dns reconciler queue

### LINKED ISSUE #396: DNS reconciler

dadolhay:
- Moves the complete reconciler to under queues
- Creates a queue for it with automatic repeat. It makes certain that only 1 repeat job can exist
- Adds a new table, **Uses a unique enum** so only ever 1 row can exist in the table
- Alters the reconciler main code, so that it only runs if there is no concurrency problem (because of multiple instances in the swarm), and it is requested to run
- Introduces 'limp mode'. If bulk update fails, it will try to run it one by one, and it points the offending update out in the logs

Closes #396


prisma/schema.prisma
```diff
@@ -59,6 +59,18 @@ model Challenge {
   certificate   Certificate @relation(fields: [certificateId], references: [id], onDelete: Cascade)
 }
 
+model SystemState {
+  // As the following type is unique, enum, requored and only one option
+  // it enforces that there is only one column in the table
```
humphd, line 64: Do you mean `only one row in the table`?  We could add other global state here (i.e., columns) if we needed to, no? Or do I misunderstand?

prisma/schema.prisma
```diff
@@ -59,6 +59,18 @@ model Challenge {
...
+  unique               StateEnumType @unique @default(unique)
+  reconciliationNeeded Boolean       @default(false)
+}
+
+// This is to force only a single column in the State table
```
humphd, line 69: Same question here: row vs. column?

prisma/seed.ts
```diff
@@ -16,11 +16,19 @@ async function seed() {
...
+  await prisma.systemState.deleteMany().catch(() => {});
   await prisma.dnsRecord.deleteMany().catch(() => {});
   await prisma.challenge.deleteMany().catch(() => {});
   await prisma.certificate.deleteMany().catch(() => {});
   await prisma.user.deleteMany().catch(() => {});
 
+  await prisma.systemState.create({
+    data: {
+      unique: 'unique',
+      reconciliationNeeded: false,
```
humphd, line 28: We could actually seed some DNS Records for users now, since they will get reflected in Route53 if we set this to true. That can happen in a follow up vs. here, just thinking out loud.

app/models/system-state.server.ts
```diff
@@ -0,0 +1,23 @@
+import { StateEnumType } from '@prisma/client';
+
+import { prisma } from '~/db.server';
+
+import type { SystemState } from '@prisma/client';
+
```
humphd, line 6: I think this file could use a comment.  Something like:

```ts
/**
 * The SystemState represents global (i.e., cross-instance), shared settings among Starchart instance(s).
 * Initially, this includes info on whether or not the DNS Reconciler needs to be run at the next opportunity.
 * We use this global SystemState to make sure that we never have more than 1 instance of the reconciler
 * running at a time.
 */
```

Fix that to be be what makes sense to you, but let's try to explain it to future readers.

app/models/system-state.server.ts
```diff
@@ -0,0 +1,23 @@
...
+export function getIsReconciliationNeeded(): Promise<SystemState['reconciliationNeeded']> {
+  return prisma.systemState
+    .findUnique({
+      select: { reconciliationNeeded: true },
+      where: { unique: StateEnumType.unique },
+    })
+    .then((data) => data!.reconciliationNeeded);
```
humphd, line 13: The use of `data!.` here makes me a bit nervous here, since it will bring down the system pretty quick if it fails at runtime.  I get that it shouldn't fail, but...

Given that this *could* always return `undefined`, could we wrap this access in a `try/catch` and return `true` if `data` is missing at runtime?  Doing so would force the reconciler to call `setIsReconciliationNeeded` soon and fix the situation so that the record does get added.

app/queues/reconciler/reconciler-queue.server.ts
```diff
@@ -0,0 +1,44 @@
...
+  logger.info('Starting DNS reconciler queue');
+
+  const jobName = `reconciler-scheduler`;
+
+  try {
+    // Remove all previously existing repeatable jobs
+    // This is important because multiple repeatable jobs can exist and they persist
+    // within redis (even with the same key)
+    const repeatableJobs = await reconcilerQueue.getRepeatableJobs();
+    Promise.all(repeatableJobs.map(({ key }) => reconcilerQueue.removeRepeatableByKey(key)));
```
humphd, line 31: I would `await` this so you don't race on 33 when adding a new one.

app/queues/reconciler/reconciler-queue.server.ts
```diff
@@ -0,0 +1,44 @@
...
+  var __reconciler_queue_init__: boolean;
+}
+
+reconcilerQueue.on('error', (err) => {
+  logger.error('Reconciler encountered an error', err);
+});
+
+// function to add jobs
+export const addReconcilerJob = async () => {
+  if (process.env.NODE_ENV !== 'production') {
```
humphd, line 14: Can you explain this to me?  If we're in development, you're not going to run this?

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,127 @@
...
+};
+
+/**
+ * NORMAL MODE
+ *
+ * Execute the complete changeSet at once
+ */
+const pushChangesBulk = (changeSet: Change[]): boolean => {
+  logger.debug(
+    `NORMAL MODE - Reconciler intends to push the following ${Math.min(
```
humphd, line 44: Prepend `Reconciler` to this: `Reconciler NORMAL MODE` so it's easier to grok this in the logs. Also Can you move this `Math.min(...)` to a variable above this call so this read better vs. breaking across 3 lines?

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,127 @@
...
+  // Return boolean => Is additional reconciliation needed
+  return changeSet.length > 1000;
+};
+
+/**
+ * LIMP MODE
+ *
+ * Try each change in the set one by one, isolate the offending one
+ */
+const pushChangesLimp = async (changeSet: Change[]) => {
```
humphd, line 63: `pushChangesIndividually`

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+  );
+  executeChangeSet(changeSet.slice(0, CHANGE_SET_MAX_SIZE));
+
+  // Return boolean => Is additional reconciliation needed
+  return changeSet.length > 1000;
+};
+
+/**
+ * LIMP MODE
```
humphd, line 59: This is really clever, well done.

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * Try each change in the set one by one, isolate the offending one
+ */
+const pushChangesLimp = async (changeSet: Change[]) => {
+  for (const change of changeSet) {
+    try {
+      logger.debug(`LIMP MODE - Reconciler intends to push the following change`, {
```
humphd, line 66: `Reconciler LIMP MODE - ...`

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,127 @@
...
+  logger.debug(
+    `NORMAL MODE - Reconciler intends to push the following ${Math.min(
+      1000,
+      changeSet.length
+    )} changes`,
+    {
+      changeSet,
+    }
+  );
+  executeChangeSet(changeSet.slice(0, CHANGE_SET_MAX_SIZE));
```
humphd, line 52: You need to `await` this, and given the code below, you should probably `try/catch` here too.

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+  }
+};
+
+const reconcilerWorker = new Worker(
+  reconcilerQueueName,
+  async () => {
+    // When BullMQ is running on multiple nodes (Docker swarm), we need to make sure our concurrency is 1
+    const activeJobs = (await reconcilerQueue.getJobs('active'))
+      // BullMQ bug, sometimes I get an array element with `undefined`, that should not be possible
```
humphd, line 85: Worth filing upstream?

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,127 @@
...
+      .filter((v) => !!v);
+
+    if (activeJobs.length > 1) {
+      logger.debug('Reconciler - Inhibiting concurrent run');
+      return;
+    }
+
+    // Only run worker if needed
+    if (!(await getIsReconciliationNeeded())) {
+      logger.debug('Reconciler - Run not requested');
```
humphd, line 95: `Reconciler - skipping current job, reconciler not needed.`

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,127 @@
...
+    const activeJobs = (await reconcilerQueue.getJobs('active'))
+      // BullMQ bug, sometimes I get an array element with `undefined`, that should not be possible
+      .filter((v) => !!v);
+
+    if (activeJobs.length > 1) {
+      logger.debug('Reconciler - Inhibiting concurrent run');
+      return;
+    }
+
+    // Only run worker if needed
```
humphd, line 93: `// Only run if reconciler was explicitly requested`

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    if (!changeSet.length) {
+      logger.debug('Reconciler - found no changes to be pushed');
+      await setIsReconciliationNeeded(false);
+      return;
+    }
+
+    let isAdditionalReconciliationNeeded = true;
+    try {
+      isAdditionalReconciliationNeeded = await pushChangesBulk(changeSet);
```
humphd, line 109: Add comments for normal vs. limp mode calls here, explaining the logic in brief detail.

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,127 @@
...
+    } catch (error) {
+      logger.error('Reconciler - Change set failed, switching to limp mode', { error });
+
+      await pushChangesLimp(changeSet);
+    }
+
+    /**
+     * Update system state
+     *
+     * If  changeSet is < 1000 elements, then dns data that has been altered
```
humphd, line 119: extra whitespace in front of `changeSet`.

Leave a note about the significance of 1000 (i.e., AWS changeset limit, etc).

app/root.tsx
```diff
@@ -36,6 +37,8 @@ export async function loader({ request, context }: LoaderArgs) {
   });
 }
 
+addReconcilerJob();
```
humphd, line 40: This seems like the wrong place to do this.  I get that Remix is annoying with this stuff, and I've struggled a lot to find an optimal solution.

It feels like something that could live in `server.ts`, despite the fact that I ripped all that code out of there.  I'm not sure what to do.  If you want to leave this here for now, and we fix it later, that's also doable.  I don't want to block this work on something stupid.

For follow-up, I want to figure out how to move the bulk of the queue stuff out of the `app/` and have it be associated with the server anyway.

humphd (changes requested):
This is great.  I love normal vs. limp mode! Lots of great problem solving and interesting patterns in here.

I've left a bunch of questions and comments.

prisma/schema.prisma
dadolhay: Yes, you are correct!

prisma/seed.ts
dadolhay: Correct, reconciler will automatically push the (seed) data to route53

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,127 @@
...
+     * have now been reconciled
+     */
+    await setIsReconciliationNeeded(isAdditionalReconciliationNeeded);
```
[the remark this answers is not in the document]
dadolhay: It theoretically could, but, MySQL would need to go down for that, and this is not critical information (also the last thing that we do, if it fails, not really an issue). If it fails, 2 minutes later it will do an other reconciliation, will see that all data is pushed, and finish without doing anything

app/root.tsx
dadolhay: Yes, this gets called multiple times during development, I simply could not find a better place (and not for the lack of trying!) Also all other queues are "implicitly" initialized, just because of the file import. Since This piece does not get called from the UI or anywhere else, and if I only don an `import 'filename'` it does not work at all (code in the file is not executed)

I agree, we should do a follow-up

app/models/system-state.server.ts
dadolhay: Actually cross-instance concurrency is inhibited in the worker, when I'm checking the number of running workers in the queue. This one is only to not keep polling Route53 every 2 minutes if we know nothing has changed

```
 * We use this global SystemState to make sure that we stop hitting the Route53 API when no changes were made
 * to our Records table`
```

dadolhay:
@humphd Thank you! Updates pushed, and I'll ask for some help with the e2e tests.
My only feeling is that it might be failing because of the updated database schema, but the video shows it standing on the main screen (before login) not doing anything.

No rush, we can talk about it tomorrow! Again, thanks for the very quick reviews!

app/models/system-state.server.ts
dadolhay: Thinking about this, you might just solved my e2e testing issue

app/queues/reconciler/reconciler-queue.server.ts
dadolhay: yep, that's a bug

app/queues/reconciler/reconciler-queue.server.ts
dadolhay: There can only ever be 1, but you are right, that's cleaner

app/queues/reconciler/reconciler-worker.server.ts
dadolhay: Await => yes, you are correct
try/catch => the error is handled in the worker. If this fails, it should take the pushChangesBulk fn with it

app/queues/reconciler/reconciler-worker.server.ts
dadolhay: Thank you!

app/queues/reconciler/reconciler-worker.server.ts
dadolhay: Thank You!

app/queues/reconciler/reconciler-worker.server.ts
dadolhay: Then I'd have to create a reproduction for them, which would escalate to adding a docker-compose for bringing up a redis ... I simply have no capacity RN. If it keeps bothering us, let's re-visit!

app/queues/reconciler/reconciler-queue.server.ts
```diff
@@ -28,7 +27,9 @@ export const addReconcilerJob = async () => {
     // This is important because multiple repeatable jobs can exist and they persist
     // within redis (even with the same key)
     const repeatableJobs = await reconcilerQueue.getRepeatableJobs();
-    Promise.all(repeatableJobs.map(({ key }) => reconcilerQueue.removeRepeatableByKey(key)));
+    repeatableJobs.forEach(async ({ key }) => {
```
humphd, line 30: I'd still use `await Promise.all()` and `.map` here, but if you want a loop, use `for-of` you're going to `await` inside.

humphd (changes requested):

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,132 @@
...
+
+const reconcilerWorker = new Worker(
+  reconcilerQueueName,
+  async () => {
+    // When BullMQ is running on multiple nodes (Docker swarm), we need to make sure our concurrency is 1
+    const activeJobs = (await reconcilerQueue.getJobs('active'))
+      // BullMQ bug, sometimes I get an array element with `undefined`, that should not be possible
+      .filter((v) => !!v);
+
+    if (activeJobs.length > 1) {
```
humphd, line 88: I missed the explanation on this (or it's been lost in GitHub with the updated commit).  Why `> 1`?

app/root.tsx
humphd, line 40: I'll file an issue on this for later.  I don't like this pattern, and yet I have no good ideas to offer instead.

humphd (changes requested):

Genne23v (changes requested):
The code is well organized and comments are easy to understand. I added only one comment. Feel free to update or keep it as is.

app/root.tsx
humphd: Filed #450

app/queues/reconciler/reconciler-queue.server.ts
dadolhay: You are right, Updating back, adding the missing await

app/queues/reconciler/reconciler-worker.server.ts
dadolhay: Normally, when a BullMQ worker is added, it will behave as single threaded ... will only execute one job at a time.
But, if we add multiple instances, i.e. our docker swarm, we have one worker per instance, taking jobs from the queue.
---
When you add a repeat job to a queue, it is a special thing. What it causes the queue system to keep adding (delayed) jobs when the repeat pattern/integer dictates it.
---
In theory, it would be possible, that a job is running long (more than the 2 mins job repeat time), so the second swarm node would also pick up the next scheduled job, causing the system to run two reconcilers in parallel.
---
For this reason, I'm asking the BullMQ system ... tell me how many active jobs are there (this includes the current job too, that we are in right now). If the answer is > 1, it means that there was a pre-existing job already running, when we were started ==> we must exit to inhibit concurrency

dadolhay: I guess, I should copy-paste the above into the code :)

app/root.tsx
dadolhay: I agree, thank you!

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  return changeSet;
+};
+
+/**
+ * NORMAL MODE
+ *
+ * Execute the complete changeSet at once
+ */
+const pushChangesBulk = async (changeSet: Change[]): Promise<boolean> => {
```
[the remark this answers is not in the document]
dadolhay: I'd normally agree, if the main purpose of the function would be to
return something. However, in this case the main purpose is to execute
the changeset, and it only returns the result of that main operation. 

For this reason I'd not reflect the retval in the fn name

app/queues/reconciler/reconciler-queue.server.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+reconcilerQueue.on('error', (err) => {
+  logger.error('Reconciler encountered an error', err);
+});
+
+// function to add jobs
+export const addReconcilerJob = async () => {
+  if (process.env.NODE_ENV !== 'production' && global.__reconciler_queue_init__) {
+    // Only do this setup once if in dev
```
Eakam1007, line 15: Out of curiosity, does this actually run multiple times in production? Why?

Eakam1007 (approved):

app/queues/reconciler/reconciler-queue.server.ts
dadolhay: Nope, on prod it should only run once every startup
If you check that condition, (an early return) it only returns if not priduction and already initialized

Eakam1007: Oh, so it could run multiple times in dev, so this prevents that. I suppose that happens due to live reload?

dadolhay:
@humphd When you are around, can I get an approve if all is well?

app/queues/reconciler/reconciler-worker.server.ts
```diff
@@ -0,0 +1,154 @@
...
+      });
+    }
+  }
+};
+
+const reconcilerWorker = new Worker(
+  reconcilerQueueName,
+  async () => {
+    /**
+     * When a BullMQ worker is added, it will behave as single
```
humphd, line 84: This is a great discussion of what's happening here, thanks for adding it.

humphd (approved):
🚢 ship it!

Genne23v (approved):

app/queues/reconciler/reconciler-worker.server.ts
dadolhay: Absolutely!

PR #449, opened 2023-03-26
Add bad word filter

### LINKED ISSUE #28: Filter bad words in domain names

Genne23v:
Close #28 

This PR replaces bad words with `x` in creating/updating new DNS record.


SerpentBytes (dismissed):

app/routes/__index/dns-records/$dnsRecordId.tsx
```diff
@@ -61,7 +64,7 @@ export const action = async ({ request }: ActionArgs) => {
     id: Number(data.id),
     username: user.username,
     type: data.type,
-    subdomain: data.subdomain,
+    subdomain: filter.clean(data.subdomain),
```
Eakam1007, line 67: Do you think it would be better to show an error on the UI if these words exist in the subdomain instead of saving the record with them filtered out?

humphd (changes requested):
Let's add this as a validation step, and fail the record if any of these are included with a message indicating they need to pick a different name.

app/routes/__index/dns-records/$dnsRecordId.tsx
Genne23v: Thanks for the comment! I changed it to decline inappropriate subdomain with proper message. Please review

app/routes/__index/dns-records/$dnsRecordId.tsx
```diff
@@ -42,7 +42,8 @@ export const action = async ({ request }: ActionArgs) => {
       return isNameValid(fqdn, user.username);
     },
     {
-      message: 'Record name is invalid',
+      message:
+        'Record name is invalid or inappropriate (only alphanumerical character, -, and _ allowed)',
```
Ririio, line 46: Is it possible to add on the message the specific value that is considered inappropriate?

Genne23v: Currently it uses backend function that returns `boolean`. To show specific error message, it takes much bigger scope. I will write another issue to follow up if that is acceptable.

app/lib/dns.server.ts
```diff
@@ -378,6 +381,11 @@ export const isNameValid = (fqdn: string, username: string) => {
   }
   const subdomain = fqdn.substring(0, fqdn.length - toRemove.length);
 
+  // Decline if subdomain has inappropriate word(s)
+  if (filter.isProfane(subdomain)) {
```
humphd, line 385: To confirm, this won't check the username portion, right?  We don't want to fail something that is someone's name by accident.

This is only the `foo` in `foo.username.root.domain` right?

app/routes/__index/dns-records/$dnsRecordId.tsx
humphd: It might be better to be vague.

app/lib/dns.server.ts
Genne23v: Yes, only `subdomain` will be checked

humphd (dismissed):

app/routes/__index/dns-records/$dnsRecordId.tsx
Genne23v: I was thinking some users try to use special characters, but they don't know why it's invalid. So I tried to at least give what's allowed. Should it be just `Record name is invalid or inappropriate`?

Ririio: We can provide the information on the tooltip no?

SerpentBytes (dismissed):

humphd (approved):

Ririio (approved):

PR #456, opened 2023-03-27
Replacing Varchar to Text for 'privateKey' and 'certificate

Ririio:
Updates the current type for `privateKey` and `certificate` from `Varchar` -> `Text` to accomodate for a longer string


cychu42 (approved):
LGTM. Nice catch.

sfrunza13 (approved):

PR #465, opened 2023-03-28
Fix reconciler TXT record handling

### LINKED ISSUE #462: TXT record values must be wrapped in "...", split at 255 chars, etc.
See https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/ResourceRecordTypes.html#TXTFormat

We fought with Route53 all night tonight to get TXT records to work.  It turns out we need to:

1. wrap the value in extra "..."
2. break it at the 255 character mark
3. not exceed 4,000 characters

[edit]
- break it at the 255 character mark
- escape backslash and quotation mark characters \ => \\  and " => \"
- escape not \o40 - \o177 ascii characters in 3 digit octal format " " => "\040"
- wrap the chunks in extra "..."
- join them with a space

dadolhay:
Closes #462

### DIFF
--- a/app/queues/reconciler/route53Utils.server.ts
+++ b/app/queues/reconciler/route53Utils.server.ts
@@ -52,7 +52,7 @@ export const toRoute53RecordValue = (type: DnsRecordType, value: string): string
     .map((segment, index) => value.substring(index * 255, (index + 1) * 255))
     // escape
     .map((segment) => segment.replace(/([^!-~]|[\\"])/g, escapeFn))
-    // add parenthesis around the segments
+    // add quotation marks around the segments
     .map((segment) => `"${segment}"`);
 
   // Finally join the segments together with a white space
--- a/app/queues/reconciler/route53Utils.server.ts
+++ b/app/queues/reconciler/route53Utils.server.ts
@@ -10,6 +10,35 @@ import { DnsRecordType } from '@prisma/client';
  * Note, TXT records should not include quotation marks
  */
 
+/**
+ * in each segment, replace characters that are not between \o040 - \o176
+ * use a replace fn, to generate our 3 digit octal codes, i.e., '!' => '\041'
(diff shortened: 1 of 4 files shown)


humphd (changes requested):
Please add a few unit tests for these 2 utility functions so we don't waste time debugging this on staging.

dadolhay:
@humphd Added!

app/queues/reconciler/route53Utils.server.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  // Create an uninitialized array with the length to hold our split up strings (max 255 chars)
+  const segments = new Array(Math.ceil(value.length / 255))
+    // Initialize with undefined
+    .fill(undefined)
+    // Loop through, using the index split out the appropriate parts from the original string
+    .map((segment, index) => value.substring(index * 255, (index + 1) * 255))
+    // escape
+    .map((segment) => segment.replace(/([^!-~]|[\\"])/g, escapeFn))
+    // add parenthesis around the segments
```
humphd, line 55: parenthesis?  Do you mean "double-quotes"?

humphd (dismissed):
Excellent, thank you.  The only thing this doesn't do is the 4,000 limit:

> The maximum length of a value in a TXT record is 4,000 characters.

However, I'm OK if we leave that out.

dadolhay:
> Excellent, thank you. The only thing this doesn't do is the 4,000 limit:
> 
> > The maximum length of a value in a TXT record is 4,000 characters.
> 
> However, I'm OK if we leave that out.

Actually, I don't think our db column holds that amount of info. Even if it did, we should inhibit the insert IMO

app/queues/reconciler/route53Utils.server.ts
dadolhay: yep! thanks!

humphd (approved):

Eakam1007 (approved):

PR #466, opened 2023-03-28
First attempt at effective name for admin mock

### LINKED ISSUE #306: Admin user impersonating
Effective username is populated for users, empty for admin so they are redirected to admin page and they can choose to select a user to impersonate.

Effective username is property on session, for admin null, for others populated with their username.

Create shell page for admin to redirect to for admin accounts.
RequireAdminUser, some form of route guard.

Two functions, fill effective user and clear effective user (redirect to root dashboards). Both only for admin group.

Do not allow this for deactivated users.

Update useUser to use Effective User. Rename to useEffectiveUser.

Username and Effective Username should be logged by Winston if present.

sfrunza13:
Fixes #306 or trying to anyways.

I found that I could not set multiple key value pairs on session to commit it so I made a new cookie instead, curious to see what people think.

### DIFF
--- a/app/models/user.server.ts
+++ b/app/models/user.server.ts
@@ -8,6 +8,7 @@ export interface User extends PrismaUser {
   // the base domain to use for all DNS records created by this user
   // (e.g., jsmith.starchart.com with records using *.jsmith.starchart.com)
   baseDomain: string;
+  isAdmin: boolean;
 }
 
 export async function getUserByUsername(username: PrismaUser['username']) {
@@ -17,7 +18,11 @@ export async function getUserByUsername(username: PrismaUser['username']) {
   }
 
   // Decorate with the user's base domain as well
-  return { ...user, baseDomain: buildUserBaseDomain(username) };
+  return {
+    ...user,
+    baseDomain: buildUserBaseDomain(username),
+    isAdmin: await isAdmin(username),
+  };
 }
 
 /**
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -10,7 +10,8 @@ import {
   useLoaderData,
 } from '@remix-run/react';
 
-import { getUser } from './session.server';
+import { getUser, getEffectiveUser } from './session.server';
+
 import theme from './theme';
 
 import type { MetaFunction, LoaderArgs, LinksFunction } from '@remix-run/node';
@@ -32,6 +33,14 @@ export const links: LinksFunction = () => [
 export async function loader({ request, context }: LoaderArgs) {
   return json({
     user: await getUser(request),
+    /**
+     * the effectiveUsername, this is for admin impersonating users. Regular
+     * users will have an effectiveUsername that is always the same as their
+     * username, admins (identified by group) will start with an empty string
+     * that can be changed to another users username on the admin portal for
+     * them to impersonate
+     **/
+    effectiveUser: await getEffectiveUser(request),
     nonce: context.nonce as string,
   });
 }
--- a/app/session.server.ts
+++ b/app/session.server.ts
@@ -1,9 +1,11 @@
 import { createCookie, createCookieSessionStorage, redirect } from '@remix-run/node';
 
-import type { User } from '~/models/user.server';
+import { isAdmin } from '~/models/user.server';
 import { isUserDeactivated } from '~/models/user.server';
 import { getUserByUsername } from '~/models/user.server';
 import secrets from '~/lib/secrets.server';
+
+import type { User } from '~/models/user.server';
 
 if (!secrets.SESSION_SECRET?.length) {
   throw new Error('SESSION_SECRET must be set');
@@ -43,9 +45,56 @@ export async function getUsername(request: Request): Promise<User['username'] | 
   return username;
 }
 
+export async function getEffectiveUsername(
+  request: Request
+): Promise<User['username'] | undefined> {
+  const cookie = request.headers.get('Cookie');
+  const effectiveUsername = await effectiveUsernameCookie.parse(cookie);
+  return effectiveUsername as User['username'];
+}
+
+export async function getEffectiveUser(request: Request) {
+  // get the effective username in the cookie at the time
+  const [username, effectiveUsername] = await Promise.all([
+    getUsername(request),
+    getEffectiveUsername(request),
+  ]);
+
+  if (!username) {
+    return null;
+  }
+
+  // if the user doesn't have an effective username,
+  // we're using their normal username (not impersonating)
+  let user;
+  if (!effectiveUsername) {
+    user = await getUserByUsername(username);
+  } else {
+    user = await getUserByUsername(effectiveUsername);
+  }
+
+  if (user) {
+    return user;
+  }
+
+  throw await logout(request);
+}
+
+export async function setEffectiveUsername(
+  username: string,
+  name: string | undefined
+): Promise<string> {
+  if (await isAdmin(username)) {
+    return await effectiveUsernameCookie.serialize(name);
+  }
+
+  return await effectiveUsernameCookie.serialize(username);
+}
+
 export async function getUser(request: Request) {
   const username = await getUsername(request);
-  if (username === undefined) {
+
+  if (!username) {
     return null;
   }
 
@@ -61,13 +110,17 @@ export async function requireUsername(
   request: Request,
   redirectTo: string = new URL(request.url).pathname
 ) {
-  const username = await getUsername(request);
+  const [username, effectiveUsername] = await Promise.all([
+    getUsername(request),
+    getEffectiveUsername(request),
+  ]);
+
   if (!username) {
     const searchParams = new URLSearchParams([['redirectTo', redirectTo]]);
     throw redirect(`/login?${searchParams}`);
   }
 
-  return username;
+  return effectiveUsername || username;
 }
 
 export async function requireUser(request: Request) {
@@ -79,6 +132,17 @@ export async function requireUser(request: Request) {
   }
 
   throw await logout(request);
+}
+
+export async function requireAdmin(request: Request) {
+  const username = await requireUsername(request);
+
+  const user = await getUserByUsername(username);
+
+  if (user?.isAdmin) {
+    return user;
+  }
+  throw redirect('/');
 }
 
 export async function createUserSession({
@@ -94,14 +158,22 @@ export async function createUserSession({
 }) {
   const session = await getSession(request);
   session.set(USER_SESSION_KEY, username);
+
+  const headers = new Headers();
+
+  headers.append('Set-Cookie', await setEffectiveUsername(username, undefined));
+
+  headers.append(
+    'Set-Cookie',
+    await sessionStorage.commitSession(session, {
+      maxAge: remember
+        ? 60 * 60 * 24 * 7 // 7 days
+        : undefined,
+    })
+  );
+
   return redirect(redirectTo, {
-    headers: {
-      'Set-Cookie': await sessionStorage.commitSession(session, {
-        maxAge: remember
-          ? 60 * 60 * 24 * 7 // 7 days
-          : undefined,
-      }),
-    },
+    headers,
   });
 }
 
@@ -111,6 +183,7 @@ export async function logout(request: Request, redirectTo?: string) {
   const headers = new Headers();
   headers.append('Set-Cookie', await sloUsernameCookie.serialize(session.get(USER_SESSION_KEY)));
   headers.append('Set-Cookie', await sessionStorage.destroySession(session));
+  headers.append('Set-Cookie', await effectiveUsernameCookie.serialize(null));
 
   return redirect(redirectTo ? redirectTo : '/', {
     headers,
@@ -126,4 +199,13 @@ export const sloUsernameCookie = createCookie('sloUsername', {
   /* An Hour before sloUsername Cookie Expires, after which logout will
      redirect to session-less sign-in button page */
   maxAge: 60 * 60,
+});
+
+export const effectiveUsernameCookie = createCookie('effectiveUsername', {
+  domain: process.env.NODE_ENV === 'production' ? new URL(process.env.APP_URL).hostname : undefined,
+  sameSite: 'strict',
+  httpOnly: true,
+  secure: process.env.NODE_ENV === 'production',
+  /* A week same as session if remembered */
+  maxAge: 60 * 60 * 24 * 7, // 7 days
 });
--- a/app/session.server.ts
+++ b/app/session.server.ts
@@ -82,10 +82,10 @@ export async function getEffectiveUser(request: Request) {
 
 export async function setEffectiveUsername(
   username: string,
-  name: string | undefined
+  effectiveUsername?: string
 ): Promise<string> {
   if (await isAdmin(username)) {
-    return await effectiveUsernameCookie.serialize(name);
+    return await effectiveUsernameCookie.serialize(effectiveUsername);
   }
 
   return await effectiveUsernameCookie.serialize(username);
--- a/app/utils.ts
+++ b/app/utils.ts
@@ -95,6 +95,14 @@ export function useOptionalUser(): User | undefined {
   return data.user;
 }
 
+export function useOptionalEffectiveUser(): User | undefined {
+  const data = useMatchesData('root');
+  if (!data || !isUser(data.effectiveUser)) {
+    return undefined;
+  }
+  return data.effectiveUser;
+}
+
 export function useUser(): User {
   const maybeUser = useOptionalUser();
   if (!maybeUser) {
@@ -103,6 +111,17 @@ export function useUser(): User {
     );
   }
   return maybeUser;
+}
+
+export function useEffectiveUser(): User {
+  const maybeEffectiveUser = useOptionalEffectiveUser();
+  if (!maybeEffectiveUser) {
+    throw new Error(
+      'No effective user found in root loader, but user is required by useEffectiveUser. If user is optional, try useOptionalUser instead.'
+    );
+  }
+
+  return maybeEffectiveUser;
 }
 
 /**
--- a/app/components/dns-record/form.tsx
+++ b/app/components/dns-record/form.tsx
@@ -11,7 +11,7 @@ import {
 } from '@chakra-ui/react';
 import { Form } from '@remix-run/react';
 import type { DnsRecord } from '@prisma/client';
-import { useUser } from '~/utils';
+import { useEffectiveUser } from '~/utils';
 import FormField from './form-field';
 import { useMemo } from 'react';
 import type { z } from 'zod';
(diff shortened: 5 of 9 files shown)


app/session.server.ts
```diff
@@ -36,6 +36,22 @@ export async function getUsername(request: Request): Promise<User['username'] |
...
+export async function getEffectiveName(request: Request): Promise<string> {
+  const cookie = request.headers.get('Cookie');
+  const effectiveName = await effectiveNameCookie.parse(cookie);
+  return effectiveName;
+}
+
+export async function setEffectiveName(username: string, name: string): Promise<string> {
+  const user = await getUserByUsername(username);
+
+  if (user?.group.includes('admin')) {
```
humphd, line 48: Let's fix the `User` returned in https://github.com/DevelopingSpace/starchart/blob/main/app/models/user.server.ts#L7-L21 and re-use/modify the code in https://github.com/DevelopingSpace/starchart/blob/main/app/models/user.server.ts#L61-L77 to add properties to the `User` object that indicate if this is an admin.

```
User.isAdmin
```

This way we can do the `/mycustomdomain(-dev)?-admins/.test(group)` stuff to set it on the object.

humphd (changes requested):

app/session.server.ts
```diff
@@ -36,6 +36,22 @@ export async function getUsername(request: Request): Promise<User['username'] |
...
+    return await effectiveNameCookie.serialize(name);
+  }
+
+  return await effectiveNameCookie.serialize(user?.username);
```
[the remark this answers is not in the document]
sfrunza13: right, I had tried to use getUser() before and this is a relic from that. We should have a user and if we don't I will throw.

humphd (changes requested):
Looking better, but I want to see us decorate the `user` object with more details.  The front-end is going to need to know the correct `username` to use, whether this is an admin, etc via the `User` we return in https://github.com/DevelopingSpace/starchart/blob/main/app/root.tsx#L34.

Please add this info to that `User` object so it can get passed around the app.

app/session.server.ts
sfrunza13: Is making use of isAdmin when setting effectiveName enough or should I do as you said and make a boolean property on a user for whether they are Admin?

humphd: Whatever you do, we need to put it on the `User` so we can pass that info around.  But using `isAdmin()` like you are is good internally for your needs.

sfrunza13:
> 

So it's ok to change the prisma schema for user?

humphd:
You don't need to change the database, change the object we send back when you query it, same as I do with `baseDomain` which isn't in the db either.

app/models/user.server.ts
```diff
@@ -17,7 +18,7 @@ export async function getUserByUsername(username: PrismaUser['username']) {
   }
 
   // Decorate with the user's base domain as well
-  return { ...user, baseDomain: buildUserBaseDomain(username) };
+  return { ...user, baseDomain: buildUserBaseDomain(username), isAdmin: isAdmin(username) };
```
humphd, line 21: Throw the `effectiveUsername` on here too.  So we know: 

- are you an admin?
- what is your real username? (who are you really)
- what is your current effective username? (which user are you actually pretending to be)

humphd (changes requested):

app/models/user.server.ts
sfrunza13: I'm not sure how to do this since my effective user name is currently passed as a cookie from request to request on the header and we don't have access to the header in the getUserByUsername method.

app/models/user.server.ts
```diff
@@ -3,21 +3,29 @@ import logger from '~/lib/logger.server';
...
 export interface User extends PrismaUser {
   // the base domain to use for all DNS records created by this user
   // (e.g., jsmith.starchart.com with records using *.jsmith.starchart.com)
   baseDomain: string;
+  effectiveUsername: string;
+  isAdmin: boolean;
 }
 
-export async function getUserByUsername(username: PrismaUser['username']) {
+export async function getUserByUsername(username: PrismaUser['username'], request: Request) {
```
humphd, line 16: OK, I agree this is the wrong place to do this.  Let's do it in the `root.tsx` loader.

On the server side, we'll read the username/effective username out of session.  On client, we'll send it with the root on the loader.

humphd (changes requested):

app/models/user.server.ts
sfrunza13: So for client to use it I would make a utils method using useMathcesData() like we have for user?

humphd: The client would get it from the loader, which can access the `request` and also the db, and put all the data on the User object we return in `root.tsx`.

app/models/user.server.ts
```diff
@@ -8,6 +8,8 @@ export interface User extends PrismaUser {
   // the base domain to use for all DNS records created by this user
   // (e.g., jsmith.starchart.com with records using *.jsmith.starchart.com)
   baseDomain: string;
+  effectiveUsername: string;
```
humphd, line 11: Please document what this is so future maintainers understand the intent.

app/root.tsx
```diff
@@ -30,8 +30,13 @@ export const links: LinksFunction = () => [
 ];
 
 export async function loader({ request, context }: LoaderArgs) {
+  const effectiveUsername = await getEffectiveName(request);
+  const userBeforeEffectiveName = await getUser(request);
```
humphd, line 34: Question: if both `getEffectiveName` and `getUser` both require `request`, can you not return the `effectiveUsername` on the result of `getUser` (i.e., you could do this internally in `getUser`)?

app/session.server.ts
```diff
@@ -36,6 +38,20 @@ export async function getUsername(request: Request): Promise<User['username'] |
   return username;
 }
 
+export async function getEffectiveName(request: Request): Promise<string> {
```
humphd, line 41: `getEffectiveUsername`

humphd: Also, change all `effectiveName` uses everywhere to `effectiveUsername`

humphd (changes requested):

humphd (changes requested):
Testing this locally, here is what I get on the client:

```json
{
  "username": "user1",
  "displayName": "Johannes Kepler",
  "email": "user1@myseneca.ca",
  "group": "mycustomdomain-faculty",
  "createdAt": "2023-03-28T17:18:24.557Z",
  "updatedAt": "2023-03-28T17:18:24.557Z",
  "deactivated": false,
  "baseDomain": "user1.starchart.com",
  "isAdmin": {},
  "effectiveUsername": null
}
```

According to the type of `User`, the `effectiveUsername` is `string`, but here we get `null`.  Should we set the `effectiveUsername` to be the same as `username` if it's not set (e.g., regular user)?

Next, the `isAdmin` is an `Object` vs. a `boolean`.

app/session.server.ts
```diff
@@ -90,6 +90,19 @@ export async function requireUser(request: Request) {
   throw await logout(request);
 }
 
+export async function requireAdmin(request: Request) {
+  const username = await requireUsername(request);
+
+  const user = await getUserByUsername(username);
+
+  if (user) {
+    if (user.isAdmin) {
```
humphd, line 99:
```ts
if (user?.isAdmin) {
  return user;
}
throw redirect('/');
```

app/utils.ts
```diff
@@ -105,6 +105,26 @@ export function useUser(): User {
   return maybeUser;
 }
 
+export async function useEffectiveUser(): Promise<User> {
```
humphd, line 109: I think this should be:

```ts
useEffectiveUsername()
```

The `user` object will be identical.  It's just that we're going to use this hook to get the specific "username" that the current user is using, which will either be:

1. `user.effectiveUsername` if the cookie is set and we're using someone else's username
2. `user.username` if the cookie is not set and we're using our own username

```ts
export async function useEffectiveUsername(): User['username'] {
  // The user must be logged in to call this.  Use `useOptionalUser()` if you're not sure
  const { username, effectiveUsername } = useUser();
 
  // Your username is either your real username, or the username you are temporarily using as an admin 
  return effectiveUsername || username;
}
```

app/models/user.server.ts
```diff
@@ -13,7 +13,7 @@ export interface User extends PrismaUser {
    username, admins (identified by group) will start with an empty string
    that can be changed to another users username on the admin portal for 
    them to impersonate */
-  effectiveUsername: string;
+  effectiveUsername?: string;
```
humphd, line 16: The type here should be `User['username']` so it matches what `username` is when we pass it around.

humphd (changes requested):

app/session.server.ts
```diff
@@ -36,15 +38,30 @@ export async function getUsername(request: Request): Promise<User['username'] |
   return username;
 }
 
+export async function getEffectiveUsername(request: Request): Promise<string> {
+  const cookie = request.headers.get('Cookie');
+  const effectiveUsername = await effectiveUsernameCookie.parse(cookie);
+  return effectiveUsername;
```
humphd, line 44: Let's return this like so:

```ts
return effictiveUsername as User['username'];
```

And you can drop the return type on the function.

app/session.server.ts
```diff
@@ -36,15 +38,30 @@ export async function getUsername(request: Request): Promise<User['username'] |
...
   const username = await getUsername(request);
+  const effectiveUsername = await getEffectiveUsername(request);
   if (username === undefined) {
     return null;
   }
 
   const user = await getUserByUsername(username);
   if (user) {
-    return user;
+    return { ...user, effectiveUsername: effectiveUsername };
```
humphd, line 64: When the property name and value are the same, just use it once:

```ts
return { ...user, effectiveUsername };
```

app/session.server.ts
```diff
@@ -36,15 +38,30 @@ export async function getUsername(request: Request): Promise<User['username'] |
...
+    return await effectiveUsernameCookie.serialize(name);
+  }
+
+  return await effectiveUsernameCookie.serialize(username);
+}
+
 export async function getUser(request: Request) {
   const username = await getUsername(request);
+  const effectiveUsername = await getEffectiveUsername(request);
   if (username === undefined) {
```
humphd, line 58:
```ts
if (!username) {
```

humphd (changes requested):

app/models/user.server.ts
```diff
@@ -17,7 +24,11 @@ export async function getUserByUsername(username: PrismaUser['username']) {
   }
 
   // Decorate with the user's base domain as well
-  return { ...user, baseDomain: buildUserBaseDomain(username) };
+  return {
+    ...user,
+    baseDomain: buildUserBaseDomain(username),
```
humphd, line 29: This is not going to work either, since the `baseDomain` includes your username, which will need to be the `effectiveUsername` in order to work if I'm impersonating someone.

Need to think about this...

app/utils.ts
sfrunza13: I'm not sure how to go about this still, I have replaced what I had with what is suggested and it works for when we display it like on header instead of rendering a property of the user object I just render the effective username but we would have to have a different effective user object for the use of baseDomain for example. Or we could change how the method works so that getBaseDomain checks whether isAdmin and a non-empty effective username exist together and if that is the case will instead use the effectiveUsername.

app/models/user.server.ts
```diff
@@ -8,6 +8,13 @@ export interface User extends PrismaUser {
   // the base domain to use for all DNS records created by this user
   // (e.g., jsmith.starchart.com with records using *.jsmith.starchart.com)
   baseDomain: string;
+  /* the effectiveUsername, this is for admin impersonating users. Regular
+   users will have an effectiveUsername that is always the same as their
+   username, admins (identified by group) will start with an empty string
+   that can be changed to another users username on the admin portal for 
+   them to impersonate */
+  effectiveUsername?: User['username'];
```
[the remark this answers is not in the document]
sfrunza13: I don't think we do need it since everything kind of comes from the cookie now, is there a better way of doing this than with the cookie? Maybe keeping track of it on this property we are thinking about removing, but this is dynamically appended every time we load same as baseDomain so that does not make sense. Anyways, I am ok with removing it I think.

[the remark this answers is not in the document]
sfrunza13: Honestly it's kind of hard keeping track of stuff

app/root.tsx
```diff
@@ -31,16 +31,16 @@ export const links: LinksFunction = () => [
...
-      effectiveUser: await getEffectiveUser(request),
-      nonce: context.nonce as string,
-    }
-    // ,
-    // { headers: { 'Set-Cookie': await setEffectiveUsername('user3', 'user1') } }
-    // This is how I tested it, after refreshing once since user3 is admin it will mock user1
-  );
+  return json({
+    user: await getUser(request),
+    /* the effectiveUsername, this is for admin impersonating users. Regular
```
humphd, line 36: Can you format this differently?

```ts
user: await getUser(request),

/**
 * effictiveUser: ...
 * ...
 * ...
 */
effectiveUser: await ...
```

Also, `effectiveUser` vs. `effectiveUsername`.

app/session.server.ts
```diff
@@ -58,7 +59,7 @@ export async function getEffectiveUser(request: Request) {
 
   const user = await getUserByUsername(effectiveUsername);
   if (user) {
-    return { ...user, effectiveUsername }; // should we have an effectiveUsername when mocking?
+    return user; // should we have an effectiveUsername when mocking?
```
humphd, line 62: You can drop this comment

humphd (changes requested):

app/models/user.server.ts
humphd: So I think we're now good here, and we'll pass in the right username/effectiveUsername when we call this.  If that's not right, please correct me.

humphd (approved):
I *think* this is right.

In a follow-up, we should add/update e2e tests for this stuff.

Does anything call `useUser()` now, or is it all `useEffectiveUser()`?    Maybe the header will do it when an admin logs in and we want to show different usernames?  

We should also file an issue to hook up the `isAdmin` stuff to the `/admin` route.

sfrunza13:
Everything calls useEffectiveUser, if there is no effectiveUsername it will call getUser() in session server. As it stands there is no call to useUser(). I think we may be able to take it out of the root loader.

humphd:
> Everything calls useEffectiveUser, if there is no effectiveUsername it will call getUser() in session server. As it stands there is no call to useUser(). I think we may be able to take it out of the root loader.

OK, thanks for confirming.  I think we'll still want it to show that an admin is impersonating another user, but probably in no other cases.

app/models/user.server.ts
Eakam1007: Maybe I am confused, but we need to change places where `requireUser` is being used? That function calls getUsername, and is used in a few places such as in the action of the new DNS record route:
https://github.com/DevelopingSpace/starchart/blob/14c32e101a22539adbff62ac007011a299cc282e/app/routes/__index/dns-records/new.tsx#L14-L16

sfrunza13: You're right, I will have to change that as well. I was thinking maybe in a following PR but I'm not sure, I could do it here as well.

humphd: I would do it here.  It could be smart about looking for either vs needing to be renamed (i.e., `requireUser` and `requireUsername` are still fine).

humphd (approved):

Eakam1007 (dismissed):

humphd:
@sfrunza13 can you rebase and fix this conflict?  Then we can get this in too!

sfrunza13:
> 

yes, I will do this

app/session.server.ts
```diff
@@ -43,9 +45,47 @@ export async function getUsername(request: Request): Promise<User['username'] |
...
+): Promise<User['username'] | undefined> {
+  const cookie = request.headers.get('Cookie');
+  const effectiveUsername = await effectiveUsernameCookie.parse(cookie);
+  return effectiveUsername as User['username'];
+}
+
+export async function getEffectiveUser(request: Request) {
+  const username = await getUsername(request);
+  // get the effective username in the cookie at the time
+  const effectiveUsername = await getEffectiveUsername(request);
```
humphd, line 59: Run these at the same time:

```ts
const [username, effectiveUsername] = await Promise.all([getUsername(request), getEffectiveUsername(request)]);
```

app/session.server.ts
```diff
@@ -43,9 +45,47 @@ export async function getUsername(request: Request): Promise<User['username'] |
...
+  if (!username) {
+    return null;
+  }
+
+  // if there is no effective username let the admin be himself
+  if (!effectiveUsername || effectiveUsername === '') {
```
humphd, line 65: It makes me nervous that `effectiveUsername` could be `''`.  When could this happen and why?  Can we avoid that?

app/session.server.ts
```diff
@@ -43,9 +45,47 @@ export async function getUsername(request: Request): Promise<User['username'] |
...
+
+export async function getEffectiveUser(request: Request) {
+  const username = await getUsername(request);
+  // get the effective username in the cookie at the time
+  const effectiveUsername = await getEffectiveUsername(request);
+  if (!username) {
+    return null;
+  }
+
+  // if there is no effective username let the admin be himself
```
humphd, line 64: I might update this wording slightly:

```
// if the user doesn't have an effective username, we're using their normal username (not impersonating)
```

app/session.server.ts
```diff
@@ -43,9 +45,47 @@ export async function getUsername(request: Request): Promise<User['username'] |
...
+  if (!effectiveUsername || effectiveUsername === '') {
+    return await getUser(request);
```
humphd, line 66: Here you use `getUser()` and in 69 you use `getUserByUsername()`.  Can we use the same path in both cases?  Not critical, but it feels odd that we have different code paths.

app/session.server.ts
```diff
@@ -62,12 +102,14 @@ export async function requireUsername(
   redirectTo: string = new URL(request.url).pathname
 ) {
   const username = await getUsername(request);
+  const effectiveUsername = await getEffectiveUsername(request);
```
humphd, line 105: Same comment as above, run at same time.

humphd (changes requested):

humphd:
Also, please squash this down into 1 commit at this point.

app/session.server.ts
sfrunza13: This was what I remembered from our discussion, admins would have an empty effectiveUsername until they select one

humphd: Can "empty" be `undefined` vs `''`?

app/session.server.ts
```diff
@@ -43,9 +45,56 @@ export async function getUsername(request: Request): Promise<User['username'] |
...
+  if (user) {
+    return user;
+  }
+
+  throw await logout(request);
+}
+
+export async function setEffectiveUsername(
+  username: string,
+  name: string | undefined
```
humphd, line 85: nit: you can use `name?: string` to mean "this is a `string` or undefined".

humphd: Should `name` really be called `effectiveUsername`?  I think using `name` is ambiguous.

humphd (changes requested):
This is great, thanks for persevering through the many reviews.  I've left 2 final updates, and this is R+ from me.

humphd (approved):

Eakam1007 (approved):

sfrunza13:
The above was some misclicks

PR #480, opened 2023-03-30
Allow only alphanumerical and hyphen in record

### LINKED ISSUE #437: New DNS Record should only allow letters, digits and hyphen for name
Today on staging, we ran into an issue where Let's Encrypt refuses to allow any name with underscores.  Doing a bit of digging, we found that they only support **letters, digits, and hyphens**.  Route53 supports more than this, but let's limit the names that our users can enter to only allow letters, digits, and hyphens.

More details here https://community.letsencrypt.org/t/underscore-in-subdomain-fails/31431/6

Genne23v:
Close #437 

- Update regex to allow only alphanumeric characters and `-`
- Record name can't start or end with `-`
- Record name can't have consecutive `-` in the middle

### DIFF
--- a/app/lib/dns.server.ts
+++ b/app/lib/dns.server.ts
@@ -190,9 +190,9 @@ export const getChangeStatus = async (changeId: string) => {
 
 /* Domain name rules
 1. Full domain name pattern should be [subdomain].[username].rootDomain.com
-2. Subdomain can contain only alphanumerical characters, '-', and '_'
+2. Subdomain can contain only alphanumerical characters and '-'
 3. Subdomain should not start or end with -
-4. Subdomain cannot contain multiple consecutive '-' or '_'
+4. Subdomain cannot contain multiple consecutive '-'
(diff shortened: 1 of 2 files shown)


Eakam1007 (approved):

dadolhay (commented):
Why are we inhibiting users to start with `_` ? Providers requiring domain verification may very well require the use of it (just like Let's encrypt does)

dadolhay (changes requested):
Looking at the ticket I don't think this is correct.

Let's encrypt does not allow _ ut the cert is going to be issued on the user's domain, so for that to be a problem the username would have to contain _.

I think we are solving a non-issue here

dadolhay:
@Genne23v @humphd 
Correct, I think is not needed, at least not the part that is limiting the use of _ characters

humphd:
Cool, let's take that bit out, and fix @Genne23v?

Genne23v:
> 

@humphd I think I can just close this PR. The change is limiting underscores in subdomain.

humphd:
Sounds good.

PR #483, opened 2023-03-30
Tests for functions that interact with system state

### LINKED ISSUE #482: Add tests for functions that interact with system state
We can have some tests written for the functions in `~\models\system-state.server.ts`.

cychu42:
Closes #482

## Changes
* Add tests for the file `~\models\system-state.server.ts`.
* Adjust `setIsReconciliationNeeded()` to allow `initialize()` to execute, in cases where the system state row doesn't exist in database table `SystemState`. Previously, it instead produces a serialization error and does not execute the function call for `initialize()` in the catch block.

## Test
1. setup test database via `npm run setup:test`
2. run the test file via `npm run test system-state.server.test.ts`

### DIFF
--- a/app/models/system-state.server.ts
+++ b/app/models/system-state.server.ts
@@ -11,7 +11,7 @@ import type { SystemState } from '@prisma/client';
  * to our Records table
  */
 
-function initialize() {
+export function initialize() {
   return prisma.systemState.create({
     data: {
       unique: StateEnumType.unique,
--- a/app/models/system-state.server.ts
+++ b/app/models/system-state.server.ts
@@ -29,7 +29,7 @@ export function getIsReconciliationNeeded(): Promise<SystemState['reconciliation
     .then((data) => data?.reconciliationNeeded ?? true);
 }
 
-export async function setIsReconciliationNeeded(
+export function setIsReconciliationNeeded(
   reconciliationNeeded: SystemState['reconciliationNeeded']
 ) {
   return prisma.systemState
--- a/app/models/system-state.server.ts
+++ b/app/models/system-state.server.ts
@@ -29,11 +29,12 @@ export function getIsReconciliationNeeded(): Promise<SystemState['reconciliation
     .then((data) => data?.reconciliationNeeded ?? true);
 }
 
-export function setIsReconciliationNeeded(
+export async function setIsReconciliationNeeded(
   reconciliationNeeded: SystemState['reconciliationNeeded']
 ) {
+  let result;
   try {
-    return prisma.systemState.update({
+    result = await prisma.systemState.update({
       data: { reconciliationNeeded },
       where: { unique: StateEnumType.unique },
     });
@@ -46,4 +47,5 @@ export function setIsReconciliationNeeded(
 
     return initialize();
   }
+  return result;
 }
--- a/app/models/system-state.server.ts
+++ b/app/models/system-state.server.ts
@@ -32,20 +32,18 @@ export function getIsReconciliationNeeded(): Promise<SystemState['reconciliation
 export async function setIsReconciliationNeeded(
   reconciliationNeeded: SystemState['reconciliationNeeded']
 ) {
-  let result;
-  try {
-    result = await prisma.systemState.update({
+  return prisma.systemState
+    .update({
       data: { reconciliationNeeded },
       where: { unique: StateEnumType.unique },
-    });
-  } catch (error) {
-    /**
-     * This should never happen, as the table should always be seeded.
-     * In case it isn't, let's seed it here Next queue run will set the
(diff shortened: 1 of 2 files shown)


app/models/system-state.server.ts
```diff
@@ -29,11 +29,12 @@ export function getIsReconciliationNeeded(): Promise<SystemState['reconciliation
...
 }
 
-export function setIsReconciliationNeeded(
+export async function setIsReconciliationNeeded(
   reconciliationNeeded: SystemState['reconciliationNeeded']
 ) {
+  let result;
   try {
-    return prisma.systemState.update({
+    result = await prisma.systemState.update({
```
Genne23v, line 37: Could you explain why not returning directly?

cychu42: Before, try block wasn't catching a serialization error when the `update` fails to find the row. When I checked test coverage, the line (47 or 48) that calls `initialize()` was not covered.

I think what happened before was that `update` itself returns the error as shown in [documentation](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#return-type-4), and the return statement returns it , making it land outside of the try block, so catch block was not properly triggered to run `initialize()` .
With this change, I can get catch block to trigger and run `initialize()`.

Eakam1007: Wouldn't this change affect the places where this function is being called?

Eakam1007: I think we can simplify this in a way similar to how `getIsReconciliationNeeded` is setup:
```ts
  return prisma.systemState
    .update({
      data: { reconciliationNeeded },
      where: { unique: StateEnumType.unique },
    })
    .catch(() => {
      return initialize();
    });
```

test/unit/system-state.server.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    await prisma.systemState.deleteMany().catch(() => {});
+  });
+
+  test('can get system state correctly', async () => {
+    let result = await getIsReconciliationNeeded();
+    expect(result).toBe(state.reconciliationNeeded);
+  });
+
+  test('can get system state correctly when it does not exist', async () => {
+    await prisma.systemState.deleteMany().catch(() => {});
```
Eakam1007, line 51: An alternative would be to only initialize in the first test, and then delete it there. Not sure if it is any better. What do you think? Same for the next describe block

Eakam1007 (changes requested):

app/models/system-state.server.ts
humphd: See also #493

test/unit/system-state.server.test.ts
cychu42: For me, it make sense to start with a default state (system state exist). If anything, I feel I should initialize it again at the end of this test, to restore the default state for other tests that might come after this.

Eakam1007: Alrighty, makes sense. 

Though if you wanna add more tests to this, I would avoid readding the state. Tests should not depend on each other imo. Better to use a `beforeEach` hook in that case

test/unit/system-state.server.test.ts
```diff
@@ -51,6 +51,9 @@ describe('getIsReconciliationNeeded()', () => {
     await prisma.systemState.deleteMany().catch(() => {});
     let result = await getIsReconciliationNeeded();
     expect(result).toBe(true);
+
+    // Restore state
+    await initialize();
```
Eakam1007, line 56: I would remove this, because we immediately delete the state after this test ends

cychu42:
I updated the code. Thank you for suggesting a different way of adjusting `setIsReconciliationNeeded`.

Eakam1007 (changes requested):

test/unit/system-state.server.test.ts
cychu42: Yeah, you are right. I guess whoever wants to add tests after this can figure it out when they read the code.

app/models/system-state.server.ts
```diff
@@ -29,21 +29,21 @@ export function getIsReconciliationNeeded(): Promise<SystemState['reconciliation
     .then((data) => data?.reconciliationNeeded ?? true);
 }
 
-export function setIsReconciliationNeeded(
+export async function setIsReconciliationNeeded(
```
Eakam1007, line 32: I believe we don't need the async anymore with this

Eakam1007 (changes requested):
One small thing, otherwise looks great

Eakam1007 (approved):

Genne23v (approved):
LGTM!

cychu42:
Thank you all for the feedback.

PR #485, opened 2023-03-30
Create docker-production.yml similar to docker-staging.yml

humphd:
This creates a Docker swarm stack file for production.  It's very similar to staging (by design!) but all the `-dev` stuff has been removed and the hosted zone/root domain are different.

I've also removed the `DATABASE_SETUP=1` bit, since it's not something I want to run accidentally on production.

We won't need this for a few weeks, but I wanted to get it in, now that we have staging finalized and working well.

### DIFF
--- a/docker-production.yml
+++ b/docker-production.yml
@@ -1,0 +1,65 @@
+# Stack Definition for Production
+services:
+  # Redis is used to back our worker queues. It is not exposed.
+  redis:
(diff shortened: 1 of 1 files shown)


cychu42 (approved):
I'm not spotting issue from what I see. LGTM.
I can't wait to see it running on production.

PR #487, opened 2023-03-31
Add checks for deactivated users

### LINKED ISSUE #308: Allow users to be deactivated
While discussing the deletion of users based on the design for the Admin UI (#106), it was concluded that deletion would be a multi step process. The first step in this would set a flag on the user that is being deleted, effectively setting the user as `deactivated`. This would prevent users from being able to perform actions while their account is being deleted.

Further, while checking if a user is logged in (username is present in the session), we should only return the user if the flag is not set (the flag for `deactivated` is false for the user). This is to handle the case where a user gets deleted from the database but is still logged in (i.e. their session has not expired).

At minimum, the following needs to be done:
- Update the schema to add a `deactivated` flag to the user model
- Update `requireUsername` and `requireUser` methods to check for the flag and not return the username if the user is `deactivated`
- Add/update tests as needed

### LINKED ISSUE #394: Add checks for a user being deactivated when jobs are run
Based on #308 and #318, we now have a `deactivated` flag to specify which users are in the process of being deleted. This is supposed to be the first step in the deletion process. 
Thus, we need to add checks to different processes and jobs for creating certificates/DNS Records etc. to make sure that we handle users being deactivated.

Eakam1007:
Resolves #394, resolves #308 

This PR updates the `requireUsername` (and therefore also `requireUser`) function to logout the user if the user is deactivated. 
This in turn ensures that no actions can be performed on routes where we use these functions as the user will be immediately logged out, and redirected to the login page.
It also updates the certificate workers to throw an unrecoverable error if the user is deactivated

### DIFF
--- a/app/models/dns-record.server.ts
+++ b/app/models/dns-record.server.ts
@@ -4,6 +4,7 @@ import { prisma } from '~/db.server';
 import { setIsReconciliationNeeded } from './system-state.server';
 
 import type { DnsRecord } from '@prisma/client';
+import { isDeactivated } from './user.server';
 
 export function getDnsRecordsByUsername(username: DnsRecord['username']) {
   return prisma.dnsRecord.findMany({
@@ -36,6 +37,10 @@ export function getUserDnsRecordCount(username: DnsRecord['username']) {
 export async function createDnsRecord(
   data: Required<Pick<DnsRecord, 'username' | 'type' | 'subdomain' | 'value'>> & Partial<DnsRecord>
 ) {
+  if (await isDeactivated(data.username)) {
+    throw new Error('User is deactivated');
+  }
+
   if (process.env.USER_DNS_RECORD_LIMIT) {
     if ((await getUserDnsRecordCount(data.username)) >= Number(process.env.USER_DNS_RECORD_LIMIT)) {
       throw new Error('User has reached the maximum number of dns records');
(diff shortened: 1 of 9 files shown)


app/models/dns-record.server.ts
```diff
@@ -36,6 +37,10 @@ export function getUserDnsRecordCount(username: DnsRecord['username']) {
 export async function createDnsRecord(
   data: Required<Pick<DnsRecord, 'username' | 'type' | 'subdomain' | 'value'>> & Partial<DnsRecord>
 ) {
+  if (await isDeactivated(data.username)) {
```
humphd, line 40: `isUserDeactivated` would be a better name I think.

app/session.server.ts
```diff
@@ -59,6 +60,12 @@ export async function requireUsername(
     const searchParams = new URLSearchParams([['redirectTo', redirectTo]]);
     throw redirect(`/login?${searchParams}`);
   }
+
+  // Logout user if they are deactivated
```
humphd, line 64: I wonder if you should do this in `getUsername` instead, so it happens earlier, and for all paths where we get a user/username?

humphd (changes requested):
Adding our auth-expert @sfrunza13 for review too.

app/session.server.ts
Eakam1007: I went with `requireUsername` because `getUsername` seemed like only getting the username instead of requiring a valid username. But I will look into the paths to see if it is better to move this logic there

app/models/dns-record.server.ts
Eakam1007: Yeah, that is better. I can update this

humphd (approved):
Nice

sfrunza13 (approved):

PR #489, opened 2023-03-31
FE: Fixing the style of certificate text

Ririio:
Should fix the issue with certificate not having the correct format when displayed

image


humphd (approved):
This is great!

Please file separate issues on:

1. copy-paste not including the newlines
2. multiple inclusions of the value (should only be 1)

Eakam1007:
Is this fixing a specific issue? If so, could you please link it to this PR?

Eakam1007 (approved):

PR #497, opened 2023-04-02
Bug: fix issue with pending page not displaying

### LINKED ISSUE #494: Test Certificate UI flow, confirm it works on staging
I'm trying the Certificate UI on staging, and it seems to hang with no update screen:

Screenshot 2023-04-01 at 2 28 12 PM

The Request a Certificate button is disabled, but nothing else happens.  If I refresh, it comes back to the same screen, but the button is now enabled.

I haven't check the logs yet, but this doesn't seem to be working?

cc @Ririio

Ririio:
Fixes #494 

`getCertificateByUsername` only returns status that are `issued`, I reverted it back to its previous state of returning any certificate regardless of the status which should fix the issue with pending page not showing

### DIFF
--- a/app/models/certificate.server.ts
+++ b/app/models/certificate.server.ts
@@ -11,7 +11,7 @@ export async function getCertificateByUsername(username: Certificate['username']
    */
   return prisma.certificate
     .findMany({
-      where: { username, status: 'issued' },
+      where: { username },
       orderBy: { validFrom: 'desc' },
       take: 1,
     })


humphd (approved):
It looks like this change got lost in a rebase during https://github.com/DevelopingSpace/starchart/commit/5ddbcb53facae2f5cde80f8c014fd2389068f969#diff-09e23cf8b6b305b3adc79e7ec861a7367b24638e8638db0fc771443874982b93 (cc @Myrfion)

@Ririio you might want to scan through that PR to see if anything else is out of sync.

SerpentBytes (approved):

Eakam1007 (approved):

PR #503, opened 2023-04-03
Update avatar colour

humphd:
I updated the avatar colours.  I haven't been loving the gray background.

Screenshot 2023-04-03 at 8 49 44 AM


Eakam1007 (approved):
That does look better

PR #512, opened 2023-04-04
Use journald log driver, update DEPLOY docs on logging

### LINKED ISSUE #117: Figure out logging and log rotation in production
We're going to use RedHat Enterprise 8 (RHEL8) in production.  Let's figure out the right way to have our Docker containers do logging with the system logger so that we don't fill the disk with log files (i.e., let it handle logging, log file rotation, etc).

humphd:
Fixes #117 

This switches Docker to use the `journald` log driver instead of the default `json-file` log driver, see https://docs.docker.com/config/containers/logging/journald/.  Doing so means that the OS (RHEL8) will take care of our log files/rotation for us.

I've also updated the docs to show how to use this.  You can get logs via `docker` or `journalctl`.

This is running on staging, where I've confirmed that it's working.  In order to check both Redis and MyCustomDomain, I had to recreate the containers.  Updating Redis to a new version is the easiest way, and also takes care of doing the update at the same time.

### DIFF
--- a/DEPLOY.md
+++ b/DEPLOY.md
@@ -117,10 +117,19 @@ To stop and remove the serivce:
 docker stack rm starchart
 ```
 
-You can then view logs for any of the services by name (e.g., `mycustomdomain`):
+You can then view logs for any of the services by name (e.g., `mycustomdomain` container inside the `starchart` service):
 
 ```sh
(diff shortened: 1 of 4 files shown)


Eakam1007 (approved):

Genne23v (approved):
It looks good to me!

PR #529, opened 2023-04-05
Connect Admin UI to landing page

humphd:
We don't currently have a way for an admin to get to the Admin page.  This adds another card to the landing page when a suer is an admin.

I've also cleaned up a few things in the landing page.  However, the responsive widths aren't perfect, so maybe one of you could advise me on how to get it to wrap sooner.

## Regular User
Screenshot 2023-04-05 at 2 30 26 PM

## Admin User
Screenshot 2023-04-05 at 2 30 42 PM

### DIFF
--- a/app/routes/__index/index.tsx
+++ b/app/routes/__index/index.tsx
@@ -19,12 +19,7 @@ export default function IndexRoute() {
         paddingY={{ base: '2', md: '7' }}
         gap={{ md: '16' }}
       >
-        <Flex
-          width={{ base: '100%', md: '3xl' }}
-          maxWidth="42em"
-          padding="5"
-          flexDirection="column"
-        >
+        <Flex width={{ base: '100%', md: '3xl' }} padding="5" flexDirection="column">
           <Heading size="lg" textAlign="center">
             Welcome to My.Custom.Domain!
           </Heading>
--- a/app/routes/__index/index.tsx
+++ b/app/routes/__index/index.tsx
@@ -20,7 +20,7 @@ export default function IndexRoute() {
         gap={{ md: '16' }}
       >
         <Flex
-          width={{ base: '100%', md: '75%' }}
+          width={{ base: '100%', md: '3xl' }}
           maxWidth="42em"
           padding="5"
           flexDirection="column"
--- a/app/routes/__index/index.tsx
+++ b/app/routes/__index/index.tsx
@@ -1,17 +1,15 @@
 import { Flex, Text, VStack, Link, Heading } from '@chakra-ui/react';
 import type { LoaderArgs } from '@remix-run/node';
-import { json } from '@remix-run/node';
 
 import { requireUsername } from '~/session.server';
 import LandingPageCard from '~/components/landing-page/landing-page-card';
+import { useUser } from '~/utils';
 
-export const loader = async ({ request }: LoaderArgs) => {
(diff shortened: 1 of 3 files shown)


Eakam1007:
I believe this should make them more responsive. I tested this on my PC:
image

app/routes/__index/index.tsx
```diff
@@ -23,11 +21,13 @@ export default function IndexRoute() {
       >
         <Flex
           width={{ base: '100%', md: '75%' }}
-          textAlign="center"
+          maxWidth="42em"
```
Eakam1007, line 24: You could probably edit the values for the width instead of setting a maxWidth. The `base` applies to all breakpoints until the `md` breakpoint. The `md` style applies to all breakpoints including `md` and larger

humphd: So what should I do?  Can you suggest some code for me?

humphd: oh, I see you did above!

humphd: OK, I did your fix above, but I'm not sure how to do the same for this.  If you can recommend something, I'd appreciate it.

humphd:
@Eakam1007 I'm stuck on how to do the selectors I need for these links to the instructions pages in testing now.

humphd:
I've redone the landing page cards as actual cards, which seems to work better.

app/routes/__index/index.tsx
Eakam1007: Instead of maxWidth,
```
width={{ base: '100%', md: '3xl' }}
```
I think that would look better. Currently, the text width is too narrow compared to the cards on smaller sizes:
image
After this change,
image

Eakam1007:
> @Eakam1007 I'm stuck on how to do the selectors I need for these links to the instructions pages in testing now.

Since the links are no longer part of the paragraph now, I think this would be good solution `getByRole('link', { name: 'these instructions' }).first()` for the first link
and `getByRole('link', { name: 'these instructions' }).nth(1)` for the second.

Eakam1007:
We could extend these to the admin cards but we do not have a way to mock login as admin yet

humphd:
> > @Eakam1007 I'm stuck on how to do the selectors I need for these links to the instructions pages in testing now.
> 
> Since the links are no longer part of the paragraph now, I think this would be good solution `getByRole('link', { name: 'these instructions' }).first()` for the first link and `getByRole('link', { name: 'these instructions' }).nth(1)` for the second.

Great, thank you.  Fixed!

app/routes/__index/index.tsx
```diff
@@ -22,12 +20,14 @@ export default function IndexRoute() {
         gap={{ md: '16' }}
       >
         <Flex
-          width={{ base: '100%', md: '75%' }}
-          textAlign="center"
+          width={{ base: '100%', md: '3xl' }}
+          maxWidth="42em"
```
Eakam1007, line 24: I believe this can be removed now

Eakam1007 (changes requested):
One small question, rest looks great!

I have created #531 to allow tests to login in as admin. Could you please create an issue for adding tests for the admin card?

Eakam1007 (approved):

dadolhay (approved):

Issue #524, opened 2023-04-05
Issues using certificate with AWS


humphd:
I did a real-world test of our system today.  My goal was to do the following:

1. Run an nginx container on AWS via Elastic Container Service, and put a Load Balancer in front of it
2. Create a CNAME record on My.Custom.Domain that points to the Load Balancer's domain
3. Create an HTTPS certificate on My.Custom.Domain
4. Import this certificate into AWS Certificate Manager
5. Associate this certificate with the Load Balancer, and allow traffic on https://mycustomdomain...

I am pleased to report that I was able to accomplish all of this, but I encountered some issues.  Here is the process.

1. I created an nginx cluster with load balancer at http://ec2co-ecsel-1rzway2anzrs6-856795930.us-east-1.elb.amazonaws.com/
2. I created a CNAME record at aws.davidhumphrey.stage.mystudentproject.ca which points to ec2co-ecsel-1rzway2anzrs6-856795930.us-east-1.elb.amazonaws.com (you can test it at http://aws.davidhumphrey.stage.mystudentproject.ca/)
3. I created a certificate, and it worked well.  I also received an email, see #523

The next part was more problematic.  In Amazon Certificate Manager, you are presented with the following screen:

<img width="624" alt="Screenshot 2023-04-05 at 9 29 34 AM" src="https://user-images.githubusercontent.com/427398/230095373-1f18ee72-9e89-4f26-ac01-f8d429ff09f5.png">

If I copy/paste our Public Key as the Certificate Body, and our Private Key as the Certificate Private Key, I get this:

<img width="626" alt="Screenshot 2023-04-05 at 9 31 09 AM" src="https://user-images.githubusercontent.com/427398/230095762-1a5901ac-7e32-4f08-8433-3b1638c385a4.png">

AWS is not liking that we're using the certificate chain vs. only the certificate.  If I edit this PEM file to only include the first certificate, and delete the second and third in the chain, it works.

If I paste our "Public Key" into the "Certificate chain" box, it works.

I was then able to do step 5. and add a Listener on port 443 for the Load Balancer, as well as updating its Security Group for HTTPS traffic and now I can create a secure connection at https://aws.davidhumphrey.stage.mystudentproject.ca/

NOTE: this will prompt you with a warning, since we're using the Let's Encrypt Staging server, and it's not in the Browser's trusted list of CAs.  You can click "Advanced" and allow the connection (I promise you, I'm not trying to hack you :)

[...]

PR #530, opened 2023-04-05
Add separate Certificate Views for different use cases/services

humphd:
I loved @dadolhay's idea in the call today to have some guided-views for working with the various certificate parts, in ways that match the service you are targeting.

I've made a start of this by splitting the certificate UI into panels.  So far I have a General panel, node.js panel, and AWS panel.

This needs the work from #524 to split the cert into more parts.  But once that is done, we can show people exactly what they need, and call it what they expect.

Screenshot 2023-04-05 at 3 36 26 PM

Screenshot 2023-04-05 at 3 36 35 PM

Screenshot 2023-04-05 at 3 36 45 PM

### DIFF
--- a/app/components/certificate/panels/aws.tsx
+++ b/app/components/certificate/panels/aws.tsx
@@ -1,0 +1,53 @@
+import { Divider, TabPanel, Text, Link } from '@chakra-ui/react';
+import CertificateDisplay from '../certificate-display';
+
+interface AwsPanelProps {
+  publicKey: string;
+  privateKey: string;
+}
+
+export default function AwsPanel({ publicKey, privateKey }: AwsPanelProps) {
+  return (
+    <TabPanel>
+      <Text marginBottom={4}>
+        You can{' '}
+        <Link
+          href="https://docs.aws.amazon.com/acm/latest/userguide/import-certificate-api-cli.html"
+          isExternal
+        >
+          import your certificate
+        </Link>
+        into{' '}
+        <Link href="https://docs.aws.amazon.com/acm/latest/userguide/acm-overview.html" isExternal>
+          AWS Certificate Manager
+        </Link>{' '}
+        and use it with{' '}
+        <Link href="https://docs.aws.amazon.com/acm/latest/userguide/acm-services.html" isExternal>
+          many AWS services
+        </Link>
+        .
+      </Text>
+
+      <CertificateDisplay
+        title="Certificate body"
+        value={publicKey}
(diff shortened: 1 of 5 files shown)


app/components/certificate/panels/aws.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+      <CertificateDisplay
+        title="Certificate body"
+        value={publicKey}
+        description="Public certificate body."
+        pathname="cert"
+      />
+      <Divider />
+      <CertificateDisplay
+        title="Certificate private Key"
```
SerpentBytes, line 39: Should the letter K in the key be capitalized?

humphd: No, good catch.  Fixing...

humphd: Fixed.

app/components/certificate/panels/node.tsx
```diff
@@ -0,0 +1,45 @@
...
+export default function NodePanel({ publicKey, privateKey }: NodePanelProps) {
+  return (
+    <TabPanel>
+      <Text marginBottom={4}>
+        You can use your certificate with{' '}
+        <Link
+          href="https://nodejs.org/api/https.html#httpscreateserveroptions-requestlistener"
+          isExternal
+        >
+          node.js to create a secure HTTPS server
```
SerpentBytes, line 18: I think we could capitalize the first letter to match the [registered name](https://nodejs.org/en).

humphd: OK, done.

SerpentBytes (approved):

Eakam1007:
Wouldn't it be better to also have the chain and the full chain in the general tab or would that be added as part of #524 ?

humphd:
> Wouldn't it be better to also have the chain and the full chain in the general tab or would that be added as part of #524 ?

Yes, that's what I want to do as soon as we have that data.  Some tabs would use all 4 pieces, some only 3, some only 2, etc.  General can show and explain everything.

Eakam1007:
> Yes, that's what I want to do as soon as we have that data. Some tabs would use all 4 pieces, some only 3, some only 2, etc. General can show and explain everything.

Alright, I will try to get that in tomorrow. I have most of it figured out

Eakam1007 (approved):

Eakam1007 (approved):

SerpentBytes (approved):

PR #532, opened 2023-04-06
Split certificate public key and chain

### LINKED ISSUE #490: Confirm certificate received from let's encrypt
The certificate seems to be rendering duplicates of itself as can be seen here

image

Looking through the database, something seems to be occuring from the back-end for this to occur

image

### LINKED ISSUE #524: Issues using certificate with AWS

Eakam1007:
Resolves #524, Resolves #490

- Created another field to store the chain
- Updated certificate code to split the returned result into a public key (stored in the certificate field), and the intermediate certificate(s)/chain (stored in chain field)

Now the UI only shows the public key. I confirmed this by decoding the certificate using an online tool
image

### DIFF
--- a/app/lib/lets-encrypt.server.ts
+++ b/app/lib/lets-encrypt.server.ts
@@ -416,11 +416,18 @@ class LetsEncrypt {
     });
 
     this.#order = await this.#client.finalizeOrder(this.#order, csr);
-    const certificate = await this.#client.getCertificate(this.#order);
+
+    // fullChain is the publicCertificate + the intermediate certificate(s) for Let's Encrypt
+    const fullChain = await this.#client.getCertificate(this.#order);
+
+    // certificate is the public certificate, chain is the intermediate certificate(s) for Let's Encrypt
+    const [certificate, ...chain] = acme.crypto.splitPemChain(fullChain);
 
     return {
       privateKey: key.toString(),
       certificate,
+      chain,
+      fullChain,
       validFrom: new Date(),
       validTo: new Date(this.#order.expires!),
     };
--- a/app/lib/lets-encrypt.server.ts
+++ b/app/lib/lets-encrypt.server.ts
@@ -417,7 +417,7 @@ class LetsEncrypt {
 
     this.#order = await this.#client.finalizeOrder(this.#order, csr);
 
-    // fullChain is the publicCertificate + the intermediate certificate(s) for Let's Encrypt
+    // fullChain is the public certificate + the intermediate certificate(s) for Let's Encrypt
     const fullChain = await this.#client.getCertificate(this.#order);
 
     // certificate is the public certificate, chain is the intermediate certificate(s) for Let's Encrypt
--- a/app/lib/lets-encrypt.server.ts
+++ b/app/lib/lets-encrypt.server.ts
@@ -422,13 +422,12 @@ class LetsEncrypt {
 
     // certificate is the public certificate, chain is the intermediate certificate(s) for Let's Encrypt
     const [certificate, ...chainArray] = acme.crypto.splitPemChain(fullChain);
-    const chain = chainArray.join('');
+    const chain = chainArray.join('\r\n');
 
     return {
       privateKey: key.toString(),
       certificate,
       chain,
-      fullChain,
       validFrom: new Date(),
       validTo: new Date(this.#order.expires!),
     };
--- a/app/lib/lets-encrypt.server.ts
+++ b/app/lib/lets-encrypt.server.ts
@@ -421,7 +421,8 @@ class LetsEncrypt {
     const fullChain = await this.#client.getCertificate(this.#order);
 
     // certificate is the public certificate, chain is the intermediate certificate(s) for Let's Encrypt
-    const [certificate, ...chain] = acme.crypto.splitPemChain(fullChain);
+    const [certificate, ...chainArray] = acme.crypto.splitPemChain(fullChain);
+    const chain = chainArray.join('');
 
     return {
       privateKey: key.toString(),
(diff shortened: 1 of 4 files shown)


humphd (commented):
This is looking good to me.

Eakam1007:
I want to test if that array join needs new lines and some other places where some changes might be needed but I ran out of time today

app/lib/lets-encrypt.server.ts
```diff
@@ -416,11 +416,19 @@ class LetsEncrypt {
...
 
     this.#order = await this.#client.finalizeOrder(this.#order, csr);
-    const certificate = await this.#client.getCertificate(this.#order);
+
+    // fullChain is the publicCertificate + the intermediate certificate(s) for Let's Encrypt
+    const fullChain = await this.#client.getCertificate(this.#order);
+
+    // certificate is the public certificate, chain is the intermediate certificate(s) for Let's Encrypt
+    const [certificate, ...chainArray] = acme.crypto.splitPemChain(fullChain);
+    const chain = chainArray.join('');
```
humphd, line 425: Join them with `\r\n`

app/lib/lets-encrypt.server.ts
```diff
@@ -416,11 +416,19 @@ class LetsEncrypt {
...
 
     return {
       privateKey: key.toString(),
       certificate,
+      chain,
+      fullChain,
```
dadolhay, line 431: I think the `fullChain` is redundant data. If I understand correctly, FE (or the certificate model) code would be able to re-join cert and chain at any time to get back the fullChain, very easily, and it is generally not advisable to store redundant data in the db.

dadolhay (changes requested):

app/lib/lets-encrypt.server.ts
humphd: Good point.  Let's synthesize it in the front-end if needed, but not store in db.

Eakam1007: Alright, makes sense

Eakam1007:
@humphd Do you want me to update the UI here as well?

app/lib/lets-encrypt.server.ts
```diff
@@ -416,11 +416,18 @@ class LetsEncrypt {
     });
 
     this.#order = await this.#client.finalizeOrder(this.#order, csr);
-    const certificate = await this.#client.getCertificate(this.#order);
+
+    // fullChain is the publicCertificate + the intermediate certificate(s) for Let's Encrypt
```
humphd, line 420: space after `public` before `Certificate`, lower case `c` in `certificate`

humphd (approved):
One small thing.

I'd do the UI in a follow-up so we can get this landed sooner.

app/lib/lets-encrypt.server.ts
Eakam1007: Weird how I did a typo in the comment. Thanks for catching that

humphd (approved):

humphd:
@cychu42 any issues making this change to the db schema that we're missing?

dadolhay (approved):
Looks nice!

Eakam1007:
> LGTM. The only thing might be seed file being potentially outdated for not having a `chain` value, but it won't affect running seeding since it's an optional field.

I can update the seed file if needed later though I think we only seed for the old starchart-user. I am not sure if we even need that user anymore

PR #533, opened 2023-04-06
Enhc: Adding the ability to request for new certificate

### LINKED ISSUE #513: Allow ordering new cert after failure or when approaching expiry
We need the Cert UI to allow the user to order a new certificate even if they already have one in a few cases:

1. If the certificate is in the error state.  We should show some message to say that it failed, but allow the user to try again
2. If the certificate will expire in the next 30 days, we should let them order a new one

Assigning to @cychu42 and @Ririio.

We need this in 0.9

Ririio:
Fixes #513 

- Adds a button to request a new certificate
  - if certificate expiration date is less than 30 days, it will become available to click
- Adding a function to check for 'pending' status called `getCertificatePendingByUsername`
  - The `getCertificateyByUsername` function only works if the certificate has a `validTo` property, otherwise it ignores and places it a the bottom

### How to test ###

> Pre-requisite
- Make sure you have prisma studio open `npm run db:studio`
- A certificate must already exists under the current user
- `validTo` must be set to be less than 30 days from now

> Initial Testing
1. Request for a new certificate
2. Press the 'repeat-icon' besides the expiration date
3. The pending page would load
4. A new certificate should be displayed
5. The button will now be disabled

### DIFF
--- a/app/components/certificate/certificate-available.tsx
+++ b/app/components/certificate/certificate-available.tsx
@@ -20,6 +20,7 @@ interface CertificateAvailableProps {
   validToFormatted: string;
   publicKey: string;
   privateKey: string;
+  isRenewable: boolean;
 }
 
 export default function CertificateAvailable({
@@ -27,6 +28,7 @@ export default function CertificateAvailable({
   validToFormatted,
   publicKey,
   privateKey,
+  isRenewable,
 }: CertificateAvailableProps) {
   return (
     <>
@@ -35,6 +37,7 @@ export default function CertificateAvailable({
         certRequested={true}
         validFromFormatted={validFromFormatted}
         validToFormatted={validToFormatted}
+        isRenewable={isRenewable}
       />
 
       <Heading as="h2" size="lg" marginTop={4}>
--- a/app/components/certificate/description.tsx
+++ b/app/components/certificate/description.tsx
@@ -1,10 +1,13 @@
-import { Flex, Heading, Text, HStack, VStack } from '@chakra-ui/react';
+import { RepeatIcon } from '@chakra-ui/icons';
+import { Flex, Heading, Text, HStack, VStack, IconButton } from '@chakra-ui/react';
+import { Form } from '@remix-run/react';
 
 interface DescriptionSectionProps {
   certRequested: boolean;
   validFromFormatted?: string;
   validToFormatted?: string;
   description: string;
+  isRenewable?: boolean;
 }
 
 export default function DescriptionSection({
@@ -12,6 +15,7 @@ export default function DescriptionSection({
   validFromFormatted,
   validToFormatted,
   description,
+  isRenewable,
 }: DescriptionSectionProps) {
   return (
     <Flex
@@ -22,9 +26,12 @@ export default function DescriptionSection({
       alignItems={{ base: 'center', md: 'flex-start' }}
       textAlign={{ base: 'center', md: 'left' }}
     >
-      <Heading as="h1" size={{ base: 'lg', md: 'xl' }}>
-        Certificate
-      </Heading>
+      <Flex width="100%">
+        <Heading as="h1" size={{ base: 'lg', md: 'xl' }}>
+          Certificate
+        </Heading>
+      </Flex>
+
       <Text>{description}</Text>
       {certRequested && (
         <HStack gap="6" marginTop="2">
@@ -38,6 +45,20 @@ export default function DescriptionSection({
                 <Text fontWeight="bold">Expires On</Text>
                 <Text>{validToFormatted}</Text>
               </VStack>
+
+              <Flex justifyContent="flex-end">
+                <Form method="post">
+                  <IconButton
+                    type="submit"
+                    aria-label="renew-certificate"
+                    icon={<RepeatIcon />}
+                    backgroundColor="transparent"
+                    color="black"
+                    _hover={{ backgroundColor: 'brand.500', color: 'white' }}
+                    isDisabled={isRenewable ? false : true}
+                  />
+                </Form>
+              </Flex>
             </>
           )}
         </HStack>
--- a/app/components/certificate/description.tsx
+++ b/app/components/certificate/description.tsx
@@ -55,7 +55,7 @@ export default function DescriptionSection({
                     backgroundColor="transparent"
                     color="black"
                     _hover={{ backgroundColor: 'brand.500', color: 'white' }}
-                    isDisabled={isRenewable ? false : true}
+                    isDisabled={!isRenewable}
                   />
                 </Form>
               </Flex>
--- a/app/models/certificate.server.ts
+++ b/app/models/certificate.server.ts
@@ -12,7 +12,7 @@ export async function getCertificateByUsername(username: Certificate['username']
   return prisma.certificate
     .findMany({
       where: { username },
-      orderBy: { validFrom: 'desc' },
+      orderBy: { id: 'desc' },
       take: 1,
     })
     .then(([certificate]) => certificate);
--- a/app/routes/__index/certificate/index.tsx
+++ b/app/routes/__index/certificate/index.tsx
@@ -4,6 +4,8 @@ import { json } from '@remix-run/node';
 import { typedjson, useTypedLoaderData } from 'remix-typedjson';
 import { useRevalidator } from '@remix-run/react';
 import { useInterval } from 'react-use';
+import { useMemo } from 'react';
+import dayjs from 'dayjs';
 
 import { requireUser, requireUsername } from '~/session.server';
 import pendingSvg from '~/assets/undraw_processing_re_tbdu.svg';
@@ -85,6 +87,15 @@ export default function CertificateIndexRoute() {
     certificate?.status === 'pending' ? 15_000 : null
   );
 
+  const canRenew = useMemo((): boolean => {
+    if (certificate?.validTo) {
+      const validTo = dayjs(certificate.validTo!);
+      const thirtyDays = dayjs().add(30, 'day');
+      return validTo.isBefore(thirtyDays);
+    }
+    return false;
+  }, [certificate]);
+
   if (certificate?.status === 'pending') {
     return (
       <Loading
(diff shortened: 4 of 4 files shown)


cychu42:
We went over this together and cover the UI plus back-end code.

app/models/certificate.server.ts
```diff
@@ -18,6 +18,19 @@ export async function getCertificateByUsername(username: Certificate['username']
     .then(([certificate]) => certificate);
 }
 
+export async function getCertificatePendingByUsername(username: Certificate['username']) {
```
dadolhay, line 21: @Ririio 
Instead of this I'd modify `getCertificateByUsername` to get the cert with the highest `id` instead of `validFrom`.
Since `id` is an autoincrement, not null field, certificates will be added in a sequential order (each gets an id 1 larger than the last)

dadolhay: With this change the logic becomes easier, you only have to do 1 sql query

dadolhay (changes requested):

app/models/certificate.server.ts
Ririio: oh what the heck that's true lmfao

cychu42: Good idea.

app/components/certificate/certificate-available.tsx
```diff
@@ -8,13 +8,16 @@ interface CertificateAvailableProps {
...
   privateKey: string;
+  isRenewValid: boolean;
 }
 
 export default function CertificateAvailable({
   validFromFormatted,
   validToFormatted,
   publicKey,
   privateKey,
+
```
Myrfion, line 19: empty line

app/components/certificate/description.tsx
```diff
@@ -38,6 +48,20 @@ export default function DescriptionSection({
                 <Text fontWeight="bold">Expires On</Text>
                 <Text>{validToFormatted}</Text>
               </VStack>
+
+              <Flex justifyContent="flex-end">
+                <Form method="post" onSubmit={() => setIsDisabled(true)}>
```
Myrfion, line 53: Could you use `useTransition`? and check if it is in the submitting state?

https://remix.run/docs/en/1.15.0/hooks/use-transition#transitionstate

app/routes/__index/certificate/index.tsx
```diff
@@ -85,6 +85,19 @@ export default function CertificateIndexRoute() {
...
+    let currentDate = new Date();
+    let validTo = new Date(certificate.validTo!);
+    let thirtyDays = (30 * 24 * 60 * 60 * 1000) / (1000 * 60 * 60 * 24);
+    let differenceInDays = (validTo.getTime() - currentDate.getTime()) / (1000 * 60 * 60 * 24);
+
+    if (differenceInDays <= thirtyDays) {
+      return true;
+    }
+
+    return false;
```
Myrfion, line 98: maybe just 

```typescript
return differenceInDays <= thirtyDays
```

app/components/certificate/description.tsx
Ririio: It should redirect to pending when you click submit so I don't think it's necessary

dadolhay (changes requested):

app/routes/__index/certificate/index.tsx
```diff
@@ -85,13 +85,13 @@ export default function CertificateIndexRoute() {
...
-    return (
-      <Loading
-        img={pendingSvg}
-        desc="We have received your request, and will notify you when your certificate is ready"
-      />
-    );
+  function checkIfRenewValid(): boolean {
+    let currentDate = new Date();
+    let validTo = new Date(certificate.validTo!);
+    let thirtyDays = (30 * 24 * 60 * 60 * 1000) / (1000 * 60 * 60 * 24);
```
cychu42, line 91: Wouldn't it make sense to have this be 30?
Two sides of divide on `(24 * 60 * 60 * 1000) / (1000 * 60 * 60 * 24)` just cancel each other out.
In fact, I think we can probably use 30 instead of a variable.

humphd: You can use dayjs for this too to simplify.

app/routes/__index/certificate/index.tsx
```diff
@@ -85,13 +85,13 @@ export default function CertificateIndexRoute() {
...
   }
 
-  if (certificate?.status === 'pending') {
-    return (
-      <Loading
-        img={pendingSvg}
-        desc="We have received your request, and will notify you when your certificate is ready"
-      />
-    );
+  function checkIfRenewValid(): boolean {
```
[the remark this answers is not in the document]
Ririio: ohh that's a good idea, I'll keep useMemo in mind with re-rendering a calculation multiple times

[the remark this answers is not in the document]
Ririio: BTW... should this also be using `useMemo`?

```
  function formatDate(val: Date): string {
    let date = val.toLocaleDateString('en-US', {
      month: 'short',
      day: '2-digit',
      year: 'numeric',
    });

    return date;
  }
  ```
  Just want to make sure I'm understanding its purpose...

File: app/routes/__index/certificate/index.tsx (hunk 5 above)
Comment (Ririio) [reply app/routes/__index/certificate/index.tsx]: oh great that makes it much easier

File: app/routes/__index/certificate/index.tsx (hunk 6 above)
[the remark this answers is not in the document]
Comment (cychu42) [reply app/routes/__index/certificate/index.tsx]: I think there's no need to do that. The function is called to format date when a value is passed in (line 127-128), while `useMemo` hook triggers when a dependency is updated.  From what I read, it's to [**memo**ized a value](https://www.w3schools.com/react/react_usememo.asp) to avoid having to recalculate it.

However, I notice there's two copies of the same function `formatDate`? On line 68-76 and 90-98.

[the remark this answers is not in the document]
Comment (cychu42) [reply app/routes/__index/certificate/index.tsx]: I think the second `formatDate` on line 90-98 is not needed?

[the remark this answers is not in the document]
Comment (Ririio) [reply app/routes/__index/certificate/index.tsx]: That's odd how did it get duplicated

File: app/routes/__index/certificate/index.tsx [hunk 7]
```diff
@@ -85,6 +77,25 @@ export default function CertificateIndexRoute() {
...
+    let date = val.toLocaleDateString('en-US', {
+      month: 'short',
+      day: '2-digit',
+      year: 'numeric',
+    });
+
+    return date;
+  }
+
+  const checkIfRenewValid = useMemo((): boolean => {
```
dadolhay, line 90: The `valid` and `check` are not the best here.
`Check` indicates that this is a function, while it is not
`valid` indicates validity, while we are checking if we should allow the user to renew or not

Please rename this to `canRenew`

dadolhay (changes requested):

app/routes/__index/certificate/index.tsx
Ririio: Sure sounds reasonable

app/routes/__index/certificate/index.tsx
```diff
@@ -85,6 +77,25 @@ export default function CertificateIndexRoute() {
     certificate?.status === 'pending' ? 15_000 : null
   );
 
+  function formatDate(val: Date): string {
```
dadolhay, line 80: Was there a reason to move this fn into the functional component?
This way this re-generates every time the component renders, and I cannot see a good reason for it. Am I missing something?

If there was no explicit reason for this change, then please move it back

dadolhay (changes requested):

app/routes/__index/certificate/index.tsx
Ririio: There was a duplicate, and I accidentally deleted the original

app/components/certificate/description.tsx
```diff
@@ -1,18 +1,25 @@
...
 }
 
 export default function DescriptionSection({
   certRequested,
   validFromFormatted,
   validToFormatted,
   description,
+  canRenew,
 }: DescriptionSectionProps) {
+  const [isDisabled, setIsDisabled] = useState(canRenew ? false : true);
```
humphd, line 21: This is an anti-pattern in React.  You don't want to use props as the value of state, since the prop will change, and leave the state out of sync.  Below, use `canRenew ? false : true` as your value for the `disabled` attribute.

app/routes/__index/certificate/index.tsx
```diff
@@ -108,6 +119,7 @@ export default function CertificateIndexRoute() {
             privateKey={certificate.privateKey!}
             validFromFormatted={formatDate(certificate.validFrom!)}
             validToFormatted={formatDate(certificate.validTo!)}
+            canRenew={canRenew}
```
humphd, line 122: I would use `isRenewable` or `renewable` as the name vs. `canRenew`

humphd (changes requested):

app/components/certificate/description.tsx
```diff
@@ -38,6 +45,20 @@ export default function DescriptionSection({
...
+              <Flex justifyContent="flex-end">
+                <Form method="post">
+                  <IconButton
+                    type="submit"
+                    aria-label="renew-certificate"
+                    icon={<RepeatIcon />}
+                    backgroundColor="transparent"
+                    color="black"
+                    _hover={{ backgroundColor: 'brand.500', color: 'white' }}
+                    isDisabled={isRenewable ? false : true}
```
humphd, line 58: `isDisabled={!isRenewable}`

app/routes/__index/certificate/index.tsx
```diff
@@ -85,6 +87,15 @@ export default function CertificateIndexRoute() {
     certificate?.status === 'pending' ? 15_000 : null
   );
 
+  const canRenew = useMemo((): boolean => {
```
humphd, line 90: I would rename this to `isRenewable` too, so it matches everything else you're doing.

humphd (changes requested):

app/components/certificate/description.tsx
Ririio: I think I just didn't want to confuse myself so I wrote it like that, I'll change it now

dadolhay (approved):

humphd (approved):

PR #534, opened 2023-04-06
Add an info button linking to DNSChecker

### LINKED ISSUE #440: Replace DNS Record status UI/logic
The work in #396 is going to make it hard for us to continue with our DNS Record status UI and logic.  Instead, we're going to move to a world where we write to the database and then "reconcile" what is on Route53 in bulk--we won't update the status of individual records, since the database will be our source of truth.  As a result, we should fix our UI so that users can still get some info on whether or not their DNS record is fully propagated.

There are a few ways we could do this:

1. Write a Resource Route in Remix that allows us to send a list of domains, and it tries to resolve them server-side, and reports back with whether or not it worked for each one.  This is nice because we control it; however, it's also only reporting back what our server can reach vs. what's happening globally (DNS is distributed around the globe!)
2. Add a button that links users to an external site, and let it do this for us.  For example, I made a domain on staging, and you can test it at https://dnschecker.org/#A/foo.davidhumphrey.stage.mystudentproject.ca or https://www.nslookup.io/domains/foo.davidhumphrey.stage.mystudentproject.ca/dns-propagation/a/.  These sites (there are others, I just picked 2) let you construct a URL and have it do the queries for you using lots of DNS servers around the world.  This is nice because it gives a better picture of what's happening across many DNS servers.

SerpentBytes:
## Description
Closes #440 by adding a button linking to https://dnschecker.org/ to show DNS record info.

### Steps to test
- Run the project locally
- Create a new DNS Record
- Interact with the info button:

https://user-images.githubusercontent.com/50856799/230491653-1c7aec14-0ca0-4aed-8417-00ff04b436a5.mov

### DIFF
--- a/app/components/dns-records-table.tsx
+++ b/app/components/dns-records-table.tsx
@@ -15,9 +15,10 @@ import {
   useDisclosure,
   Spinner,
   HStack,
+  Link,
 } from '@chakra-ui/react';
 import type { DnsRecord } from '@prisma/client';
-import { EditIcon, DeleteIcon, RepeatIcon, CopyIcon } from '@chakra-ui/icons';
+import { EditIcon, DeleteIcon, RepeatIcon, CopyIcon, InfoOutlineIcon } from '@chakra-ui/icons';
 import DnsRecordDeleteAlertDialog from './dns-record-delete-alert-dialog';
 
 import { Form, useNavigate, useTransition } from '@remix-run/react';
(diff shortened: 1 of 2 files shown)


Eakam1007 (dismissed):

humphd (changes requested):
This isn't quite what I had in mind.

If I have 3 records, I want to have this button for *each* of the records, so I can go and look them up individually.  It should also open them in a new tab `target="_blank"`.

I also think we need to consider which icon to use for the button.  Probably something with a circle and `i` for "info" or a `?` etc.

SerpentBytes:
This isn't done. I am still figuring out the spacing.

humphd:
> This isn't done. I am still figuring out the spacing.

@Eakam1007, @Myrfion, and @Ririio are all good at this, if you need a hand.

One thing you might do is use identical components as the others, so they work more alike (I know that this is a link vs. a button, but you can style it the same).

Chakra has `<HStack>` and `<ButtonGroup>` that are helpful for doing the spacing.

app/components/dns-records-table.tsx
```diff
@@ -108,31 +109,33 @@ export default function DnsRecordsTable(props: DnsRecordsTableProps) {
...
+                            <Tooltip label="Check DNS Record">
+                              <IconButton
+                                icon={<InfoOutlineIcon color="black" boxSize="5" />}
+                                aria-label="Check DNS record"
+                                variant="ghost"
+                                ml="2"
+                              />
+                            </Tooltip>
+                          </Link>
+                          <Tooltip label="Copy subdomain to clipboard">
```
humphd, line 127: Let's harmonize our naming/cases.  Above you have a tooltip label of `Check DNS Record` so let's say `Copy DNS Record to clipboard` or something.

humphd (dismissed):
Love this.  Tested locally, and it's great.

Eakam1007 (approved):

PR #536, opened 2023-04-06
Add nginx cert config panel

humphd:
Created another certificate config panel for nginx:

Screenshot 2023-04-06 at 3 07 34 PM

I'm going to need the work from #532 to create the full chain in the UI

### DIFF
--- a/app/components/certificate/certificate-available.tsx
+++ b/app/components/certificate/certificate-available.tsx
@@ -13,6 +13,7 @@ import Description from './description';
 import GeneralPanel from './panels/general';
 import NodePanel from './panels/node';
 import AwsPanel from './panels/aws';
+import NginxPanel from './panels/nginx';
 
 interface CertificateAvailableProps {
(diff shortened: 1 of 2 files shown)


Eakam1007 (commented):
Hmm, just realized that these links are not really obvious (that they are links)
image
On hover:
image
Would it be better to do the same thing as on the landing page for these:
image

humphd:
Can we do some global fix for links so they have text decoration/red colour by default?  It's stupid that we have to fix this in a lot of places.

Eakam1007:
> Can we do some global fix for links so they have text decoration/red colour by default? It's stupid that we have to fix this in a lot of places.

We probably can in another issue, cause we don't want the links in the header to be affected for example. Maybe a shared class

Eakam1007 (approved):
Looks good, we can add the data after the other pr is merged

Genne23v (approved):
Looks good to me

Issue #351, opened 2023-03-15
bug: notification jobs not being added on startup


SerpentBytes:
## Description
Operating system: MacOS

One of the recent changes to the codebase is preventing the `init` function responsible for adding notification-related background jobs from getting triggered. Jobs currently being started via `/dev` are working as intended. It is possible that the `init` function is not being invoked as neither the success nor failure log messages are outputted. 

### Steps to Reproduce
- Run the project
- Look for log messages from the notification-specific `init` function

SerpentBytes:
Update: Exposing `init` and invoking it from the `server.ts` seems to make it work on my machine. The way we had it before.

SerpentBytes:
> Is this still a problem?

The expiration notifications [init](https://github.com/DevelopingSpace/starchart/blob/main/app/queues/notifications/expiration-notification.server.ts#L29) function is not being invoked, so it's still a problem.

SerpentBytes:
This relates to the renewal reminder notifications we roll out a month before expiration.

PR #537, opened 2023-04-06
Move reconciler out of app/, split DNS from Route53 code + test

humphd:
The reconciler has nothing to do with the rest of the web app.  I've isolated it out into its own directory in the root.  I've also split the DNS code into "DNS" and Route53 parts, and added tests for the parts of Route53 that we are using.


humphd:
Also fixes #351.

This modifies things a bit more, creating a `services/` root folder, and moving all non-app background services there.  The whole thing gets started when the server is started.

humphd:
I'm pretty sure I can move most/all of the expiration-request.server.ts and possibly a bunch of the notifications queue stuff under `services/` too, but I haven't had time.  Can happen here or in follow-up.

SerpentBytes (approved):

test/unit/route53-client.server.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+  };
+
+  beforeAll(async () => {
+    hostedZoneId = await createHostedZone(rootDomain);
+    process.env.AWS_ROUTE53_HOSTED_ZONE_ID = hostedZoneId;
+    process.env.ROOT_DOMAIN = rootDomain;
+  });
+
+  afterAll(() => resetHostedZone());
```
dadolhay, line 29: @humphd 
I usually try and do the cleanup in the beforeAll section (before initialization):
If a previous test has failed, was interrupted (ctrl+c), or by any other source, there is some garbage in the system, that the test may or may not be able to tolerate, I want to clean that out before I run the test, so the test can truley be atomic (and not depend on the previous test cleaning up after itself)

Just a suggestion

humphd: I tried that but trashing the hosted zone while while the modules are loaded seems to cause tests to fail.

dadolhay (approved):

PR #540, opened 2023-04-07
process certificate expiration

### LINKED ISSUE #528: Delete expired certificates automatically
We need to delete expired certificates from the database.

@SerpentBytes if this is a duplicate, please close this.  But I think we need it.

SerpentBytes:
## Description
Resolves #528 by deleting expired certificates from the database and sending an email notification.

### Steps to test
- Set the job to repeat every 5 seconds
- Run the project
- Request a certificate
- Run `npm run studio` and set the `validTo` date in the past
- Confirm the expired certificate is deleted from the DB
- Navigate to http://localhost:8025 to see the email notification:
list view

### DIFF
--- a/app/queues/common/expiration-request.server.ts
+++ b/app/queues/common/expiration-request.server.ts
@@ -4,6 +4,11 @@ import { redis } from '~/lib/redis.server';
 import logger from '~/lib/logger.server';
 
 import { getExpiredDnsRecords, deleteDnsRecordById } from '~/models/dns-record.server';
+import {
+  deleteCertificateById,
+  getExpiredCertificates,
+  getCertificateByUsername,
+} from '~/models/certificate.server';
 import { addNotification } from '../notifications/notifications.server';
 
 const { EXPIRATION_REPEAT_FREQUENCY_S, JOB_REMOVAL_FREQUENCY_S } = process.env;
@@ -37,15 +42,15 @@ const expirationRequestWorker = new Worker(
       logger.info('process DNS record expiration');
       let dnsRecords = await getExpiredDnsRecords();
       await Promise.all(
-        dnsRecords.map(async ({ id, user }) =>
+        dnsRecords.map(async ({ id, subdomain, user }) =>
           Promise.all([
-            // delete records from Route53 and DB
+            // delete DNS records from DB
             deleteDnsRecordById(id),
-            // add notification jobs (assuming deletion went successfully)
+            // add notification jobs
             addNotification({
               emailAddress: user.email,
-              subject: 'DNS record expiration subject',
-              message: 'DNS record expiration message',
+              subject: 'My.Custom.Domain DNS record expired',
+              message: `${user.displayName}, your DNS record with subdomain: ${subdomain} has expired.`,
             }),
           ])
         )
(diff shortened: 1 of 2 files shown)


Eakam1007 (changes requested):

Eakam1007 (changes requested):
I think if they can't even see the expired records/certificate, `renew` is kind of misleading as well. Maybe we can remove the entire sentence?

app/queues/common/expiration-request.server.ts
```diff
@@ -37,23 +38,42 @@ const expirationRequestWorker = new Worker(
...
+              message: `${user.displayName}, your DNS record with subdomain: ${subdomain} has expired.`,
             }),
           ])
         )
       );
     } catch (err) {
       throw new UnrecoverableError(`Unable to process DNS record expiration: ${err}`);
     }
-    logger.info('TODO: process certificate expiration');
+    try {
```
dadolhay, line 57: This way, you will also send notification about expired certificate, even if the cert was renewed in the meantime

dadolhay: Those certificates remain in the DB, but we are only ever showing the newest one for the user

dadolhay (changes requested):

app/queues/common/expiration-request.server.ts
SerpentBytes: Thanks for your feedback. 
So we only need to notify the user about the newest one?

SerpentBytes:
@dadolhay I have modified the code to notify only about the most recent and successfully issued certificate.

app/queues/common/expiration-request.server.ts
dadolhay: Yes, if they have already renewed it, then there is no point in notifying them about a cert that has been renewed in the meantime, and they can no longer see in the UI (at least IMO)

app/queues/common/expiration-request.server.ts
```diff
@@ -37,23 +42,41 @@ const expirationRequestWorker = new Worker(
...
+        let mostRecentCertificate = await getCertificateByUsername(username);
+        // add notification jobs
+        if (mostRecentCertificate.id === id) {
+          await addNotification({
+            emailAddress: user.email,
+            subject: 'My.Custom.Domain certificate expired',
+            message: `${user.displayName}, your certificate with domain: ${domain} has expired.`,
+          });
+          // delete certificate from DB
+          await deleteCertificateById(id);
```
dadolhay, line 74: You should still delete it, even if it was not the most recent. If the cert is expired, there is no point in keeping it, it is useless, cannot have any reason to revoke it, etc.

dadolhay (changes requested):

app/queues/common/expiration-request.server.ts
SerpentBytes: That's by accident 😛

dadolhay: :P

dadolhay (approved):

app/queues/common/expiration-request.server.ts
```diff
@@ -37,23 +42,41 @@ const expirationRequestWorker = new Worker(
...
     } catch (err) {
       throw new UnrecoverableError(`Unable to process DNS record expiration: ${err}`);
     }
-    logger.info('TODO: process certificate expiration');
+    try {
+      logger.info('process certificate expiration');
+      let certificates = await getExpiredCertificates();
+      certificates.map(async ({ id, username, domain, user }) => {
+        let mostRecentCertificate = await getCertificateByUsername(username);
+        // add notification jobs
```
humphd, line 66: Can you add a comment to explain this extra check?  I think what I'm reading here is that you'll clean-up old certificates without notifying, but if this is the current certificate, you'll send a notification.  Is that right?  If so, please say that.

humphd (changes requested):
One question and this is good.

humphd (approved):

dadolhay (approved):

Eakam1007 (approved):

PR #541, opened 2023-04-07
Wrap components with body layout

### LINKED ISSUE #488: Inconsistent left and top margins across pages
We should use the same left and top margins across our various pages.  Right now, moving between them feels weird, since things move around.

I think this is possibly happening due to the DNS Records table wanting more width?  We should figure out a solution for consistency.

Screenshot 2023-03-31 at 8 37 55 AM

Genne23v:
Close #488 

Add a layout for consistent page look. 

https://user-images.githubusercontent.com/59734889/230536942-95076926-09fe-4bff-ad66-c670e5b1f8d5.mp4

### DIFF
--- a/app/components/admin/admin-metric-card.tsx
+++ b/app/components/admin/admin-metric-card.tsx
@@ -16,7 +16,7 @@ export default function AdminMetricCard({
   tooltipText,
 }: AdminMetricCardProps) {
   return (
-    <Card w={{ sm: '100%', md: 300 }} mr={{ sm: '0', md: '4' }} mt="4">
+    <Card w={{ sm: '100%', md: 300 }} minW={270} maxW={400} mr={{ sm: '0', md: '4' }} mt="4">
       <CardBody flexDirection="row" display="flex" justifyContent="space-between">
         <Flex alignItems="center">
           <Flex
--- a/app/components/certificate/certificate-available.tsx
+++ b/app/components/certificate/certificate-available.tsx
@@ -43,12 +43,12 @@ export default function CertificateAvailable({
       <Heading as="h2" size="lg" marginTop={4}>
         Using Your Certificate
       </Heading>
-      <Text>
+      <Text maxW={750}>
         Your <em>certificate</em> is made-up of a number of separate parts, and different services
         will require you to use them in various combinations. These parts include:
       </Text>
 
-      <OrderedList>
+      <OrderedList maxW={750}>
         <ListItem>
           <strong>Public Certificate</strong>: public (not secret) certificate that verifies domain
           ownership
@@ -66,12 +66,12 @@ export default function CertificateAvailable({
         </ListItem>
       </OrderedList>
 
-      <Text>
+      <Text maxW={750}>
         You can select from the list of pre-defined services below to help guide you, or see a
         General view for all of the certificate's parts.
       </Text>
 
-      <Tabs>
+      <Tabs maxW={750}>
         <TabList>
           <Tab>General</Tab>
           <Tab>Node.js</Tab>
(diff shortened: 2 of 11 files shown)


humphd:
This is excellent.  I've tried it locally, and it looks and works great.

One thing I'd ask you to do is to constrain the max width of paragraphs of text.  When a paragraph of text is longer than ~50-75 characters, it becomes hard to read.  See https://baymard.com/blog/line-length-readability.

We could do that in a follow-up vs. blocking this if you think it's a lot of work.

Genne23v:
Thank you for your feedback. I pushed the change that sets the text max width. 

One drawback is though that we have a lot of white space on the right. 

Screenshot 2023-04-07 at 9 48 25 AM

Myrfion (dismissed):

Ririio:
> Thank you for your feedback. I pushed the change that sets the text max width.
> 
> One drawback is though that we have a lot of white space on the right.

That's odd... why did the width of the container lessen? I thought he's only referring to the size of the Text component being smaller not the entire container

humphd (changes requested):
We shouldn't do this to the certificate display, as it wraps the lines

Screenshot 2023-04-07 at 11 44 09 AM

I wonder if you should make the max width a bit wider?

app/components/body-layout.tsx
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export default function BodyLayout({ heading, children }: LayoutProps) {
+  return (
+    <Container maxW="container.xl" px={{ base: 4, md: 45 }}>
+      <Heading as="h1" size={{ base: 'lg', md: 'xl' }} mt={{ base: 6, md: 12 }}>
+        {heading}
+      </Heading>
+      {children}
+    </Container>
```
Eakam1007, line 16: You could put this `Container` around the `Outlet` in [app/routes/_index.tsx](https://github.com/DevelopingSpace/starchart/blob/7bb2b2652f9b8c490f5cd43f919e85d508a8434c/app/routes/__index.tsx) instead and have it render automatically instead of having to import this component on every index route.
Then, the `Heading` can stay in its own respective page.

Eakam1007 (changes requested):

app/components/body-layout.tsx
humphd: Do we want to have tables go wider than a `Container` (admin, dns records)?  Or should we constrain that too?

Eakam1007: Do you mean have the tables go to different widths than the text?
I think we can constrain the text separately as needed, and this will constrain the entire page to have consistent sizes across all pages

humphd: Right: do we want the tables to be wider than a Container?  Maybe not, since we can put max widths on the columns too and have it all fit without scrolling.

Eakam1007: Yea, that makes sense. I think the container size  (`xl`) is big enough that we won't need to do anything different for the tables

Genne23v: I increased all width on certificate page and replaced the layout template with wrapping root index with Container. Please review if everything is good. Thanks!

humphd (dismissed):

Eakam1007:
Is this intended behavior? The domains table vs create domain page have different margins on the left side:  

image


Then, there's this text on the domains table. I feel like the text should have the same width as the table:  

image

Genne23v:
Well, I initially thought that forms can be closer to the centre as it has a much narrow width, but I think it makes more sense to be aligned with other pages as Certificate page is similar to the form. 

I just pushed another commit for these form layout. Please review. 

Thank you for checking these pages!

humphd (approved):

Eakam1007 (approved):

Issue #442, opened 2023-03-25
Sort DNS Records such that multiple records with same name are grouped


humphd:
In #438 we are moving to allow DNS Records of type `A`, `AAAA`, and `TXT` (i.e., not `CNAME`) to have multiple values.  Our current UI assumes this is always a single value.  We need to make 2 changes:

1. DNS Records table needs to be able to show one or more values.  We could join them with `,` for example
2. The New DNS Record and Edit DNS Record form needs to allow for one or more input boxes for the value(s).  You should be able to add or remove input boxes to include multiple values.  By default we'd show only 1.

Myrfion:
@cychu42 sorry just seen your message on this one, the issue got resolved in your PR right?

humphd:
When a user has more than 1 record with the same name and different type, it would be nice to "group" them together by sorting the results:

<img width="1233" alt="Screenshot 2023-04-10 at 11 20 45 AM" src="https://user-images.githubusercontent.com/427398/230931611-77c79a8c-1fe2-4e3a-a8c4-1a0699762df1.png">

dadolhay:
```js
const recordsSorted = records.sort(
  (({subdomain: aSubdomain, type: aType}, {subdomain: bSubdomain, type: btype}) =>
    [...aSubdomain].reverse().join('').localeCompare([...bSubdomain].reverse().join('')) ||
    atype.localeCompare(btype)
  );
```
- reverse both subdomains
- compare them
- If equal, then compare the record type the same way

This is not tested, you might need to change the compare order (like compare b with a instead of a with b)

cychu42:
> @cychu42 sorry just seen your message on this one, the issue got resolved in your PR right?

I only added restriction to not duplicate CNAME records.

PR #542, opened 2023-04-07
Prevent duplicate CNAME for a user

cychu42:
As part of #442 
This prevents duplicate CNAME records from the same user with the same name/`subdomain`.

## Test
From our website, you can try to create multiple CNAME records with the same name/`subdomain` as the same user.
You should see an error that says "DNS Record already exists".

### DIFF
--- a/app/models/dns-record.server.ts
+++ b/app/models/dns-record.server.ts
@@ -112,25 +112,14 @@ export async function doesDnsRecordExist(
   data: Pick<DnsRecord, 'username' | 'type' | 'subdomain' | 'value'>
 ) {
   const { username, type, subdomain, value } = data;
-  let count = 0;
-  if (type === 'CNAME') {
-    count = await prisma.dnsRecord.count({
-      where: {
-        username,
-        type,
-        subdomain,
-      },
-    });
-  } else {
-    count = await prisma.dnsRecord.count({
-      where: {
-        username,
-        type,
-        subdomain,
-        value,
-      },
-    });
-  }
+  const count = await prisma.dnsRecord.count({
+    where: {
+      username,
+      type,
+      subdomain,
+      value: type === 'CNAME' ? undefined : value,
+    },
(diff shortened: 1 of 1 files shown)


app/models/dns-record.server.ts
```diff
@@ -112,14 +112,25 @@ export async function doesDnsRecordExist(
...
-  const count = await prisma.dnsRecord.count({
-    where: {
-      username,
-      type,
-      subdomain,
-      value,
-    },
-  });
+  let count = 0;
+  if (type === 'CNAME') {
```
Genne23v, line 116: I think we can achieve same with this code
```js
const count = await prisma.dnsRecord.count({
    where: {
      username,
      type,
      subdomain,
      value: type === 'CNAME' ? undefined : value,
    },
  });
```

Genne23v (changes requested):

app/models/dns-record.server.ts
cychu42: Good idea!

SerpentBytes (dismissed):

app/models/dns-record.server.ts
```diff
@@ -117,7 +117,7 @@ export async function doesDnsRecordExist(
       username,
       type,
       subdomain,
-      value,
+      value: type === 'CNAME' ? undefined : value,
```
dadolhay, line 120: It took me a minute to understand what you are doing here :)

Please add some comments what you are doing for the next person coming this way

dadolhay (changes requested):

app/models/dns-record.server.ts
cychu42: Sure. Done!

dadolhay (approved):

Genne23v (approved):

cychu42:
Thanks for the reviews.

PR #547, opened 2023-04-08
Add impersonation logic and require admin for admin page

### LINKED ISSUE #508: Make impersonation feature work
## Description
Figure out how to make the impersonation feature work.

sfrunza13:
Closes #508 

To test use an admin account such as user3 user3pass on local and press the mask next to a user's name on the user table in admin page. 

This will redirect you to root and show in the header whom you are currently mocking.

The header will also display a new theater mask with a line through it indicating that should you click it it will stop the impersonation and return you to your original account.

Logging our of the application in either fashion (SLO or not) will also terminate impersonation.

To test require admin on the admin page log onto the application with a non-admin account and try to access the admin page.


app/components/header.tsx
```diff
@@ -94,8 +99,25 @@ export default function Header() {
...
         </Hide>
+        {user.username !== originalUser.username && (
+          <Tooltip label="Revert to original user">
+            <Form method="post">
+              <input type="hidden" name="originalName" value={originalUser.username} />
+              <IconButton
+                type="submit"
+                aria-label="Revert to original user"
+                icon={<TbMasksTheaterOff color="white" size={24} />}
+                variant="ghost"
```
SerpentBytes, line 116: I ran it locally, and I think it's hard to see the button when hovering.

app/components/header.tsx
```diff
@@ -94,8 +99,25 @@ export default function Header() {
 
       <Flex justifyContent="flex-end" alignItems="center" color="white" gap="5" width="100%">
         <Hide below="lg">
-          <Text id="header-user">{user?.username}</Text>
+          <Text id="header-user">
+            {user.username === originalUser.username
+              ? user?.username
+              : `Mocking: ${user?.username}`}
```
humphd, line 105: I'm not sure "Mocking" is the right term here.  We could use "Logged in As:" or "Impersonating:"

app/components/header.tsx
```diff
@@ -94,8 +99,25 @@ export default function Header() {
...
+          </Text>
         </Hide>
+        {user.username !== originalUser.username && (
+          <Tooltip label="Revert to original user">
```
humphd, line 109: I wonder if a better place for this is in the sub-menu where we have "Sign-out".  So you can "Stop Impersonating" or "Sign Out" (2 options).  Then it wouldn't be so cluttered in the header.

Another idea: change the icon for the User Avatar to the Impersonate masks, to indicate that's what's happening?

humphd (changes requested):
I tested this locally, and it works great.  I think we should iterate a bit more on the UI.  I left some ideas.

app/components/header.tsx
```diff
@@ -94,16 +98,36 @@ export default function Header() {
...
+          </Text>
         </Hide>
 
         <Menu>
           <MenuButton
             as={IconButton}
-            icon={<Avatar bg="brand.500" showBorder={true} borderColor="white" size="sm" />}
+            icon={
+              user.username !== originalUser.username ? (
+                <FaTheaterMasks size="35" />
```
humphd, line 113: This doesn't seem to be working for me locally:

Screenshot 2023-04-08 at 3 33 00 PM

humphd (approved):
Looks good.  Just the icon switch is not working for me.  Not sure why?

SerpentBytes (approved):

app/components/header.tsx
sfrunza13: I'm unsure why this is, which account did you log into to impersonate user3?

humphd: I was logged in as `han.solo`.

sfrunza13: I can't seem to reproduce it

humphd: Can you put up some screenshots?  It's likely something local for me.  If it works for you, we can ship to staging and test there.

sfrunza13: image
image

PR #550, opened 2023-04-09
Bump version to 0.9

SerpentBytes:
## Description
Bump project version to 0.9


humphd (changes requested):
Should this not also update `package-lock.json` at the same time?

Use `npm version` to do it.

PR #557, opened 2023-04-10
Refactor certificate ui and model to use fullChain

humphd:
This builds on the work @Eakam1007 did to add the `chain` to the Certificate model.  I'm dynamically creating the `fullChain` in the model when possible, and using it in the front-end.

I've also consolidated the certificate download code, so that it works for all 4 certificate parts we have now with a single route endpoint.


app/components/certificate/certificate-available.tsx
```diff
@@ -8,28 +8,37 @@ import {
...
 
 export default function CertificateAvailable({
+  certificate,
   validFromFormatted,
   validToFormatted,
-  publicKey,
-  privateKey,
-  isRenewable,
 }: CertificateAvailableProps) {
+  const isRenewable = useMemo((): boolean => {
```
sfrunza13, line 33: Never seen this used before, nice.

Myrfion (approved):
Honestly don't even have anything to add, really liked the download route solution, very remix kinda thing to do

sfrunza13 (approved):
This is very nice, I learned some stuff. Making components for a few different types of cert panels that all use the same CertDisplay component to share download logic.

PR #563, opened 2023-04-10
#249

### LINKED ISSUE #249: Automate release at the end of milestone

sfrunza13:
fixes #249 
If I am correct, the permissions only have to be there so long as we are testing on forks because they are downgraded for some security reason. Once this is on main I think we can remove the permissions bit at the top and it will still work. 

image
image

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,6 +5,10 @@ on:
     tags:
       - 'v*'
 
+permissions:
+  actions: write
+  contents: write
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,368 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  release:
+    name: 'Release'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: 'marvinpinto/action-automatic-releases@latest'
+        with:
+          repo_token: '${{ secrets.GITHUB_TOKEN }}'
+          prerelease: false
+  Build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up node 18
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: npm
+
+      - name: Restore node_modules cache
+        id: node_modules-cache-restore
+        uses: actions/cache/restore@v3
+        with:
+          # path to node_modules to cache
+          path: ./node_modules
+          # cache module with hash of package-lock.json
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+
+      - name: Install node_modules
+        # install node modules manually only if cache was not restored
+        if: steps.node_modules-cache-restore.outputs.cache-hit != 'true'
+        run: npm ci
+
+      - name: Run build
+        run: npm run build
+
+      - name: Cache node_modules
+        uses: actions/cache@v3
+        with:
+          path:
+            ./node_modules
+            # cache node modules using the same key as restore.
+          key: ${{ steps.node_modules-cache-restore.outputs.cache-primary-key }}
+
+  Prettier:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Check prettier
+        uses: creyD/prettier_action@v4.3
+        with:
+          prettier_options: --check .
+
+  Unit-Tests:
+    runs-on: ubuntu-latest
+    services:
+      mysql:
+        image: mysql:8
+        ports:
+          - 3306:3306
+        env:
+          MYSQL_DATABASE: starchart_test
+          MYSQL_USER: starchart
+          MYSQL_PASSWORD: starchart_password
+          MYSQL_ROOT_PASSWORD: root_password
+      redis:
+        image: redis:7.0.8-alpine3.17
+        ports:
+          - 6379:6379
+      dns:
+        image: motoserver/moto:4.1.1
+        ports:
+          - '5053:5000'
+      mailhog:
+        image: jcalonso/mailhog:latest
+        ports:
+          - 8025:8025
+          - 1025:1025
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up node 18
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,372 @@
...
+permissions:
+  actions: write
+  contents: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  release:
```
SerpentBytes, line 17: Is "release" all lowercase due to naming conflict?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,372 @@
...
+    name: 'Release'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: 'marvinpinto/action-automatic-releases@latest'
+        with:
+          repo_token: '${{ secrets.GITHUB_TOKEN }}'
+          prerelease: false
+  Build:
```
humphd, line 25: I don't think we need to do all these build checks, since the commit has already been tagged.  That is, we won't tag a commit that isn't passing CI.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,372 @@
...
+            ./node_modules
+            # cache node modules using the same key as restore.
+          key: ${{ steps.node_modules-cache-restore.outputs.cache-primary-key }}
+
+  Docker-Build-Deploy:
+    # We'll only build and push when landing on main, not for PRs
+    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+    runs-on: ubuntu-latest
+    # Don't bother unless everything else passes
+    needs: [Build, Prettier, Unit-Tests, E2E-Tests, Dockerfile-Lint, ESLint, Type-Check]
```
humphd, line 333: You can delete and skip all these, and just go straight to the Docker bit.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,372 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push Starchart Docker image
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          build-args: |
+            SAML_IDP_METADATA_PATH=./config/idp-metadata-staging.xml
+          push: true
+          tags: ${{ env.IMAGE }}:sha-${{ env.GITHUB_SHA_SHORT }}, ${{ env.IMAGE }}:main, ${{ env.IMAGE }}:latest
```
humphd, line 363: Instead of `:main`, let's use `:release` as our generic "this is a release image" tag (i.e., `:main` means, "this is a commit on main")

.github/workflows/release.yaml
```diff
@@ -0,0 +1,372 @@
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push Starchart Docker image
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          build-args: |
+            SAML_IDP_METADATA_PATH=./config/idp-metadata-staging.xml
```
humphd, line 361: `-staging` seems wrong here.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,372 @@
...
+          push: true
+          tags: ${{ env.IMAGE }}:sha-${{ env.GITHUB_SHA_SHORT }}, ${{ env.IMAGE }}:main, ${{ env.IMAGE }}:latest
+
+      - name: Deploy to Staging via Webhook
+        uses: navied/secure-actions-webhook@0.2.1
+        env:
+          HMAC_SECRET: ${{ secrets.WEBHOOK_STAGING_SECRET }}
```
humphd, line 368: `_PRODUCTION_`, I've already created the secret.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,372 @@
...
+        with:
+          url: https://mycustomdomain-dev.senecacollege.ca/hooks/deploy
```
humphd, line 370: drop `-dev`

humphd (changes requested):

.github/workflows/release.yaml
sfrunza13: do we have a prod metadata?

.github/workflows/release.yaml
sfrunza13: I don't know if there was a naming conflict, I had just made it lower case since I had seen examples using this action where the job names were lower case.

Eakam1007:
Could you please add an appropriate title to summarize what this does?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,63 @@
...
+    name: 'Release'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: 'marvinpinto/action-automatic-releases@latest'
+        with:
+          repo_token: '${{ secrets.GITHUB_TOKEN }}'
+          prerelease: false
+
+  Docker-Build-Deploy:
+    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
```
humphd, line 23: Do we need this?  This whole workflow should only run on tags already.

humphd (changes requested):

.github/workflows/release.yaml
sfrunza13: I don't think we need it, I kept it because I don't see the benefit of starting this if tags are pushed to anything but main.

humphd: Add the `main` branch as an additional requirement at the top where you do the `v**` bit

.github/workflows/release.yaml
```diff
@@ -0,0 +1,64 @@
...
+          tags: ${{ env.IMAGE }}:sha-${{ env.GITHUB_SHA_SHORT }}, ${{ env.IMAGE }}:release, ${{ env.IMAGE }}:latest
+
+      - name: Deploy to Staging via Webhook
+        uses: navied/secure-actions-webhook@0.2.1
+        env:
+          HMAC_SECRET: ${{ secrets.WEBHOOK_PRODUCTION_SECRET }}
+        with:
+          url: https://mycustomdomain.senecacollege.ca/hooks/deploy
+          hmacSecret: ${{ env.HMAC_SECRET }}
+          data: '{"tag": "sha-${{ env.GITHUB_SHA_SHORT }}"}'
```
humphd, line 64: Let's use a tag of `"${{ env.VERSION_TAG }}"` instead of `sha-...` so we deploy the `vX.Y.Z` tag (easier to correlate back to a release that way).

humphd (changes requested):
I think it is looking good.  One more thing that occurs to me.

humphd (approved):
I think this is good.

Once we land this, we can do a test with a few tags this week.  Ping me when you have time, I have to do a couple of things to production to make it work.

SerpentBytes (approved):

PR #566, opened 2023-04-11
Restructure expiration notification worker code

### LINKED ISSUE #11: Notify users about SSL certificate expiration
To help keeping users informed of their SSL certificate expiration, we should send them an email when that happens. The email can contain further information on how to renew their certificates.

The code can check for expiration date for each active certificate in the database and notify the owners of any newly expired certificate. This can be done once per day.

SerpentBytes:
## Description
Closes #11. The code to process expiration notifications was already implemented in a previous PR. This PR restructures the code slightly to add context to the reminder notices we roll out a month before DNS records and certificate expiration. We are also checking if the status is "issued" for certificates prior to notifying.

### Steps to Test:
- Run the project locally
- Modify the `validTo` (for Certs) and `expiresAt` for (DNS records)
- Navigate to http://localhost:8025 to view notification after 5 mins:
list view
detailed view

### DIFF
--- a/services/expiration/expiration-notification.server.ts
+++ b/services/expiration/expiration-notification.server.ts
@@ -53,34 +53,30 @@ export function init() {
         logger.debug(`Notifications: processing job ${job.name}`);
         let dnsRecords = await getExpiringDnsRecords();
         await Promise.all(
-          dnsRecords.map(async ({ id, subdomain, expiresAt, lastNotified, user }) => {
-            if (!lastNotified || lastNotified < dayjs().subtract(30, 'd').toDate()) {
-              await updateStatusAndNotify(type, id, {
-                emailAddress: user.email,
-                subject: 'My.Custom.Domain DNS record approaching expiration',
-                message: `${
-                  user.displayName
-                }, this is a friendly reminder that your DNS record with subdomain: ${subdomain} will expire on: ${expiresAt.toLocaleDateString(
-                  'en-CA'
-                )}. Log in to My.Custom.Domain to renew.`,
-              });
-            }
+          dnsRecords.map(async ({ id, subdomain, expiresAt, user }) => {
+            await updateStatusAndNotify(type, id, {
+              emailAddress: user.email,
+              subject: 'My.Custom.Domain DNS record approaching expiration',
+              message: `${
+                user.displayName
+              }, this is a friendly reminder that your DNS record with subdomain: ${subdomain} will expire on: ${expiresAt.toLocaleDateString(
+                'en-CA'
+              )}. Log in to My.Custom.Domain to renew.`,
+            });
           })
         );
         let certificates = await getExpiringCertificates();
         await Promise.all(
-          certificates.map(async ({ id, domain, validTo, lastNotified, user }) => {
-            if (!lastNotified || lastNotified < dayjs().subtract(30, 'd').toDate()) {
-              await updateStatusAndNotify(type, id, {
-                emailAddress: user.email,
-                subject: 'My.Custom.Domain certificate approaching expiration',
-                message: `${
-                  user.displayName
-                }, this is a friendly reminder that your certificate with domain: ${domain} will expire ${
-                  validTo ? `on: ${validTo.toLocaleDateString('en-CA')}` : `in less than 30 days.`
-                }. Log in to My.Custom.Domain to renew. `,
-              });
-            }
+          certificates.map(async ({ id, domain, validTo, user }) => {
+            await updateStatusAndNotify(type, id, {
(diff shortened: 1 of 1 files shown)


services/expiration/expiration-notification.server.ts
```diff
@@ -137,57 +164,48 @@ const updateNotificationStatus = (type: RecordType, id: number) => {
...
-              lastNotified: dayjs()
-                .subtract(NOTIFICATION_FREQUENCY * 4, 'd')
-                .toDate(),
-            },
-          ],
+const getExpiringCertificates = () => {
+  return prisma.certificate.findMany({
+    where: {
+      validTo: {
+        lte: dayjs().add(1, 'M').toDate(),
```
humphd, line 171: Sometimes you use `30 days` and sometimes `1 month`.  Should you do the same everywhere?

services/expiration/expiration-notification.server.ts
```diff
@@ -60,18 +60,45 @@ export function init() {
...
-              await updateStatusAndNotify(type, record.id, {
-                emailAddress: record.user.email,
-                subject: `Sample ${type} expiration notice`,
-                message: `Sample ${type} expiration notice message`,
-              });
-            }
-          })
-        );
+        switch (type) {
+          case RecordType.DnsRecord:
```
dadolhay, line 64: Is there a reason why we are splitting this up to dnsRecord and certificate?
TBH it feels like complexity that we don't need:
- If we want to handle them separately => we should do it in a separate queue
- If we want to handle them together => we should do all of that in a single job
I'd be OK with either, but this splitting the same worker ... I just don't see the benefit, only the added complexity

Id drop this distinction and run them both on the same queue run. Or is there something I'm missing?

dadolhay (changes requested):

services/expiration/expiration-notification.server.ts
SerpentBytes: Thanks for your feedback. Agreed, it's better to process Cert/DNS notifications in the same queue without added complexity.

humphd (approved):

humphd (approved):

sfrunza13 (dismissed):

sfrunza13 (commented):
It looks good, I am just wondering whether you need the conditions in  the worker on lines 66 and 83, aren't those the same as the OR condition in the previous function calls to actually retrieve the records and certs from the db?

Edit: I didn't mean for this comment to dismiss approval, my bad.
Edit 2: nvm I dont think it did

SerpentBytes:
> Edit: I didn't mean for this comment to dismiss approval, my bad. Edit 2: nvm I don't think it did Edit 3: What a nice day outside, finally some good weather

As amazing it can get. 😄

services/expiration/expiration-notification.server.ts
```diff
@@ -14,11 +14,11 @@ import type { NotificationData } from '~/queues/notifications/notifications.serv
...
-export enum RecordType {
+enum RecordType {
   Certificate = 'certificate',
-  DnsRecord = 'record',
+  DnsRecord = 'dns-record',
 }
-export interface ExpirationStatusPayload {
+interface ExpirationStatusPayload {
   type: RecordType;
 }
```
dadolhay, line 23: I don't think that you need type ... or any payload for that matter.

dadolhay (changes requested):

humphd (dismissed):

services/expiration/expiration-notification.server.ts
```diff
@@ -54,24 +45,45 @@ export function init() {
...
+        let certificates = await getExpiringCertificates();
         await Promise.all(
-          records.map(async (record) => {
-            if (!record.lastNotified || record.lastNotified < dayjs().subtract(30, 'd').toDate()) {
-              await updateStatusAndNotify(type, record.id, {
-                emailAddress: record.user.email,
-                subject: `Sample ${type} expiration notice`,
-                message: `Sample ${type} expiration notice message`,
+          certificates.map(async ({ id, domain, validTo, lastNotified, user }) => {
+            if (!lastNotified || lastNotified < dayjs().subtract(30, 'd').toDate()) {
```
sfrunza13, line 73: Is this needed here when we also check a similar or condition in getExpiringCertificates?

SerpentBytes: I was hesitant to remove it because it was added in a previous PR. Great catch.

sfrunza13 (approved):

humphd (approved):
Nice simplification

dadolhay (approved):

PR #567, opened 2023-04-11
fix: Certificate show/hide font-size and wrapping problem

### LINKED ISSUE #560: Font size and wrapping issues with certificate show/hide
The font-size we use for the certificate show/hide view is too large (I think it should use `sm`):

Screenshot 2023-04-10 at 3 35 30 PM

Also, the wrapping is no good on narrow displays (we shouldn't wrap those lines):

Screenshot 2023-04-10 at 3 35 38 PM

SerpentBytes:
## Description
Potentially resolves #560.


SerpentBytes:
I am unsure if I got it right, so I am marking it as a draft.

PR #568, opened 2023-04-11
Tweaks to DNS and Cert instructions pages and styles

humphd:
Playing with the two "instruction" pages locally, I noticed some things I wanted to improve.

### DIFF
--- a/app/routes/__index/certificate/information.tsx
+++ b/app/routes/__index/certificate/information.tsx
@@ -22,21 +22,40 @@ export default function CertificateInstructionsRoute() {
   return (
     <Flex flexDirection="column" paddingBottom="20" maxWidth={750}>
       <Heading as="h1" size={{ base: 'lg', md: 'xl' }} mt={{ base: 6, md: 12 }} mb="4">
-        What is an SSL Certificate?
+        SSL Certificates
       </Heading>
       <Box>
-        <Text>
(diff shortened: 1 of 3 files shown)


app/routes/__index/certificate/information.tsx
```diff
@@ -22,90 +22,81 @@ export default function CertificateInstructionsRoute() {
...
-        </Text>
-
-        <Heading as="h3" size="md" marginTop="5">
-          Browser Security Warnings
+        <Heading as="h2" size="md" marginTop="5">
+          What is an SSL Certificate?
         </Heading>
-
-        <Text marginTop="3">
+        <Text marginY="1em">
```
Myrfion, line 31: I am curious why you decided to use "1em" instead of "4" which I think is the same thing in terms of it's value. Probably would be better to use units we use everywhere else

Myrfion: Update: "4" is not 1em, it is 1rem, which as far as I understand doesn't make any difference in the context of our app

humphd: I can switch to use `{4}` if you want, sure.

Ririio (approved):
Aside from what Tymur has mentioned, I don't see any issue

Myrfion (approved):
LGTM

PR #570, opened 2023-04-11
Sort user records in model

### LINKED ISSUE #442: Sort DNS Records such that multiple records with same name are grouped

Genne23v:
Resolve #442 

I added `orderBy` instead of handling in frontend. I'm not sure if I'm missing something with this approach. Please let me know if there is any concern with this. Thanks

### DIFF
--- a/app/models/dns-record.server.ts
+++ b/app/models/dns-record.server.ts
@@ -15,6 +15,17 @@ export function getDnsRecordsByUsername(username: DnsRecord['username']) {
         subdomain: '_acme-challenge',
       },
     },
+    orderBy: [
+      {
+        subdomain: 'asc',
+      },
+      {
+        type: 'asc',
+      },
+      {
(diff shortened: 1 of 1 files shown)


cychu42:
I can see using this PR to set this ordering as the default behavior, and then add the sorting I mentioned in a different issue later.

humphd (approved):

cychu42 (approved):

PR #576, opened 2023-04-12
Add E2E tests for DNS Records

### LINKED ISSUE #436: Add E2E tests for actions on records
Need E2E tests for record creation and editing that would include cases where user gets error because some values are invalid and successful cases

Eakam1007:
Resolves #436 

### Changes made
- Added tests for valid value and type combinations when creating DNS Records
- Added tests for invalid value and type combinations when creating DNS Records
- Added tests for editing and deleting a DNS Record
- Added common functions to fill in DNS Record form data and check the edit page for correct values

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -25,7 +25,13 @@ const config: PlaywrightTestConfig = {
    * allowing easier debugging without pushing to a branch or modifying this manually
    */
   retries: process.env.CI ? 2 : 1,
-  /* Opt out of parallel tests on CI. */
+  /**
+   * Opt out of parallel tests
+   * This is needed as we want to delete records between tests
+   * When running in parallel, the tests start running before the deletion is complete
+   * The alternative would be to make tests depend on each other, i.e., use different test values
+   * for each test
+   */
   workers: 1,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   reporter: 'html',


test/e2e/new-dns-record.spec.ts
```diff
@@ -16,23 +17,32 @@ test.describe('authenticated as user', () => {
...
+
   test('redirects to edit DNS record page when required fields are filled', async ({ page }) => {
-    await page.getByLabel('Name*').fill('test');
+    await page.getByLabel('Name*').fill('required-test');
     await page.getByRole('combobox', { name: 'Type' }).selectOption('A');
-    await page.getByLabel('Value*').fill('test');
+    await page.getByLabel('Value*').fill('192.168.1.1');
     await page.getByRole('button', { name: 'Create' }).click();
-    await expect(page).toHaveURL(/.*dns-records\/[0-9]*/);
+    await expect(page).toHaveURL('/dns-records');
```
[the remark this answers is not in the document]
Eakam1007: Yea I missed that, thanks for pointing that out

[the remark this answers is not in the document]
Eakam1007: I have added a check by navigating to the edit dns record page after creation

test/e2e/new-dns-record.spec.ts
```diff
@@ -16,27 +20,93 @@ test.describe('authenticated as user', () => {
...
+   * Clean up after all tests are run
+   * For some weird reason using afterEach to delete the one record doesn't work
+   * The test is run before the records are deleted
+   * Thus, we are using different values for the fields for the tests
+   */
+  test.afterAll(async () => {
+    await prisma.dnsRecord.deleteMany();
+  });
+
+  interface DnsRecordData {
```
humphd, line 33: Why not use the type from Prisma?  If you can't use it as is, at least consider using it's internal types (e.g., `DnsRecord['value']` vs. `string` etc)

humphd (changes requested):
Nice, one suggestion.

test/e2e/new-dns-record.spec.ts
Eakam1007: I actually forgot why I did that. I will try changing it to use the Prisma type

test/e2e/new-dns-record.spec.ts
```diff
@@ -16,27 +20,93 @@ test.describe('authenticated as user', () => {
...
-    await page.getByLabel('Value*').fill('test');
+  /**
+   * Clean up after all tests are run
+   * For some weird reason using afterEach to delete the one record doesn't work
+   * The test is run before the records are deleted
+   * Thus, we are using different values for the fields for the tests
+   */
+  test.afterAll(async () => {
+    await prisma.dnsRecord.deleteMany();
+  });
```
Eakam1007, line 31: This is caused locally due to a race condition as we run tests in parallel using multiple workers. A fix is to fix the number of workers locally to 1 to simulate what happens in CI.

test/e2e/new-dns-record.spec.ts
Eakam1007: I have resolved this and done a bunch of refactoring that also removed the need to have unique names for different records (https://github.com/DevelopingSpace/starchart/pull/576#discussion_r1165855497)

playwright.config.ts
```diff
@@ -26,7 +26,7 @@ const config: PlaywrightTestConfig = {
    */
   retries: process.env.CI ? 2 : 1,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  workers: process.env.CI ? 1 : 1,
```
humphd, line 29: Could we not simply say `workers: 1,`?

humphd (changes requested):
Excellent.  One nit.

Eakam1007:
I remembered that we also don't have edit and delete tests so I added a new file and moved some functions to a common file

playwright.config.ts
```diff
@@ -26,7 +26,7 @@ const config: PlaywrightTestConfig = {
    */
   retries: process.env.CI ? 2 : 1,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
```
humphd, line 29: It would be worth updating this comment to indicate that our tests depend on being run one at a time due to database patterns.

humphd (approved):
Great stuff.  One suggestion.

playwright.config.ts
Eakam1007: Right, I missed that. Will update that with some info on this

SerpentBytes (approved):

PR #583, opened 2023-04-12
Replace doesDnsRecordExist with isDuplicateCname

### LINKED ISSUE #505: Error for duplicating DNS records doesn't have a proper error message.
## Description

When I tried to create a DNS record with same name, type, and value on staging, it would give me this error message page:
image
While it does prevent users from duplicating DNS records, I imagine the message is not very helpful for users, so some text should be there to explain what's wrong.

## Recreating the bug
On staging, create 2 DNS records with the same name, type, and value.

Genne23v:
Resolve #505 

This PR replaces `doesDnsRecordExist` with `isDuplicateCname`.

### DIFF
--- a/app/models/dns-record.server.ts
+++ b/app/models/dns-record.server.ts
@@ -70,8 +70,8 @@ export async function createDnsRecord(
     }
   }
 
-  if (await doesDnsRecordExist(data)) {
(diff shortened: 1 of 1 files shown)


humphd (approved):

dadolhay (approved):
Cool!

PR #590, opened 2023-04-12
Hopefully this is what I need to make a changelog on merging to release

sfrunza13:
I think that what this should do is create a new release tag tagged as "release" and generate a change log titled "Change Log" with the commits between this release tag and the previous one, I'm pretty sure this will overwrite whatever was under the previous release tag.

Also I think my filter was wrong, it needs to be "branches" not "branch"

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,7 +2,7 @@ name: Release
 
 on:
   push:
-    branch:
+    branches:
       - release
 
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -2,14 +2,25 @@ name: Release
 
 on:
   push:
-    branch:
+    branches:
```
humphd, line 5: Ah, I missed this.  Nice catch

humphd (approved):

PR #601, opened 2023-04-12
Setup database migration for staging and prod

cychu42:
Part of #291

## Explanation
Helpfully these explanations make it easier to understand what I'm doing and why, in this PR.

This PR should allow staging and production to use Prisma migration by default, through deploying migration files.
Migration should keep data in the database. Only when `DATABASE_SETUP =1` is an environment variable, we would go back to force pushing and cleaning data like before.
For there to be anything to deploy, I made the first migration file too.

A script is setup to make creating migration files locally with docker easier. When Prisma creates a migration file, it requires permission to create a temporary shadow database to do this, which means the user of MySQL database needs `CREATE`, `ALTER`, `DROP`, and `REFERENCES` privilege. Therefore, I use `root` user to do this in the script, against the local docker database container. The script doesn't seed database or apply migration, just creating a migration file and also generating Prisma client (default behavior) in case a dev forgets.

## Changes
* Add migration in `docker-entrypoint.sh`, when DATABASE_SETUP !=1
* Add migration file creation script, with required permission, in `package.json`
* Create an initial migration file in `prisma/migrations`
* Add documentation for making migration files in `DEPLOY.md`

## Test
1.You can test the script `npm run db:migration` to create migration files using our local docker database container.

2.You can test deployment of the migration file via `npx prisma migrate deploy`, to an empty docker database container, or one that already has this migration applied to. It should either work or tell you no migration is needed. 
Note: this command doesn't check for database drift.

Having this run on staging would be the real test.


docker-entrypoint.sh
```diff
@@ -17,6 +17,19 @@ database_setup() {
   echo "Database setup complete"
 }
 
+database_migration() {
+  echo "Running database migration..."
+  DATABASE_URL=$(</run/secrets/DATABASE_URL)
+  export DATABASE_URL
+
+  # Deploy migration files
```
humphd, line 25: Let's add a comment that says this won't delete existing data.

docker-entrypoint.sh
```diff
@@ -17,6 +17,19 @@ database_setup() {
   echo "Database setup complete"
 }
 
+database_migration() {
```
humphd, line 20: Please add a comment above this function explaining what it is, and how it works vs. `database_setup`.  It might be good to link to the Prisma docs in both cases: https://www.prisma.io/docs/concepts/components/prisma-migrate/mental-model

humphd (changes requested):
I've left a few comments, but I think this looks great.

I have one question: the Prisma docs talk about running this via CI.  We're currently doing it differently, by running it on startup for the container.  This means both instances will run it.  I think this means we should adjust our Docker deploy settings to only bring up 1 instance at a time vs. doing both at once.  We also need to shut down the existing ones first (i.e., we can't have zero-downtime deploys, which is OK).  My reading of the docs indicates that the second instance will skip doing most of the work, since the db will already be in sync (i.e., it should be safe to do this in each instance).  Do you agree?

humphd:
For the docker changes, see https://docs.docker.com/compose/compose-file/deploy/#update_config.  Specifically, I think you need to update our docker compose files:

```yml
parallelism: 1
failure_action: rollback
order: stop-first
```

This is my reading of it, but confirm what you think.  We'll have to play with this on staging to know for sure.

cychu42:
Thinking about it, I added extra warning in `DEPLOY.md` about schema change without Prisma migration.
Also expanded comment for database_setup vs database_migration as suggested

DEPLOY.md
```diff
@@ -102,6 +102,15 @@ To do this, run the `starchart` container as a service, with the additional envi
 
 Modify the `docker-compose.yml` you are using (e.g., `docker-staging.yml` or `docker-production.yml`) and add `DATABASE_SETUP=1` in the `environment` section of the `mycustomdomain` service. You can (and should!) remove this after you get the database set up, especially on production, so that re-deploying doesn't wipe the database.
 
+### New Changes to Schema
```
humphd, line 105: `New Changes to the Database Schema`

DEPLOY.md
```diff
@@ -102,6 +102,15 @@ To do this, run the `starchart` container as a service, with the additional envi
...
+
+Because staging and production use Prisma migration, changes in schema need to become migration files for Prisma migration to run the files to apply changes. A migration file is essentially a set of SQL queries Prisma generates that would get a database to a desired state, by comparing your current schema and the latest migration file.
```
humphd, line 107: `changes in the schema...`

DEPLOY.md
```diff
@@ -102,6 +102,15 @@ To do this, run the `starchart` container as a service, with the additional envi
...
+
+To make a new migration file with the new changes, run `npm run db:migration`. You will be prompted to give the file a name.
```
humphd, line 109: "After any changes are made to the `prisma/schema.prisma` file, a new migration file needs to be created (i.e., so it can be applied to the production databases).  To make a new migration file..."

DEPLOY.md
```diff
@@ -102,6 +102,15 @@ To do this, run the `starchart` container as a service, with the additional envi
...
+
+WARNING: Outside of special circumstances, please DO NOT change schema without using Prisma migration on staging and production. It will very likely cause issues with database migration. It will then require manual fixing to get the migration history sorted out.
```
humphd, line 111: "please DO NOT change the schema"

docker-entrypoint.sh
```diff
@@ -1,22 +1,46 @@
 #!/bin/bash
 set -eo pipefail
 
+# For reset or first database setup ONLY!!!
 # Run the necessary commands to sync the Prisma schema
-# with the database. We pull the database URL out of
-# secrets, since Prisma requires it as an env var.
+# with the database by wiping existing database and
```
humphd, line 6: by wiping the existing database

docker-entrypoint.sh
```diff
@@ -1,22 +1,46 @@
...
   # Clear the DATABASE_URL from the env. The app uses it via secrets
   unset DATABASE_URL
   echo "Database setup complete"
 }
 
+# For updating database schema, assuming no changes to schema 
+# is done without Prisma migration.
+# Run the necessary commands to sync the Prisma schema with
+# the database by applying migration files.
+# See:https://www.prisma.io/docs/concepts/components/prisma-migrate/mental-model
```
humphd, line 28: Space after `:`

docker-entrypoint.sh
```diff
@@ -1,22 +1,46 @@
...
-  npx prisma db push --force-reset
+  # Wipes existing database and apply migration files
+  npx prisma migrate reset --force --skip-seed
 
   # Clear the DATABASE_URL from the env. The app uses it via secrets
   unset DATABASE_URL
   echo "Database setup complete"
 }
 
+# For updating database schema, assuming no changes to schema
```
humphd, line 24: `For normal database updates, where applying migrations is enough...`

docker-entrypoint.sh
```diff
@@ -1,22 +1,46 @@
 #!/bin/bash
 set -eo pipefail
 
+# For reset or first database setup ONLY!!!
```
humphd, line 4: What does "first" mean here?  Can you clarify this a bit.  Could you say:

> For initial setup, or to completely recreate the database. ALL DATA WILL BE LOST!

humphd (changes requested):
This is looking good.  Made a few corrections/suggestions.

I'd also like to get @dadolhay's blessing on this approach.

I'm excited to try this on staging.  Before we land it, let's co-ordinate so that I can change the service on staging (it currently runs `DATABASE_SETUP=1` always).

docker-entrypoint.sh
cychu42: Yes, you understand correctly. I was talking about initial setup.

cychu42:
OK. Thank you for catching those and making suggestions. I made the changes and added a bit along the way, like stating the database in docker needs to run locally for `npm run db:migration` to work.

DEPLOY.md
```diff
@@ -102,6 +102,15 @@ To do this, run the `starchart` container as a service, with the additional envi
...
+
+After any changes are made to the `prisma/schema.prisma` file, a new migration file needs to be created (i.e., so it can be applied to the production databases). To make a new migration file, run `npm run db:migration`, while running the project's MySQL database container in Docker locally. You will be prompted to give the file a name.
+
+WARNING: Outside of special circumstances, please DO NOT change the schema without using Prisma migration on staging and production. It will very likely cause issues with database migration. It will then require manual fixing to get the migration history sorted out.
+Think of Prsima migarion as git tracking your commits, and manually fixing git history can be complicated. See [official documentation](https://www.prisma.io/docs/guides/migrate/developing-with-prisma-migrate/troubleshooting-development) for how to do it.
```
humphd, line 112: typo on Prsima.

Also, add a blank line above this paragraph.

docker-production.yml
```diff
@@ -50,9 +50,11 @@ services:
         parallelism: 0
         failure_action: rollback
       update_config:
-        # Update both containers at once, rollback on failure
-        parallelism: 2
+        # Update one containers at once, rollback on failure
```
humphd, line 53: "Stop all existing containers, then update one container at a time, waiting 20s between each. If something fails, rollback to last-known-good"

humphd: Can you do this same set of changes on `docker-staging.yml` too, so they match?

humphd (changes requested):

DEPLOY.md
cychu42: I'll format this a bit and add bolding to WARNING and "DO NOT"

docker-production.yml
cychu42: Of course. I should have thought of that.

humphd (dismissed):
Awesome.

Just remember, ping me before you are going to merge this.

cychu42:
We would want to use `DATABASE_SETUP=1` for the first time running this PR on staging and production, since no migration has been done on them yet and Prisma would just be confused why all those tables already exist.

If nothing goes wrong from there, we can switch it off for future updates.

dadolhay:
@cychu42 @humphd 
I'd suggest using something more dramatic, like: `DANGER_DATABASE_WIPE_REINITIALIZE`

cychu42:
I changed all mention of `DATABASE_SETUP` to your spooky version, including the same name function and variable.

docker-entrypoint.sh
```diff
@@ -29,12 +54,16 @@ clear_redis() {
...
 }
 
-# See if we need to do database setup before starting.
-if [[ $DATABASE_SETUP == "1" ]]; then
-  # Run our database setup
-  database_setup
+# See if we need to do database wipe and reinitialize before starting.
+if [[ $DANGER_DATABASE_WIPE_REINITIALIZE == "1" ]]; then
+  # Run database wipe and reinitialize
+  DANGER_DATABASE_WIPE_REINITIALIZE
```
dadolhay, line 60: If this is the fn call on line 10, should this not be lower case?

cychu42: You are right.

humphd (approved):

humphd:
@dadolhay can you review this again please?

dadolhay (approved):

PR #607, opened 2023-04-12
Add user deactivation functionality

### LINKED ISSUE #308: Allow users to be deactivated

Myrfion:
- add `intent` field for each action on admin screen, to manage them properly 
- add function for user deletion, which deletes user from db and triggers reconcile 
- integrate functionality with the front end 

https://user-images.githubusercontent.com/32075241/231605805-bd3030b0-50f1-4f87-a5e7-a69563d3db6d.mov

Closes #308

### DIFF
--- a/app/components/admin/users-table.tsx
+++ b/app/components/admin/users-table.tsx
@@ -34,7 +34,7 @@ export default function UsersTable({ users, searchText }: UsersTableProps) {
   const isInputValid = searchText.length >= MIN_USERS_SEARCH_TEXT;
   const isLoading = navigation.state === 'submitting';
 
-  const shouldShowInstruction = users.length === 0 && !isInputValid && !isLoading;
+  const shouldShowInstruction = !isInputValid && !isLoading;
   const shouldShowNoUsersMessage = users.length === 0 && isInputValid && !isLoading;
   const shouldShowUsers = !(isLoading || shouldShowInstruction || shouldShowNoUsersMessage);
 
@@ -94,7 +94,7 @@ export default function UsersTable({ users, searchText }: UsersTableProps) {
                             />
                           </Form>
                         </Tooltip>
-                        <Form method="post" reloadDocument>
+                        <Form method="post">
                           <Tooltip label="Deactivate user">
                             <IconButton
                               aria-label="Deactivate user"
--- a/app/components/admin/users-table.tsx
+++ b/app/components/admin/users-table.tsx
@@ -95,16 +95,16 @@ export default function UsersTable({ users, searchText }: UsersTableProps) {
                           </Form>
                         </Tooltip>
                         <Form method="post">
-                          <Tooltip label="Deactivate user">
+                          <Tooltip label="Delete user">
                             <IconButton
-                              aria-label="Deactivate user"
+                              aria-label="Delete user"
                               icon={<DeleteIcon color="black" boxSize={5} />}
                               variant="ghost"
                               type="submit"
                             />
                           </Tooltip>
                           <input type="hidden" name="username" value={user.username} />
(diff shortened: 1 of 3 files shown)


sfrunza13 (approved):

sfrunza13:
When I originally wrote the admin thing I tried expanding the ZOD schema and it failed, I made effective username optional so I was stumped at the time as to why it would not work, I am now realizing I forgot to make the search string optional! Since they are mutually exclusive basically it never worked for me lol.

app/components/admin/users-table.tsx
```diff
@@ -93,13 +94,18 @@ export default function UsersTable({ users, searchText }: UsersTableProps) {
...
                         </Tooltip>
-                        <Tooltip label="Deactivate user">
-                          <IconButton
-                            aria-label="Deactivate user"
-                            icon={<DeleteIcon color="black" boxSize={5} />}
-                            variant="ghost"
-                          />
-                        </Tooltip>
+                        <Form method="post" reloadDocument>
+                          <Tooltip label="Deactivate user">
```
humphd, line 98: I'd use `Delete` vs. `Deactivate` everywhere, since we're really dropping the user from the db.

app/lib/user.server.ts
```diff
@@ -0,0 +1,8 @@
+import type { User } from '@prisma/client';
+import { setIsReconciliationNeeded } from '~/models/system-state.server';
+import { deleteUserByUsername } from '~/models/user.server';
+
+export async function deleteUser(username: User['username']) {
+  await deleteUserByUsername(username);
+  setIsReconciliationNeeded(true);
```
humphd, line 7: `return` here, so the `Promise` is returned"

```ts
return setIsReconciliationNeeded(true);
```

humphd (changes requested):
This is great.  A few small things.

Genne23v:
How about add a pop-up to reduce chances of mistake by admin?

humphd:
> How about add a pop-up to reduce chances of mistake by admin?

You mean like we do for deleting DNS records? That's a great idea.  It could be filed/done as a follow-up.

app/lib/user.server.ts
cychu42: Instead of doing this, what if we add `setIsReconciliationNeeded(true);` inside `deleteUserByUsername` and just use it?

app/lib/user.server.ts
```diff
@@ -4,5 +4,5 @@ import { deleteUserByUsername } from '~/models/user.server';
 
 export async function deleteUser(username: User['username']) {
   await deleteUserByUsername(username);
-  setIsReconciliationNeeded(true);
+  return setIsReconciliationNeeded(true);
```
humphd, line 7: We could do a follow-up to audit all callers of `setIsReconciliationNeeded(true)` to see if we can move them into the model methods per @cychu42's comment, if you don't want to do it here.

humphd (approved):

dadolhay (approved):

PR #608, opened 2023-04-13
Improve instructions link on DNS record page,  form and DNS record Id

### LINKED ISSUE #597: Improve instructions link on DNS Record page and form
We have default styling for most links to make them red and underlined.  The "instructions page" link on `/dns-records` and in `/dns-records/new` are black:

Screenshot 2023-04-12 at 3 41 56 PM

Screenshot 2023-04-12 at 3 44 58 PM

Also, let's change the location of this final sentence.  On the first-run page, it could be placed below the centred "Create your first DNS Record!" button.  After the first-run, we could show it below the table?

We should also add

manekenpix:
Fixes #597

This PR changes the links in [dns-records/index](https://github.com/DevelopingSpace/starchart/blob/main/app/routes/__index/dns-records/index.tsx), [dns-records/new](https://github.com/DevelopingSpace/starchart/blob/main/app/routes/__index/dns-records/new.tsx), and [dns-records/$dnsRecordId](https://github.com/DevelopingSpace/starchart/blob/main/app/routes/__index/dns-records/%24dnsRecordId.tsx) (this last one is not included in the issue, but I thought it also needed fix) to be red and underlined.

image

image

**Note**: I noticed that both _instruction_ and _instructions_ are used in links to the same page (the instructions page). 
image

Should we make sure that we use the same word in all links? If so, I can make the change in this PR.

### DIFF
--- a/app/components/landing-page/landing-page-card.tsx
+++ b/app/components/landing-page/landing-page-card.tsx
@@ -38,11 +38,11 @@ export default function LandingPageCard({
           <Text>{cardDescription}</Text>
           {instructionsPath && (
             <Text>
-              To learn more, refer to{' '}
+              To learn more, refer to our{' '}
               <Link as={RemixLink} to={instructionsPath}>
                 {instructionsPath === '/certificate/information'
-                  ? 'our information page'
-                  : 'our instruction page'}
+                  ? 'information page'
+                  : 'instructions page'}
               </Link>
               .
             </Text>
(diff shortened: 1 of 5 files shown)


Eakam1007:
> Should we make sure that we use the same word in all links? If so, I can make the change in this PR.

Thanks for pointing this out! Yes, I think it would be better for both of them to be `instructions`. I wonder if we should also move the `our` out of the link similar to the other two pages

humphd (approved):
Great fix, and wonderful to see you contributing!

Yes, I'd harmonize the wording for all three.

humphd (approved):

humphd (approved):

sfrunza13:
Good stuff, should the suggestion on the tail end of the issue #597 also be addressed here?

> Also, let's change the location of this final sentence. On the first-run page, it could be placed below the centred "Create your first DNS Record!" button

humphd:
It can happen in a follow-up or here, either works for me.

PR #613, opened 2023-04-13
Update certificate UI test

### LINKED ISSUE #478: Test: Updating Certificate
@Ririio: Do you want to file a follow-up and fix this there?  I'd like to avoid blocking this further

_Originally posted by @humphd in https://github.com/DevelopingSpace/starchart/pull/434#discussion_r1152309964_
            
            
The current iteration of certificate test does not work with the Back-End, due to the process of having the certificate taking a long while. This issue should prioritize the delay that may occur from the certificate request

Genne23v:
Resolve #478 

- Increase timeout from 30 to 150 to wait for certificate creation
- Update UI test elements as per recent UI uupdate
- Delete certificate DB before each browser test


humphd:
I think we need a different approach here.  We need to run the cert creation job faster in testing somehow, and not wait so long.

Genne23v:
> I think we need a different approach here. We need to run the cert creation job faster in testing somehow, and not wait so long.

I think I should mock Let's Encrypt.

humphd:
Sure, that would work too.

test/e2e/certificate-available.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+import { prisma } from '../../app/db.server';
+import { loggedInAsUser } from './utils';
+
+import type { Certificate } from '@prisma/client';
+
+test.describe('Certificate Page', () => {
+  loggedInAsUser();
+
+  test.beforeAll(async () => {
+    const data: Pick<
```
humphd, line 12: Avoid this pattern in TypeScript.  Because you are hoisting this variable outside of the context of the function call below, it becomes untethered from the type information.  If, instead, you define this data as a literal inside the function call, all type information can be inferred:

```ts
// 1. We can't rely on type inference in fn() by doing this, and must manually type it
const foo: ComplexTypeDeclarationHere = { avoid: 'this' };
fn(foo);

// 2. Let fn() infer the type of our data by defining it within the context in which it will be used:
fn({ do: 'this' });
```

test/e2e/certificate-available.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  test.beforeEach(async ({ page }) => {
+    await page.goto('/certificate');
+  });
+
+  test.afterAll(async () => {
+    await prisma.certificate.deleteMany({});
+  });
+
+  test('Request a Certificate', async ({ page }) => {
```
humphd, line 48: This test description doesn't describe what you're doing below.

Also, you are doing too many things in this one test.  Separate copy from download

test/e2e/landing-page.spec.ts
```diff
@@ -1,10 +1,12 @@
 import { test, expect } from '@playwright/test';
+import { prisma } from '../../app/db.server';
 import { loggedInAsUser } from './utils';
 
 test.describe('Landing Page', () => {
   loggedInAsUser();
 
   test.beforeEach(async ({ page }) => {
+    await prisma.certificate.deleteMany({});
     await page.goto('/');
```
humphd, line 10: `return page.goto('/');`

test/e2e/landing-page.spec.ts
```diff
@@ -1,10 +1,12 @@
 import { test, expect } from '@playwright/test';
+import { prisma } from '../../app/db.server';
 import { loggedInAsUser } from './utils';
 
 test.describe('Landing Page', () => {
   loggedInAsUser();
 
   test.beforeEach(async ({ page }) => {
+    await prisma.certificate.deleteMany({});
```
humphd, line 9: Why does `deleteMany` require you to pass an empty object literal?  Can't we do that in the definition of the function?

humphd (changes requested):

humphd:
@Genne23v you can run the tests locally so you don't have to keep pushing here and waiting.  `npm run e2e`

humphd:
@dadolhay suggests that we should consider using an env variable for testing, and alter the retry timeout to reduce it such that the test will run faster in testing environments.  We might also need to decrease the polling time in the front-end too, so it doesn't go too long.

@Eakam1007 suggests using `NODE_ENV=test`

Genne23v:
> @Genne23v you can run the tests locally so you don't have to keep pushing here and waiting. `npm run e2e`

Sorry for inconvenience. but local test produces a better result. This will end soon.

app/models/certificate.server.ts
```diff
@@ -71,3 +71,7 @@ export function getExpiredCertificates() {
     },
   });
 }
+
+export function deleteAllCertificatesInDb() {
```
humphd, line 75: `InDb` is not needed (all methods in the model operate on the db).

humphd (changes requested):

humphd (approved):

test/e2e/landing-page.spec.ts
```diff
@@ -1,11 +1,13 @@
...
+import { deleteAllCertificates } from '../../app/models/certificate.server';
 import { loggedInAsUser } from './utils';
 
 test.describe('Landing Page', () => {
   loggedInAsUser();
 
   test.beforeEach(async ({ page }) => {
-    await page.goto('/');
+    await deleteAllCertificates();
+    return await page.goto('/');
```
Eakam1007, line 10: Sorry, why is this here? Also, I don't think you need the return here

Eakam1007 (changes requested):

test/e2e/landing-page.spec.ts
humphd: You don't need the `await`

test/e2e/landing-page.spec.ts
```diff
@@ -1,11 +1,13 @@
...
+import { deleteAllCertificates } from '../../app/models/certificate.server';
 import { loggedInAsUser } from './utils';
 
 test.describe('Landing Page', () => {
   loggedInAsUser();
 
   test.beforeEach(async ({ page }) => {
-    await page.goto('/');
+    await deleteAllCertificates();
+    return page.goto('/');
```
Eakam1007, line 10: Same question as before, why do we need to call `deleteAllCertificates` here?

Eakam1007 (changes requested):

test/e2e/landing-page.spec.ts
Genne23v: I added a fake certificate in `certificate-available.spec`. I noticed that this landing page test state is somehow affected from the certificate from other test. So one of the button text is different which leads this landing page test to a failure.

Eakam1007: I see that you are deleting the certificates in that spec after the tests are run. Could you confirm that this is needed?

test/e2e/landing-page.spec.ts
```diff
@@ -5,7 +5,7 @@ test.describe('Landing Page', () => {
   loggedInAsUser();
 
   test.beforeEach(async ({ page }) => {
-    await page.goto('/');
+    return page.goto('/');
```
Eakam1007, line 8: I am not sure if we need this change now

Eakam1007 (changes requested):
One small thing, rest looks good

Eakam1007 (approved):

humphd (approved):

PR #614, opened 2023-04-13
Fix remix depreciation warnings

### LINKED ISSUE #320: Fix Remix deprecation warnings
Remix is deprecating the old nested folders route convention in favor of "flat routes" in v2. There will be a helper function to use the old convention in v2 but they recommend making the change to the newer convention now and making the jump to v2 easier in the future.

https://remix.run/docs/en/main/file-conventions/route-files-v2

I believe this is the 'flat-routes' in question https://github.com/kiliman/remix-flat-routes

`.` in file name becomes a `/` in the route example: `concerts.trending.tsx` = `/concerts/trending`
`$` is used for dynamic data like ids example: `users.$id.tsx`  = `/users/12`

you can also create a folder and treat it as a flat file by appending `+` to it
`_auth+/forgot-password.tsx => _auth.forgot-password.tsx`

It is apparently possible to migrate our existing routes: https://github.com/kiliman/remix-flat-routes#-migrating-existing-routes

Eakam1007:
Resolves #320

Fixed our remix depreciation warnings following their [guide](https://remix.run/docs/en/main/pages/v2)

### Changes made
- Updated remix config to use [v2 route layout](https://remix.run/docs/en/main/pages/v2#file-system-route-convention)
- Rearranged the entire routes folder to use the new v2 route convention. This moves routes out of folders into just files which are easier to see
- Renamed the layout file `_index.tsx` to `_auth.tsx`. This was needed because now the index routes are specified by `_index.tsx`, and it was not possible to specify the index route's layout as `_index.tsx`
- Updated remix config to use [v2 formMethod](https://remix.run/docs/en/main/pages/v2#formmethod). I found no instance of formMethod anywhere so nothing else was changed for this
  (Summary of v1 vs v2: `navigation.formMethod === "post";` should now be `navigation.formMethod === "POST";` etc)
- Updated remix config to use [v2 ErrorBoundry API](https://remix.run/docs/en/main/pages/v2#catchboundary-and-errorboundary)
- Merged all existing `CatchBoundry` and `ErrorBoundy` into one `ErrorBoundry` for each file with checks to render different error layouts. `CatchBoundry` no longer exists in v2
- Updated remix config to use [v2 meta](https://remix.run/docs/en/main/pages/v2#route-meta)
- Updated root meta function to return array of objects for meta tags instead of a single object with key value pairs


humphd:
Discussion on Slack: based on https://remix.run/docs/en/main/file-conventions/route-files-v2#nested-layouts-without-nested-urls, we think we should use `_auth.*` as our pathless, layout route prefix.

humphd (approved):

sfrunza13 (dismissed):
Semantically I think that using _auth makes sense because we have to be authenticated for everything to do with starchart. I like it.

Eakam1007:
Updated to fix the rest of the depreciation warnings + rebased with main

humphd (approved):
Thanks for doing the error/catch boundary change too.  This looks great.

sfrunza13 (approved):

PR #635, opened 2023-04-17
Log when impersonation starts/stops

### LINKED ISSUE #629: Impersonating someone should produce an info log
When we begin to impersonate a user, we should log that this has happened with an `.info()` log for better tracking of what happened.  We should include: 1) who is impersonating; 2) who are they impersonating.  We should also log when the impersonation ends

### LINKED ISSUE #632: Fix eslint warning
/home/runner/work/starchart/starchart/app/queues/certificate/dns-waiter-worker.server.ts
Warning:   1:[10](https://github.com/DevelopingSpace/starchart/actions/runs/4723096047/jobs/8378574828#step:6:11)  warning  'UnrecoverableError' is defined but never used  @typescript-eslint/no-unused-vars

humphd:
Adds logging for when an admin user starts and stops impersonation.

I've also fixed the ESLint warning we introduced earlier today.

Fixes #629 
Fixes #632


app/root.tsx
```diff
@@ -49,6 +50,7 @@ export const action = async ({ request }: ActionArgs) => {
   const formData = await request.formData();
   const originalName = formData.get('originalName');
   if (typeof originalName === 'string') {
+    logger.info(`Admin (${originalName}) stopped impersonation`);
```
dadolhay, line 53: Should this not be done in setEffectiveUsername (add an else part to that condition)?

dadolhay (changes requested):

app/root.tsx
humphd: I tried that, but we seem to go through that code path on normal startup (e.g., login), and there's no way to differentiate a login from stopping impersonation.  It meant that it always logged that the admin user was stopping impersonation on regular login.  I used this method as a way to indicate it's really done.  I could add a function to do it and log, but it would be the same code path as this.

sfrunza13 (dismissed):

app/root.tsx
dadolhay: My issue is code maintainability. I'm afraid that this current solution will break in the next refactor / if that fn gets called from an other place


What would you think about changing the signature of this:
```TS
export async function setEffectiveUsername(
  username: string,
  effectiveUsername: string | null
): Promise<string> {
```

to somehing like this overloaded signature (have not tested this lang element yet, but according to the docs it should work :) )
```ts
function setImpersonation(action: 'init', username: string): Promise<string>;
function setImpersonation(action: 'start', username: string, effectiveUsername: string): Promise<string>;
function setImpersonation(action: 'end', username: string): Promise<string> {
   ...
}
```

dadolhay: that way you could hook onto the action parameter, and push the proper log

cychu42: Maybe we should add `effectiveUsername` in the log message here, to make it easier to track in the log.
For example, perhaps an admin decides to do multiple impersonations, and we need to go through a lot of log messages to figure out which is which.

cychu42 (changes requested):

humphd:
OK, I've redone this.  Doing the logging so it works in all cases was more involved than I expected, but I think this does it.

What I do now is to log what happens at `setEffectiveUsername()`.  I've added two new higher-level API calls for the front-end to use when we want to `startImpersonation()` or `stopImpersonation()`.  These call through to lower-level session code, eventually setting the effective username via `setEffectiveUsername()` and logging.

While I was testing this, I notice that we allowed impersonating and deleting ourself.  This complicates things for no appreciable gain.  In the past we had an idea to allow impersonating yourself to drop your rights to that or a regular user (e.g., to be able to create DNS Records or a Certificate); however, admins can also do that in our system, so impersonating yourself does nothing.

To test this, try variations of the following, and watch the logs:

1. login as `user3:user3pass` so you become an admin
2. logout partially and fully (SLO)
3. impersonate `user1`, then try "Stop Impersonation" and also doing a logout without stopping impersonation

We log when the admin becomes another user, or stops being that user, but not for regular login/logout.

humphd:
@cychu42 I don't have access to the `Request` at this stage, so can't get it easily without further complicating the API.

sfrunza13 (approved):
Tested it and it works on my machine

cychu42 (approved):

humphd:
@dadolhay do you want to look at this before I merge?

PR #638, opened 2023-04-18
fix: certification expiration resulting in multiple emails

### LINKED ISSUE #637: Certificate expiration warning results in 4 emails
Create a certificate on staging.  Wait for a few hours, and you'll get 4 notification emails saying that the certificate is expiring (staging certs only last 1 week).

We should update the notification logic so that it doesn't spam users.  Are we not updating the database to indicate when we last told them about expiry?

cc @SerpentBytes

SerpentBytes:
## Description
Closes #637 

### Steps to test
- Run the project
- Notice the log messages
- Create a new certificate (if needed)
- Navigate to http://localhost:8025 after 5 mins

### DIFF
--- a/services/expiration/expiration-notification.server.ts
+++ b/services/expiration/expiration-notification.server.ts
@@ -192,9 +192,8 @@ const getExpiringDnsRecords = () => {
 async function addExpirationNotifications(type: string) {
   // we never want to add multiple jobs for the same user/email notification (unique jobId)
   let jobId = `${expirationNotificationQueueName}-${type}-expiry`;
-  return expirationNotificationQueue.add(expirationNotificationQueueName, {
+  return expirationNotificationQueue.add(expirationNotificationQueueName, jobId, {
     repeat: { every: 5 * 60 * 1000 },
-    jobId,
   });
 }
 
--- a/services/expiration/expiration-notification.server.ts
+++ b/services/expiration/expiration-notification.server.ts
@@ -191,9 +191,10 @@ const getExpiringDnsRecords = () => {
 // function to add jobs
 async function addExpirationNotifications(type: string) {
   // we never want to add multiple jobs for the same user/email notification (unique jobId)
-  let jobName = `${expirationNotificationQueueName}-${type}-expiry`;
-  return expirationNotificationQueue.add(jobName, {
+  let jobId = `${expirationNotificationQueueName}-${type}-expiry`;
+  return expirationNotificationQueue.add(expirationNotificationQueueName, {
     repeat: { every: 5 * 60 * 1000 },
+    jobId,
   });
 }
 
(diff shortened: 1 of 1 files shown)


services/expiration/expiration-notification.server.ts
```diff
@@ -186,8 +189,9 @@ const getExpiringDnsRecords = () => {
   });
 };
 // function to add jobs
-async function addExpirationNotifications() {
-  let jobName = `${expirationNotificationQueueName}`;
+async function addExpirationNotifications(type: string) {
+  // we never want to add multiple jobs for the same user/email notification (unique jobId)
```
humphd, line 193: NOTE: the job name and id are not the same thing.  You need to put a unique `jobId` on the notification jobs themselves.  I gave you some code already in Slack:

```ts
export const addNotification = (data: NotificationData) => {
  const jobName = `notification`;
  data.jobId = `${data.emailAddress}:${data.subject}`;
  return notificationsQueue.add(jobName, data);
};
```

The name of the job doesn't affect duplication of jobs in the queue; only the `jobId` can fix that.

SerpentBytes: Sorry, I am probably having a bad day. Thanks for reminding me!

services/expiration/expiration-notification.server.ts
```diff
@@ -186,10 +189,12 @@ const getExpiringDnsRecords = () => {
...
 // function to add jobs
-async function addExpirationNotifications() {
-  let jobName = `${expirationNotificationQueueName}`;
-  return expirationNotificationQueue.add(jobName, {
+async function addExpirationNotifications(type: string) {
+  // we never want to add multiple jobs for the same user/email notification (unique jobId)
+  let jobId = `${expirationNotificationQueueName}-${type}-expiry`;
+  return expirationNotificationQueue.add(expirationNotificationQueueName, {
     repeat: { every: 5 * 60 * 1000 },
+    jobId,
```
humphd, line 197: I don't think this is right.  The [signature for .add()](https://api.docs.bullmq.io/classes/Queue.html#add) takes a `name`, `data`, and `options`.  Here you are putting the `jobId` into the `options`, but I think it needs to go in `data`:

```ts
return expirationNotificationQueue.add(expirationNotificationQueueName, { jobId }, {
  repeat: { every: 5 * 60 * 1000 },
});
```

Test this to make sure I'm right, but this is what I see in the docs.

humphd (changes requested):

humphd (changes requested):
This isn't doing what I'm asking.

I want you to do this on the notification queue, not the expiration queue.  The id name needs to be specific to the user + subject, so that multiple requests to create an email notification get combined.  It doesn't matter how many times we say to add a notification job, as long as they use an ID that is unique per user+subject.

If you're not sure what I'm saying, I can do it.  Let me know.

PR #639, opened 2023-04-18
Rework default tags for staging and production builds

humphd:
I was updating production today, to use the latest database migrations, and I ran into an interesting bug.  You can't use an image built for staging to run on production, or vice-versa, due to how we bundle the SAML IdP metadata into the build.  An image for staging needs to be run on staging, and an image for production needs to be run on production.

This change updates our CI, docker tags, and docker-compose stack definitions to deal with this.  What we'll do now:

- in CI, build `staging*` images, and we'll use `staging` as our movable `latest` tag for staging
- in Release, build `production*` images, and we'll use `production` as our moveable `latest tag for production

We won't use `latest`, since it's not meaningful anymore, nor is `main`.

We need this in place before 1.0, and I need to update the stacks on staging and production manually to use them.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -349,13 +349,14 @@ jobs:
           build-args: |
             SAML_IDP_METADATA_PATH=./config/idp-metadata-staging.xml
           push: true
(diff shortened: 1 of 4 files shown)


Eakam1007 (approved):

Genne23v (approved):
It all looks good to me

sfrunza13 (approved):

sfrunza13:
I understand the problem and the solution, which is always exciting.

cychu42 (approved):
LGTM.

PR #642, opened 2023-04-18
Reduce number of browsers we run in e2e tests

humphd:
Yesterday we were discussing the value of reducing the length of time our e2e tests take.  Doing so requires us to use fewer browsers in CI.

Here I'm using:

- Firefox (desktop)
- Chrome (desktop + mobile)
- Webkit (mobile)
- Edge (desktop)

The inclusion of Edge adds some degree of additional UI checks for a Chromium-based browser that has differences to Chrome.

Removing desktop Webkit (no one uses it, sorry Apple, much as I love you!), and Chromium vs. Chrome is not really helping us get more coverage.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -58,28 +58,11 @@ const config: PlaywrightTestConfig = {
       name: 'setup',
       testMatch: /.*\.setup\.ts/,
     },
-    {
-      name: 'chromium',
-      use: {
-        ...devices['Desktop Chrome'],
-      },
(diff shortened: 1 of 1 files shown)


SerpentBytes (approved):

humphd:
12:26 down to 9:40, 3 mins saved.

Genne23v (approved):

Eakam1007 (approved):

PR #652, opened 2023-04-19
fix: Account for ACME order deduplication

### LINKED ISSUE #651: Unable to create certificate order on production
I've fixed the issue with AWS on production, and now it looks like we have a new problem with Let's Encrypt:

```
{"level":"info","message":"Creating certificate order for davidhumphrey.mystudentproject.ca","timestamp":"2023-04-19T19:26:02.292Z"}
{"level":"debug","message":"Initializing Let's encrypt","timestamp":"2023-04-19T19:26:02.292Z"}
{"level":"debug","message":"Creating ACME order for davidhumphrey.mystudentproject.ca","timestamp":"2023-04-19T19:26:03.119Z"}
{"level":"info","message":"Order created successfully","timestamp":"2023-04-19T19:26:03.735Z"}
{"clientVersion":"4.13.0","code":"P2002","level":"error","message":"Failed to update certificate 9 in db \nInvalid `prisma.certificate.update()` invocation:\n\n\nUnique constraint failed on the constraint: `Certificate_orderUrl_key`","meta":{"target":"Certificate_orderUrl_key"},"stack":"Error: \nInvalid `prisma.certificate.update()` invocation:\n\n\nUnique constraint failed on the constraint: `Certificate_orderUrl_key`\n    at pn.handleRequestError (/app/node_modules/@prisma/client/runtime/library.js:176:6477)\n    at pn.handleAndLogRequestError (/app/node_modules/@prisma/client/runtime/library.js:176:5907)\n    at pn.request (/app/node_modules/@prisma/client/runtime/library.js:176:5786)\n    at t._request (/app/node_modules/@prisma/client/runtime/library.js:179:10484)\n    at Worker.import_bullmq.Worker.connection [as processFn] (/app/build/index.js:1965:7)\n    at Worker.processJob (/app/node_modules/bullmq/src/classes/worker.ts:667:22)\n    at Worker.retryIfFailed (/app/node_modules/bullmq/src/classes/worker.ts:858:16)","timestamp":"2023-04-19T19:26:03.747Z"}
```

I think we're failing here:

https://github.com/DevelopingSpace/starchart/blob/main/app/queues/certificate/order-creator-worker.server.ts#L118-L127

In the database, the `orderUrl` is `NULL`:

Screenshot 2023-04-19 at 3 34 05 PM

I've tried starting and stopping the containers a few times, and retried after the process fails.  The same error happens every time.

cc @dadolhay.

dadolhay:
Closes #651

### DIFF
--- a/app/queues/certificate/certificate-error-handler.server.ts
+++ b/app/queues/certificate/certificate-error-handler.server.ts
@@ -57,6 +57,12 @@ export const initCertificateErrorHandler = () => {
 
       await certificateModel.updateCertificateById(certificateId, {
         status: CertificateStatus.failed,
+        /**
+         * Setting orderUrl to null, so if we get back the same cert request from the
+         * ACME provider next time (as it deduplicates orders) we are not violating the
+         * unique constraint of this field
+         */
+        orderUrl: null,
       });
 
       await challengeModel.deleteChallengesByCertificateId(certificateId);


app/queues/certificate/certificate-error-handler.server.ts
```diff
@@ -57,6 +57,12 @@ export const initCertificateErrorHandler = () => {
 
       await certificateModel.updateCertificateById(certificateId, {
         status: CertificateStatus.failed,
+        /**
+         * Setting orderUrl to null, so if we get back the same cert request from the
+         * ACME provider next time (as it deduplicates orders) we are not violating the
+         * unique constraint of this field
```
humphd, line 63: What is the value of the unique constraint, since we update the certificate by it's `id`?

humphd (approved):

SerpentBytes (approved):

app/queues/certificate/certificate-error-handler.server.ts
humphd: Yes, I understand why this is happening.  What I want to know is, why are we saying this URL has to be unique?  We've just seen that in normal circumstances, it can be duplicated with a repeat attempt on a failed cert order.  What was the reason that we made this field `@unique`, and should we remove that?

dadolhay: The unique there ensures that a single user cannot run two cert requests in parallel (they'd get back the same order url twice, whouch would force one of the requests to fail just like it did now). Plus I didn't want to do a migration to remove it :)

PR #657, opened 2023-04-20
tag name => Release -> Latest

### LINKED ISSUE #654: Improve Changelog when we create a release
We have some code in the release.yml workflow to create an automated release with changelog.  However, the changelog is currently empty, and it would be nice to have it include the list of changes.

sfrunza13:
Fixes #654
Changed the name from release to latest to avoid confusion with the branch we push to right before using this action.

Tested using a repo I made that exhibited the same behavior with a release branch and tag named release that changed once the tag name was changed to "latest": 
https://github.com/sfrunza13/testingReleases
image


humphd (approved):
Thanks for digging into this

SerpentBytes (approved):

PR #660, opened 2023-04-21
Use Chakra theme to style root elements

humphd:
I was working on a PR in a different project, and noticed that you can style things globally via the Chakra theme.  This change converts the way we style `html` and `body`, using Chakra vs. `styles.css`.

See https://chakra-ui.com/docs/styled-system/global-styles

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -24,8 +24,6 @@ export const meta: V2_MetaFunction = () => [
 ];
 
 export const links: LinksFunction = () => [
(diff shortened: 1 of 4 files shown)


Eakam1007 (approved):
Looks good. I wonder if, in another issue, we should move the component styles into it's own file too like you have time for styles.ts

Ririio (approved):
Oh this is interesting, I'll take a look at what else we can do with it

PR #662, opened 2023-04-22
Fix padding issue on 'new' dns record route

SerpentBytes:
## Description
This PR adds padding between the "Create" button on the New DNS Record route and the Footer. 

### Before
before screenshot 1
before screenshot 2

### After
after screenshot 1
after screenshot 2

### DIFF
--- a/app/components/certificate/certificate-display.tsx
+++ b/app/components/certificate/certificate-display.tsx
@@ -50,7 +50,7 @@ export default function CertificateDisplay({
       gap="4"
       textAlign={{ base: 'center', md: 'left' }}
       fontSize={{ base: 'sm', md: 'md' }}
-      paddingY="6"
+      paddingBottom={6}
     >
       <HStack gap="4" alignSelf={{ base: 'center', md: 'flex-start' }}>
         <Heading as="h4" size="sm">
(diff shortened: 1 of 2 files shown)


app/components/certificate/certificate-display.tsx
```diff
@@ -97,7 +97,7 @@ export default function CertificateDisplay({
         </Tooltip>
       </HStack>
       <Text>{description}</Text>
-      <Accordion allowMultiple>
+      <Accordion allowMultiple paddingY="6">
```
humphd, line 100: Does it make more sense to put this on the root `<Flex>` component, since you're trying to push down all content in here?

humphd (dismissed):

app/routes/_auth.dns-records.new.tsx
```diff
@@ -49,7 +49,7 @@ export default function NewDnsRecordRoute() {
   const actionData = useActionData();
 
   return (
-    <Container maxW="container.xl">
+    <Container maxW="container.xl" paddingY="6">
```
Eakam1007, line 52: Does this also add a madding to the top of the container? If so, is that intended?

humphd: Good catch.  Probably want `pb={6}`

Eakam1007, line 52: I think you would want to change this to paddingBottom too

Eakam1007 (approved):

app/routes/_auth.dns-records.new.tsx
SerpentBytes: Thanks for catching this.

humphd (approved):

PR #664, opened 2023-04-23
new release idea

### LINKED ISSUE #654: Improve Changelog when we create a release

sfrunza13:
fixes #654 
image

The above is the result of the new action. It works based off merges to main since the last tag.

There can be custom categories corresponding to PR labels, the ones I have right now are the defaults but they can be changed in the configurationJson. 

Anything that does not fall under a predefined category will be caught by {{UNCATOGORIZED}} in the template and stuck in the details there that we can pop open.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -16,12 +16,47 @@ jobs:
     permissions:
       contents: write
     steps:
-      - uses: 'marvinpinto/action-automatic-releases@latest'
+      - name: 'Build Changelog'
+        id: build_changelog
+        uses: mikepenz/release-changelog-builder-action@v3
         with:
-          repo_token: '${{ secrets.GITHUB_TOKEN }}'
-          automatic_release_tag: 'latest'
-          title: 'Change Log'
-          prerelease: false
+          configurationJson: |
+            {
+              "template": "#{{CHANGELOG}}\n\n<details>\n<summary>Uncategorized</summary>\n\n#{{UNCATEGORIZED}}\n</details>",
+              "categories": [
+                {
+                  "title": "## 🚀 Features",
+                  "labels": ["feature"]
+                },
+                {
+                  "title": "## 🐛 Fixes",
+                  "labels": ["fix"]
+                },
+                {
+                  "title": "## 🧪 Tests",
+                  "labels": ["test"]
+                },
+                {
+                    "title": "## 💬 Other",
+                    "labels": ["other"]
+                },
+                {
+                    "title": "## 📦 Dependencies",
+                    "labels": ["dependencies"]
+                }
+              ]
+            }
+          commitMode: false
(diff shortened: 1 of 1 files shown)


Eakam1007:
Is it always going to be named `Change Log` ?

sfrunza13:
That's what I've got now yeah. I am using last tag to head so I could do something like this 
image

Eakam1007:
Ah, in that case maybe we can manually edit the title? I simply feel that Change Log doesn't look as good as Release <latest tag>

sfrunza13:
I can change the words to be anything I thought the concern was with the v number not being there 😂

Eakam1007:
Oh sorry, I used `<...>` format to specify what I meant but github didn't show it. I meant that something like `Release v<latest_tag>` would be nice to have. Is that possible to do with what you have already?

sfrunza13:
I think it is, this action takes a from and a to tag but as it is right now I left the defaults which takes the latest tag to head. It might be possible to tag the release and then use the latest tag as the to tag and the one before that as the from tag. Idk exactly how to write that but I think it's possible.  Then I could use the to tag in the title.

sfrunza13:
image
Got it, one sec

Eakam1007 (dismissed):
That looks good to me

humphd:
I'm not understanding how this will work, since we aren't doing tags.  Can you explain?

sfrunza13:
I can't remember where our limitations were with the tags but I think that it was with using them as the thing to start the action since we couldn't push them, there's the potential to still use version tags for certain commits and have this run upon pushing to release to find the changes between the last few version tags or alternatively we can compile a general changelog from an older tag to the current head and just keep adding to the same changelog.

sfrunza13:
The actions I am using do not automatically advance any tag so I added that, what I am going to push will advance a tag called 'latest' with the release of each change log. I'm not certain why but doing this releases are created as drafts and then have to be edited manually to be published.

.github/workflows/release.yaml
```diff
@@ -16,12 +16,71 @@ jobs:
...
+                    "labels": ["dependencies"]
+                }
+              ]
+            }
+          commitMode: false
+          fromTag: latest
+          toTag: HEAD
+          token: ${{ github.token }}
+      - name: Create Release
+        uses: softprops/action-gh-release@v1
```
Eakam1007, line 57: Could it be that this creates draft releases by default?

humphd:
Should we keep this open?  It seems work on this has stalled.  Let's close if it's not happening.

humphd:
In the absence of activity, I'm going to close.  Please re-open if that's wrong.

PR #665, opened 2023-04-24
Use certificate validFrom and validTo vs. cert order

### LINKED ISSUE #655: Certificate expiry doesn't match validity of certificate
After landing the `orderUrl` fix on production and clearing out the old certificates in the db, I was able to make a certificate.  The new certificate says it's only valid for 7 days:

Screenshot 2023-04-19 at 8 43 36 PM

However, if I decode the certificate, it shows:

- Valid From: April 19, 2023
- Valid To: July 18, 2023

So something is not right with our certificate date logic.

humphd:
Fixes #655 

To test, create a cert on dev, and it should be way in the future vs. tomorrow:

Screenshot 2023-04-24 at 1 57 29 PM

### DIFF
--- a/app/lib/lets-encrypt.server.ts
+++ b/app/lib/lets-encrypt.server.ts
@@ -1,4 +1,5 @@
 import https from 'https';
+import { X509Certificate } from 'crypto';
 import dnsPromises from 'dns/promises';
 import acme from 'acme-client';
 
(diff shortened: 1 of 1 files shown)


Eakam1007 (approved):

humphd:
Confirmed on staging:

Screenshot 2023-04-25 at 8 53 09 AM

Thanks for the reviews @SerpentBytes, @Eakam1007!

PR #667, opened 2023-04-25
Fix multiple notifications at once bug

### LINKED ISSUE #650: Multiple notifications for certificate failure
Today on production, I received 2 identical notification messages for a failed certificate.  This is likely due to having 2 instances running, and both queue and sent a notification.

It would be nice to only send a single notification.

SerpentBytes:
## Description 
Fixes #650 by making adjustments to expiration notification-related code for DNS records and Certificates. Attempts to fix the annoying bug of getting too many notifications at once regarding expiring records. _Still need to work on the bug that sends multiple notifications when a certificate fails.

It seems like the root cause of getting multiple notifications was because we were checking if `lastNotified` is set to `null` OR `lastNotified` is set to date, which is 30 days in the past. Since we are sending one reminder for now, we don't need to verify whether it has been a month since we last notified the user.

### Steps to test
1. Run the project locally
2. Request a certificate or DNS record
3. Run `npm run db:studio` and set the expiration date to _a month_ in the future
4. Navigate to http://localhost:8025 to see a single notification after some time:
demo


services/expiration/expiration-notification.server.ts
```diff
@@ -173,16 +160,7 @@ const getExpiringDnsRecords = () => {
...
-        {
-          lastNotified: null,
-        },
-        {
-          lastNotified: dayjs()
-            .subtract(NOTIFICATION_FREQUENCY * 4, 'd')
-            .toDate(),
-        },
-      ],
+      lastNotified: null,
```
humphd, line 163: I think this is the right fix for certs, since "renew" means "create a new record."  But with DNS records, we might renew multiple times.

Should we set the `lastNotified` to `null` again when we renew a DNS Record?  Then this would work I think.

SerpentBytes: No, that would send another notification.

SerpentBytes: I have modified the condition to check if the date is set to a month in the past (like we were doing before).

SerpentBytes: Yes, this works. I tested it locally.

humphd (approved):

humphd:
We only run a single instance locally, so we should do this and test in staging.  I'm going to merge this so people can test, and we'll iterate.

Issue #203, opened 2023-02-14
Domains table is not responsive


humphd:
The Domains table doesn't work well on smaller viewport sizes:

<img width="220" alt="Screenshot 2023-02-13 at 8 25 06 PM" src="https://user-images.githubusercontent.com/427398/218614084-9bece904-a577-43e0-a208-84dacc18fe42.png">

We should think about how to do a version of this that is useful on small screens.  Maybe we only show the status, name, and icons to edit/delete?

Andrewnt219:
Tailwind UI has this stacked layout table @humphd 

https://tailwindui.com/components/application-ui/lists/tables

<img width="436" alt="image" src="https://user-images.githubusercontent.com/52666982/223275818-4ba2f2b0-7542-4515-b521-e6f241c4da0b.png">

Eakam1007:
We also need to handle the current mobile view when the FQDN is too long to fit in the container

Eakam1007:
I haven't had the time to get back to this and I doubt I will any time soon. So I am gonna unassign myself for now

PR #676, opened 2023-04-27
Fix DNS record alignment issue

### LINKED ISSUE #673: Set standard width for DNS Record name
If the DNS Record name is longer or shorter, it pushes the alignment of the buttons beside it out of position.  We should fix this so the column is the same for all rows:

Screenshot 2023-04-26 at 7 53 49 PM

SerpentBytes:
## Description
Closes #673

### Demo

https://user-images.githubusercontent.com/50856799/234981029-65283f23-3b93-4864-b151-4592a37e6de7.mov

### DIFF
--- a/app/components/dns-record-table-row.tsx
+++ b/app/components/dns-record-table-row.tsx
@@ -76,56 +76,30 @@ export default function DnsRecordsTableRow({
               {dnsRecord.expiresAt.toLocaleDateString('en-US')}
             </Text>
           </Show>
-          <Hide below="sm">
-            <ButtonGroup>
-              <Link
-                href={`https://www.nslookup.io/domains/${dnsRecord.subdomain}.${baseDomain}/dns-propagation/${dnsRecord.type}/`}
-                isExternal
-                target="_blank"
-              >
-                <Tooltip label="Lookup DNS record">
-                  <IconButton
-                    icon={<InfoOutlineIcon color="black" boxSize="5" />}
-                    aria-label="Lookup DNS record"
-                    variant="ghost"
-                    ml={{ xs: '0', sm: '2' }}
-                  />
-                </Tooltip>
-              </Link>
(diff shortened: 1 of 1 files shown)


humphd (dismissed):
This works great.  I wonder if we should switch to this sooner (e.g., on tablet sizes), since the table becomes hard to use unless fully expanded:

Screenshot 2023-04-27 at 5 08 39 PM

app/components/dns-record-table-row.tsx
```diff
@@ -76,30 +76,56 @@ export default function DnsRecordsTableRow({
               {dnsRecord.expiresAt.toLocaleDateString('en-US')}
             </Text>
           </Show>
-          <ButtonGroup spacing={{ base: 'auto', sm: '0.5' }} width="full">
```
Eakam1007, line 79: Is the only difference between the blocks inside the Hide and Show the styles on the ButtonGroup. That is, one has `spacing={{ base: 'auto', sm: '0.5' }} width="full"`, and the other doesn't ?

SerpentBytes: Yes.

Eakam1007: Instead of using Hide and Show for this, you can do this: 
image
Seems to work the same from my testing

Eakam1007 (changes requested):

app/components/dns-record-table-row.tsx
Eakam1007: I didn't realize that the width would be the one causing the issue

Eakam1007 (approved):

Eakam1007:
Thank you for investigating this. 
As for the tablet size, I think we need to come up with a new layout entirely. I am planning to work on it as part of #203 sometime next week

humphd (approved):
Awesome small fix!

PR #679, opened 2023-04-30
Add functionality to delete a certificate

### LINKED ISSUE #674: Allow a cert to be deleted
@SerpentBytes and I are trying to test notifications, and we need a way to create a new cert.  Without manually editing the database, there's no way to delete a certificate right now.

It would be nice if there was a way to delete a certificate for a user.  It wouldn't get used often, but when you need it, this is important.

SerpentBytes:
## Description
Closes #674 by adding functionality to delete certificates. WIP, awaiting reviews. 

### Steps to test
- Run the project locally
- Request a certificate
- Interact with the Delete button:
delete cert sc

### DIFF
--- a/app/components/certificate/description.tsx
+++ b/app/components/certificate/description.tsx
@@ -3,6 +3,7 @@ import { Form, Link as RemixLink } from '@remix-run/react';
 import {
   Flex,
   Text,
+  IconButton,
   Button,
   Stat,
   StatLabel,
@@ -68,9 +69,10 @@ export default function DescriptionSection({
                   type="submit"
                   ml={1.5}
                   rightIcon={<DeleteIcon />}
+                  variant="ghost"
                   onClick={() =>
                     toast({
-                      title: `Certificate has been successfully deleted`,
+                      title: 'Certificate has been successfully delete',
                       position: 'bottom-right',
                       status: 'success',
                     })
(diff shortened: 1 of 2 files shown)


app/components/certificate/description.tsx
```diff
@@ -57,7 +59,24 @@ export default function DescriptionSection({
             <Flex justifyContent="flex-end">
               <Form method="post">
                 <Button type="submit" rightIcon={<RepeatIcon />} isDisabled={!isRenewable}>
-                  Renew Certificate
+                  Renew
+                </Button>
+              </Form>
+              <Form method="post">
+                <input type="hidden" name="intent" value="delete-certificate" />
+                <Button
```
humphd, line 67: Let's use `variant="outline"` or `variant="ghost"` to make this less prominent.

SerpentBytes: If the goal is to make it less prominent, should we use `IconButton` and `Tooltip`?

humphd: I think having it be a button is good, just that we only have a single "Call to Action" button (i.e., refresh).

humphd (approved):

humphd:
@SerpentBytes you still have this as "Draft."  Any particular reason?  I think we can merge.

PR #765, opened 2023-10-26
Seneca domain update

humphd:
Seneca is updating their domains, and ITS has asked me to change the dev installation to use `senecapolytechnic.ca` instead of `senecacollege.ca`.  The email address domains aren't changing yet, so I think this is all that I need.

After this gets landed, I'll likely need to manually re-deploy the swarm to pick up the docker compose file change.

I've also run `npm audit fix` to get rid of the various security issues we had.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -364,6 +364,6 @@ jobs:
           HMAC_SECRET: ${{ secrets.WEBHOOK_STAGING_SECRET }}
           TAG: staging-sha-${{ env.GITHUB_SHA_SHORT }}
         with:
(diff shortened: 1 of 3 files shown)


Eakam1007 (approved):

Genne23v (approved):

humphd:
@Eakam1007, @Genne23v thanks to you both for the reviews.

BTW, we're seeing more profs starting to use this for courses, which is great.  I'm getting ready to roll it out to the Cloud Computing class soon too.

Eakam1007:
> BTW, we're seeing more profs starting to use this for courses, which is great. I'm getting ready to roll it out to the Cloud Computing class soon too.

That is really great to hear!

Issue #663, opened 2023-04-22
DNS Record vs. DNS Records


humphd:
On the `/dns-records` route, the page's title says "DNS Record" but should be "DNS Records."

<img width="1142" alt="Screenshot 2023-04-22 at 2 27 09 PM" src="https://user-images.githubusercontent.com/427398/233800766-8290cd31-b6bc-4094-ac77-64e5bbba7189.png">

The problem happens here:

https://github.com/DevelopingSpace/starchart/blob/main/app/routes/_auth.dns-records._index.tsx#L124

This would be a good first fix for a student looking to contribute for the first time to Starchart.  Let me know if you need help.

PR #769, opened 2025-01-14
Update DNS record to DNS records

lilyhuang-github:
updated 2 instances of referencing DNS record when it should've been DNS records.

This would solve issue #663 + a different issue with it being singular instead of plural while combing through.


uday-rana (approved):
Can you rebase on main so CI passes?

Edit: Figured it out

PR #777, opened 2025-01-20
Updated a bunch of dependencies

lilyhuang-github:
updated a bunch of dependencies. 
Didn't update samilfy, and chakra-ui/react chakra-ui/icons yet since they broke since they broke the project.


humphd:
Nice work. I'll wait on CI to pass before I review this.  Ping me when it's green.

lilyhuang-github:
> Nice work. I'll wait on CI to pass before I review this. Ping me when it's green.

I will. I'm trying to fix the github actions now.

lilyhuang-github:
Ill log off now. Ill try to figure out the problem causing it to fail the tests or roll back some of the updates until it works.

lilyhuang-github:
I realize It might also be better to split up this commit cause I'm realizing more what each dependencys doing kind of

humphd:
Yes, split out the ones that are causing you issues for follow-up.

NOTE: we might also need to skip some major version bumps, since things change a lot in certain packages this way.

PR #785, opened 2025-01-22
Update dependencies 

lilyhuang-github:
The dependencies shouldn't mess with the unit tests


lilyhuang-github:
This is updating a lot less packages so there shouldn't be any issues from the unit tests.

uday-rana (approved):

humphd (commented):
Looks good.  Smart to limit this in a single PR so as to not break things, and later we can do updates one-by-one for those deps.

One tip: when we update something like `bullmq` from v3 to v5, there is bound to be some breaking changes that we should investigate--our unit tests may not cover everything.  I would be cautious with these types of updates.

PR #790, opened 2025-01-29
Package update React 18 to React 19

lilyhuang-github:


lilyhuang-github:
There's an issue with the typing between a React object and a ChakraUI property and ChakraUI can't be updated since it needs Node 20 which means this update might not work yet.

humphd:
I suspect that the move to React 19 is going to be painful.  I wonder if it's worth it?  Same for the major bump with ChakraUI.  They are both very different projects in the updated versions.

PR #804, opened 2025-01-30
Rename lint jobs to be linter-agnostic

uday-rana:
We should remove any references to ESLint from CI since we dropped ESLint.


humphd (approved):

uday-rana:
@humphd Could you update this required status check? Thanks

image

PR #810, opened 2025-01-31
Move docs into /docs

uday-rana:
I think it'd be nice to clean the project root directory up to make it a little easier to navigate. This moves the documentation and images into /docs so they're not mixed in with config files and such.


humphd (changes requested):
Leave CONTRIBUTING.md in the root, the rest looks good.

humphd (approved):

PR #830, opened 2025-02-12
Update @aws-sdk/client-route-53

theoforger:
### Description
#829 failed due to new types being introduced into `@aws-sdk/client-route-53`. This PR fixes the type mismatch cause by this update.

### DIFF
--- a/services/reconciler/createChangeSetFromCompareStructures.server.ts
+++ b/services/reconciler/createChangeSetFromCompareStructures.server.ts
@@ -3,7 +3,7 @@ import logger from '~/lib/logger.server';
 
 import { DnsRecordType } from '@prisma/client';
 
-import type { Change } from '@aws-sdk/client-route-53';
+import { Change, ChangeAction } from '@aws-sdk/client-route-53';
 import type { ReconcilerCompareStructure } from './ReconcilerTypes';
 import { toRoute53RecordValue, toRoute53RRType } from './route53Utils.server';
 
@@ -33,7 +33,7 @@ export const createRemovedChangeSetFromCompareStructures = ({
       }
 
       toChange.push({
-        Action: 'DELETE',
+        Action: ChangeAction.DELETE,
         ResourceRecordSet: {
           Name: fqdn,
           Type: toRoute53RRType(type),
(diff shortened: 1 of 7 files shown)


services/reconciler/createChangeSetFromCompareStructures.server.ts
```diff
@@ -36,7 +36,7 @@ export const createRemovedChangeSetFromCompareStructures = ({
         Action: 'DELETE',
         ResourceRecordSet: {
           Name: fqdn,
-          Type: type,
+          Type: type as RRType,
```
humphd, line 39: Instead of using `as`, which disables the compiler, can we add a function to map from type to RRType?

humphd (changes requested):
Thanks for fixing, one nit.

services/reconciler/createChangeSetFromCompareStructures.server.ts
theoforger: @humphd Sure! Do we want to throw an error in case of conversion failure?

humphd: Absolutely, since that would mean there's a bug.

theoforger:
@humphd I created a utility function to handle the conversion, and also changed more code to use the new types

services/reconciler/route53-client.server.ts
```diff
@@ -126,7 +127,7 @@ export const getDnsRecordSetPage = async (
     const command = new ListResourceRecordSetsCommand({
       HostedZoneId: await getHostedZoneId(),
       StartRecordName: fqdn,
-      StartRecordType: type,
+      StartRecordType: type ? toRoute53RRType(type) : undefined,
```
uday-rana, line 130: Can we change this function so that the `type` parameter is of type `RRType`? Seems like this comment doesn't apply anymore.

```js
export const getDnsRecordSetPage = async (
  fqdn?: string,
  // Using string as it cam be any record type not just the ones we use, also AWS sdk refers to it as string
  type?: string
)```

uday-rana (commented):
I think we need to do more here. We have a type generated by Prisma for supported types called `DnsRecordType`, and we have `RRType` from the AWS SDK. We should make an intersection type and use that throughout the codebase.

uday-rana:
@theoforger Do you mind if I work on this?

uday-rana:
Let's land this for now and do the work in a follow-up.

uday-rana (approved):

humphd (approved):

PR #831, opened 2025-02-13
Bump prisma to 4.16.2 to match @prisma/client

uday-rana:
Addresses prisma and prisma client version mismatch warning.

> \> @senecacdot/starchart@1.0.0 db:generate
> \> prisma generate
> 
> ...
> 
> warn Versions of prisma@4.16.0 and @prisma/client@4.16.2 don't match.
> This might lead to unexpected behavior.
> Please make sure they have the same version.


package.json
```diff
@@ -65,7 +65,7 @@
     "ioredis": "^5.4.2",
     "isbot": "^3.8.0",
     "nodemailer": "^6.9.3",
-    "prisma": "^4.16.0",
```
theoforger, line 68: Prisma recommends installing their cli tool as a dev dependency. ([See here](https://www.prisma.io/docs/orm/tools/prisma-cli#installation)). Maybe consider moving it?

theoforger (changes requested):
@uday-rana No problem with the version bump 👍. Just a minor detail that's overlooked before:

package.json
uday-rana: Good catch. Done.

theoforger (approved):

PR #834, opened 2025-02-13
Add ESLint 9 to support oxlint

### LINKED ISSUE #803: Add ESLint back
From https://oxc.rs/docs/guide/usage/linter:

> At the current stage, oxlint is not intended to fully replace ESLint; it serves as an enhancement when ESLint's slowness becomes a bottleneck in your workflow.
> 
> We recommend running oxlint before ESLint in your lint-staged or CI setup for a quicker feedback loop, considering it only takes a few seconds to run on large codebases.

I'm not sure completely removing ESLint was the best approach. Thoughts @humphd @theoforger ?

uday-rana:
Fixes #803

## Changes

- Added ESLint 9, which is to be run after oxlint as a fallback for linting rules that aren't covered by oxlint. Rules that oxlint does cover won't be linted again by ESLint. This is the approach recommended by oxc:

    > We recommend running oxlint before ESLint in your lint-staged or CI setup for a quicker feedback loop, considering it only takes a few seconds to run on large codebases.

  - I created the config file by grabbing the [ESLint 8 flat config template recommended by Remix](https://github.com/remix-run/remix/blob/main/templates/remix/.eslintrc.cjs) and using [ESLint's config migrator](https://www.npmjs.com/package/@eslint/migrate-config) on it. I then tweaked it from there.

- Added two oxc plugins not enabled by default: **jsx-a11y** and **import**
  - These plugins don't implement all of the linting rules of their ESLint equivalents, so we need to keep the ESLint plugins too. See https://github.com/cockpit-project/cockpit/issues/20279
- Removed some settings from .oxlintrc.json that didn't do anything or were just re-defining the defaults.
- Fixed lint errors detected after ESLint was added.


uday-rana:
The VSCode ESLint extension removes eslint-disable comments for rules that are overridden by oxlint on saving a file, making it impossible to disable rules for oxlint when using the ESLint extension. Need to find a fix.

uday-rana:
> When saving a file, the VSCode ESLint extension removes eslint-disable comments for rules that are overridden by oxlint, making it impossible to disable rules for oxlint when using the ESLint extension. Need to find a fix.

Disabled ESLint's "no unused disable directives" rule.

theoforger (approved):
@uday-rana Great job 👍. Since ESLint 9 dropped support for `.eslintignore` file, I wonder if it's still possible to have a single source for ignored files?

uday-rana:
> @uday-rana Great job 👍. Since ESLint 9 dropped support for `.eslintignore` file, I wonder if it's still possible to have a single source for ignored files? 

We can import the ignore list from gitignore into eslint: https://eslint.org/docs/latest/use/configure/ignore#including-gitignore-files

theoforger:
@uday-rana There's also a command line option [`--ignore-path`](https://oxc.rs/docs/guide/usage/linter/cli.html#ignore-files) form oxlint. Although it's not available as an config file option, we can include it in the npm script. What do you think?

To avoid confusion maybe we can include a comment in `.oxlintrc.json` explaining this situation

uday-rana:
> @uday-rana There's also a command line option [`--ignore-path`](https://oxc.rs/docs/guide/usage/linter/cli.html#ignore-files) form oxlint. Although it's not available as an config file option, we can include it in the npm script. What do you think?
> 
> To avoid confusion maybe we can include a comment in `.oxlintrc.json` explaining this situation

That could lead to inconsistent behaviour if oxlint is run with `npx` vs through the npm script. If it's not supported in the config file we could just keep things simple for now. We also don't know if using `--ignore-path` to point to .gitignore will let us add additional overrides for directories we want in Git but don't want to format.

theoforger:
@uday-rana Makes sense! Let's stick with what we have so far.

theoforger (approved):

PR #840, opened 2025-02-19
Update dependencies with major version upgrades

theoforger:
### Description

This PR updates more dependencies. Apart from some minor version bumps, it includes the following major version migrations:

- `prisma`: v4 -> v6
- `framer-motion`: v10 -> v12
- `isbot`: v3 -> v5 ([additional changes applied](https://github.com/omrilotan/isbot/blob/main/CHANGELOG.md#400))
- `react-icons`: v4 -> v5
- `husky`: v8 -> v9 ([additional changes applied](https://github.com/typicode/husky/releases/tag/v9.0.1))
- `lint-staged`: v13 -> v15
- `prettier`: v2 -> v3


lilyhuang-github (approved):
I looked at the breaking changes and it looks good.

They also change the prettier cli to ignore files in the .gitignore [prettier](https://prettier.io/blog/2023/07/05/3.0.0.html#cli) but I don't think it'll cause any issues with what's in the .gitignore currently.

uday-rana:
> I looked at the breaking changes and it looks good.
> 
> They also change the prettier cli to ignore files in the .gitignore [prettier](https://prettier.io/blog/2023/07/05/3.0.0.html#cli) but I don't think it'll cause any issues with what's in the .gitignore currently.

I recently removed duplicate entries from .prettierignore without realising the version we were on didn't already do this. Good catch.

uday-rana (approved):

PR #843, opened 2025-02-20
Upgrade Classic Remix Compiler to Remix Vite

uday-rana:
### References

[Remix Vite: Migrating](https://remix.run/docs/en/main/guides/vite#migrating)
[How to Fix CJS/ESM Bugs in Remix](https://www.youtube.com/watch?v=jmNuEEtwkD4)
https://github.com/remix-run/remix/discussions/8978
https://github.com/remix-run/remix/discussions/6004#discussioncomment-10647463


uday-rana:
CI is failing on some runs. I'm going to try using https://github.com/kiliman/remix-express-vite-plugin

uday-rana:
I don't think this is going anywhere. I can't figure out how to get the server.ts build working together with Vite, and when using the Remix Express Vite plugin the server doesn't start at all.

If we really want to do this I see two ways forward:
- Use `tsx` to run `server.ts` in production
- Convert `server.ts` to `server.js`

cc @humphd

theoforger:
@uday-rana I'd say let's not rush this (and #895) into the upcoming release. Seems like there are still some decisions to be made

PR #844, opened 2025-02-22
Update dependencies

theoforger:
### Description

This PR updates more dependencies. No breaking changes were encountered.

#### Note
- Chakra UI was only updated to the latest v2 version, since v3 introduces many breaking changes.
- React stays at version 18 because of some breakage on mobile Safari with version 19 (reported: https://github.com/DevelopingSpace/starchart/issues/841).


theoforger:
@uday-rana @lilyhuang-github This PR is ready btw. Should be easy to review since I reverted the breaking changes

package.json
```diff
@@ -77,44 +77,44 @@
...
+    "@eslint/js": "^9.21.0",
     "@playwright/test": "^1.50.1",
-    "@remix-run/dev": "2.15.3",
+    "@remix-run/dev": "2.16.0",
     "@types/bad-words": "^3.0.3",
     "@types/compression": "^1.7.5",
     "@types/cors": "^2.8.17",
-    "@types/express": "^4.17.21",
-    "@types/node": "^18.16.16",
+    "@types/express": "^5.0.0",
```
uday-rana, line 88: Is it okay to upgrade express types to 5 when we're on Express 4?

theoforger: @uday-rana That was an oversight 😆. I just upgraded express to v5 as well. Only one small change was required according to the guide: https://expressjs.com/en/guide/migrating-5.html#path-syntax

package.json
```diff
@@ -39,27 +39,27 @@
...
-    "bullmq": "^5.41.3",
+    "bullmq": "^5.41.7",
     "compression": "^1.8.0",
     "cors": "^2.8.5",
     "cross-env": "^7.0.3",
     "dayjs": "^1.11.13",
     "docker-secret": "^2.0.6",
-    "express": "^4.21.2",
-    "framer-motion": "^12.4.5",
+    "express": "^5.0.1",
```
uday-rana, line 61: Express 5 is tagged `next` on npmjs.com, indicating it's not yet stable:

https://www.npmjs.com/package/express?activeTab=versions

image

humphd: Agree, let's not move to Express 5 yet

lilyhuang-github (approved):
Didn't find any relevant breaking changes except for the express 5 which got changed back.

Issue #901, opened 2025-04-09
Migration for MX Records


humphd:
The change to add MX Record support updated the Prisma Schema, see https://github.com/DevelopingSpace/starchart/commit/9d8039e9ad968d0bf93d16bda213fc9d2f485bbf#diff-5b443964f4f3a611682db8f7e02177b0a8c632b2039e2bd5e4dd7347815c565c 

Before we can deploy what's on `main`, we need to add a database migration.  I think the process for this would be:

1. Go to a commit before the MX Record change in the Prisma Schema
1. Run the containers
2. Run `npm run setup` to generate the Prisma Client, push the current schema, seed the db
3. Go to latest `main`, with schema changes
4. Run `npm run db:migration` to generate a new migration

theoforger:
@humphd Seems like there are more things out of sync in the database schema. This is what I got when I ran `npm run db:migration`:
```
Drift detected: Your database schema is not in sync with your migration history.

The following is a summary of the differences between the expected database schema given your migrations files, and the actual schema of the database.

It should be understood as the set of changes to get from the expected schema to the actual schema.

If you are running this the first time on an existing database, please make sure to read this documentation page:
https://www.prisma.io/docs/guides/database/developing-with-prisma-migrate/troubleshooting-development

[+] Added tables
  - Certificate
  - Challenge
  - DnsRecord
  - SystemState
  - User

[*] Changed the `Certificate` table
  [+] Added foreign key on columns (username)

[*] Changed the `Challenge` table
  [+] Added foreign key on columns (certificateId)

[*] Changed the `DnsRecord` table
  [+] Added foreign key on columns (challengeId)
  [+] Added foreign key on columns (username)

We need to reset the MySQL database "starchart" at "127.0.0.1:3306"

You may use prisma migrate reset to drop the development database.
All data will be lost.
```

PR #902, opened 2025-04-10
Add migration for MX Record enum change

humphd:
We forgot to create a migration for the MX Record ENUM change.


uday-rana (approved):

humphd:
Fixes #901

PR #1012, opened 2025-11-26
Don't parse database url before passing to prisma

humphd:
I'm getting this error on staging:

```
pool timeout: failed to retrieve a connection from pool after 10001ms
```

Based on https://github.com/prisma/prisma/issues/28612 I'm trying to switch to using the raw database URL vs. parsing it.


humphd:
Confirmed fixed.

Issue #1006, opened 2025-11-26
Startup failure - error: The datasource property `url` is no longer supported in schema files


humphd:
I received an email today that the dev and prod versions are failing:

```sh
Running database migration...
npm warn exec The following package was not found and will be installed: [prisma@7.0.1](mailto:prisma@7.0.1)
Prisma schema loaded from prisma/schema.prisma
Error: Prisma schema validation - (get-config wasm)
Error code: P1012
error: The datasource property `url` is no longer supported in schema files. Move connection URLs for Migrate to `prisma.config.ts` and pass either `adapter` for a direct database connection or `accelerateUrl` for Accelerate to the `PrismaClient` constructor. 
 
    -->  prisma/schema.prisma:3
   |
 2 |   provider = "mysql"
 3 |   url      = env("DATABASE_URL")
   |
 
Validation Error Count: 1
[Context: getConfig]
 
Prisma CLI Version : 7.0.1
```

This is due to a major version update on Prisma, which affects us because the `prisma` CLI tool gets installed on startup, and it's pulling the latest version (v7) vs. the v6 version, and they aren't compatible.

PR #1013, opened 2025-11-26
Fix prod for Prisma updates

humphd:
Various fixes to get staging to work again: #1006, #1010, #1011, #1012


humphd:
We had 10 commits on `release` that weren't in `main`, which shouldn't happen.  We also had a `release` tag for some reason, which was messing up git when trying to refer to the `release` branch.  I've removed that and manually done the merge.

Issue #1009, opened 2025-11-26
Fix and re-enable e2e tests


humphd:
The e2e tests are failing in CI, and there's an odd Safari Mobile failure for me locally, see https://github.com/DevelopingSpace/starchart/pull/1007#issuecomment-3582479905.

I've disabled them for now in https://github.com/DevelopingSpace/starchart/pull/1007/commits/bda61476c57196d1994503efc101228bfa3c6906 but we should re-enable at some point.

theoforger:
Could it be related to the infamous #845? 

I managed to reproduce https://github.com/DevelopingSpace/starchart/pull/1007#issuecomment-3582479905 on several browsers:
```bash
# Notice how it always happen on toasts, which makes sense if the CSS is flaky
Call log:
      - Expect "toContainText" with timeout 5000ms
      - waiting for getByText('Public Certificate is Downloaded')


      68 |     await page.getByRole('button', { name: 'Download Public Certificate' }).click();
      69 |     const publicDownloadToastText = page.getByText('Public Certificate is Downloaded');
    > 70 |     await expect(publicDownloadToastText).toContainText('Public Certificate is Downloaded');
         |                                           ^
      71 |     await publicDownloadToastText.waitFor();
      72 |
      73 |     await page.getByRole('button', { name: 'Download Private Key' }).click();
        at /home/theoforger/Workbench/WebstormProjects/starchart/test/e2e/certificate-available.spec.ts:70:43
```
```bash
  6 failed
    [firefox] › test/e2e/certificate-available.spec.ts:66:7 › Certificate Page › Download Certificate
    [Mobile Chrome] › test/e2e/certificate-available.spec.ts:66:7 › Certificate Page › Download Certificate
    [Mobile Safari] › test/e2e/certificate-available.spec.ts:37:7 › Certificate Page › Copy Certificate
    [Mobile Safari] › test/e2e/certificate-available.spec.ts:66:7 › Certificate Page › Download Certificate
    [Microsoft Edge] › test/e2e/certificate-available.spec.ts:66:7 › Certificate Page › Download Certificate
    [Google Chrome] › test/e2e/certificate-available.spec.ts:66:7 › Certificate Page › Download Certificate
  1 flaky
    [Google Chrome] › test/e2e/certificate-available.spec.ts:37:7 › Certificate Page › Copy Certificate
```

And here's a recording from Playwright:

[...]

Issue #845, opened 2025-02-24
React hydration error


uday-rana:
![Image](https://github.com/user-attachments/assets/3d58d83c-987f-4163-9807-d4a2ac792927)

![Image](https://github.com/user-attachments/assets/cb2a6239-0b57-4787-92a7-d20ffc63323b)

https://react.dev/errors/418?invariant=418

theoforger:
@uday-rana Hello! I'm trying to reproduce this. Did you encounter this on a dev build or a production build? What browser were you using?

uday-rana:
@theoforger In production (https://mycustomdomain.senecapolytechnic.ca/) on Firefox on Windows 10 22H2

theoforger:
Although I'm not able to reproduce this issue, I suspect it might have to do with CSP ([Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP)). According to the screenshots the policy for script is set up as `script-src 'self' 'nonce-<value>'`, reflected in the source code here:

https://github.com/DevelopingSpace/starchart/blob/82608ec3806481f89ea770496b656de77770be03/server.ts#L25C1-L35C7

Perhaps somewhere the code to utilize `nonce` is buggy, causing some UI-related scripts to be rejected.

I'll keep looking into it. Others feel free to investigate as well! 😆

theoforger:
I finally ran into this issue after leaving the "DNS records" page open and idle overnight. And other pages seem to load just fine. It might not be related to CSP after all, since I didn't see the error message here.

Strangely, every time I refresh, the style sheet seems to load for a split second and then fail:

https://github.com/user-attachments/assets/0299d9f7-9fe4-43fc-aa18-38320533542f

theoforger:
@humphd I might have found the cause of this bug:

The dates displayed on the Certificate page are rendered on the server side, which can cause a disagreement with the client if the browser is in a different time zone, therefore the hydration issue.

I encountered this during my tests for #1009 - On the failed playwright runs, I'm getting this in the console:

```
Warning: Text content did not match. Server: "Nov 27, 2025" Client: "Nov 28, 2025"
    at dd
    at http://localhost:8080/build/_shared/chunk-WYEVIKHP.js:34609:45
    at ChakraComponent2 (http://localhost:8080/build/_shared/chunk-WGRDFYGG.js:5183:15)
[...]

PR #1014, opened 2025-11-28
Use client side rendering for local dates

theoforger:
### Description
Currently the rendering of certificate valid dates happens on the server side, which could potentially lead to hydration issue and it might be the cause of #845 and #1009. This pull request move the rendering logic to the client side and attempts to re-enable e2e testing.

Update: There seem to be some Prisma-related problem that's also failing the test in addition to the hydration issue. More investigation needed

### DIFF
--- a/app/components/certificate/description.tsx
+++ b/app/components/certificate/description.tsx
@@ -1,5 +1,6 @@
 import { RepeatIcon, DeleteIcon } from '@chakra-ui/icons';
 import { Form, Link as RemixLink } from '@remix-run/react';
+import { useEffect, useState } from 'react';
 import {
   Flex,
   Text,
@@ -15,8 +16,8 @@ import {
 
 interface DescriptionSectionProps {
   certRequested: boolean;
-  validFromFormatted?: string;
-  validToFormatted?: string;
+  validFromISO?: string;
+  validToISO?: string;
   description: string;
   isRenewable?: boolean;
   link?: string;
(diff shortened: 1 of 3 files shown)


humphd:
Should we do the date change and separate from the e2e stuff, so we can land that bit?

theoforger:
> Should we do the date change and separate from the e2e stuff, so we can land that bit?

@humphd I was thinking the same. The only thing is we won't be able to verify the effectiveness in CI. If you're okay with that I can take out the e2e test for now 👍

humphd:
Yeah, let's get your research shipped at least.  Tests can come later.

app/components/certificate/description.tsx
```diff
@@ -87,3 +92,20 @@ export default function DescriptionSection({
...
   );
 }
+
+function LocalDate({ iso }: { iso: string }) {
+  const [formatted, setFormatted] = useState('');
+
+  useEffect(() => {
+    const date = new Date(iso);
+    setFormatted(
+      date.toLocaleDateString('en-US', {
```
humphd: en-CA :)

app/components/certificate/description.tsx
```diff
@@ -87,3 +92,20 @@ export default function DescriptionSection({
...
+        month: 'short',
+        day: '2-digit',
+        year: 'numeric',
+      })
+    );
+  }, [iso]);
+
+  return <>{formatted}</>;
```
humphd: You could use `<time>{formatted}</time>`

humphd (dismissed):
Looks good.  A couple of suggestions.  Thanks for jumping on this, I appreciate the help maintaining things.

theoforger:
Of course! I enjoyed working on this project. Glad to see it picked back up :)

humphd (approved):

humphd:
Confirmed fixed on staging:

Screenshot 2025-11-28 at 12 18 07 PM

PR #1016, opened 2026-01-20
Update staging SAML metadata with new SAML signing certificate

mehrdadziaei:
New metadata has two certificate in it, one the current active one that expires soon and another one that is in inactive state in Microsoft IdP.
We do not know how the saml library handles multiple certificates in the metadata.  
Ideally it should try both, but needs to be tested.

After deployment, I will make the new certificate active in Microsoft IdP and test login to see if it works.


humphd (changes requested):
Because we use the same code to build the staging and production images, and merge from `main` to `release` to do it, we should add the production version here at the same time, so it's there when we merge later.

We won't actually switch to the new production one automatically until we test and you switch things over, though.

mehrdadziaei:
I'm not sure if this process would work or not, that's why I did only for non-prod to test.
however, I just updated it for prod too, in case it works.

btw, I made idp-metadata-dev.xml to be the same as idp-metadata-staging.xml as well.

humphd (changes requested):
The change to the `config/idp-metadata-dev.xml` file will break the  development setup (NOTE: we use `dev` to mean "local dev" vs. `"staging"`, which I know you call "dev" so it's confusing).

Can you revert that change?  The rest is good.

After that, I'll merge and we can test on the staging server to see if it works.

humphd (approved):
Looks good.  I'll merge so we an see how it behaves on staging.

mehrdadziaei:
Thanks, 
I swapped the certificate for staging in Azure and the sso login is still working, didn't expect it to be that easy!

humphd:
Fantastic. Are you OK if I merge this and we try on production? Let me know when and I'll do it.

mehrdadziaei:
Since the SAML library can use both certificates, there was no outage for switching active certificate.
You can deploy to prod anytime after hours and let me know,
then I will swap the certificates on Azure.

PR #1019, opened 2026-05-06
Fix and re-enable e2e tests

humphd:
I spent some time debugging and fixing the e2e tests.  I'm able to pass everything locally at this point.  I had to do the following to make it all work:

- Re-enable `E2E-Tests` in CI
- Pin the version of MySQL to 8.0
- Update the `command` for MySQL so the password works properly in testing
- Changed how the `Sign out` link works to use a proper HTML `<form>` which passes cookies correctly (Remix is faking it with `fetch()` internally, and missing the one we need for slo)
- Added the `nonce` to the streaming remix update script
- Fixed the piping order for how Remix deals with styles and emotion
- Updated playwright and changed how we do the config to use the recommended function
- Updated our CSP to allow for `form-action` and `style-src` to work
- Fixed the module format with how our Prisma client is generated

Let's see if this works in CI...it does not :)

I had to do more work, but now it all passes:

- The way I was building the Prisma client was fragile, especially since we used to cache it.  I changed CI to always prune old generated clients
- I updated the service containers in CI to match how we do things with docker compose
- I added a little startup "wait for ready" script on the login container in e2e tests, so it's always ready
- I fixed a race and error handling crash in `app/models/system-state.server.ts` so that when tests or other conditions mean we try to write the system state when it's already there, we don't blow up unnecessarily
- I made the `header.spec.ts` test less prone to failure if the login container didn't respond properly on logout.  We don't really care about logout, only that the UI tries to do it.
- I fixed a race in the `test/unit/route53-client.server.test.ts` test with resetting and creating the hosted zone, which sometimes didn't exist in test runs

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,6 +26,11 @@ jobs:
       - name: Install node_modules
         run: npm ci
 
+      - name: Clean old Prisma artifacts
+        run: |
+          rm -rf node_modules/.prisma
+          rm -rf generated/client
+
       - name: Generate prisma client types
         run: npm run db:generate
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,7 +21,7 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 22
-          #cache: npm
+          cache: npm
 
       - name: Install node_modules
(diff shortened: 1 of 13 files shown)


humphd:
This is working now.  I've updated my original comment to include all that I had to do.  It was more complicated than I expected.

PR #1021, opened 2026-05-06
More dep updates (May 6, 2026)

humphd:
Now that I have tests passing again, I wanted to do more updates.  First, non-breaking changes, then I'll fix some breaking API version bumps.


humphd:
Turns out a few deps can't be moved, due to `@remix-run/dev`.  However, all the remaining high severity dep issues are just dev deps.

